PR #161, opened 2017-09-03
Customization

schani:


src/Language/CSharp.purs
```diff
@@ -2,29 +2,34 @@ module Language.CSharp
...
 import Data.Maybe (Maybe(..))
 import Data.Set (Set)
 import Data.String.Util (camelCase, legalizeCharacters, startWithLetter, stringEscape, isLetterOrLetterNumber)
+import Options (booleanOption, Option)
 
 forbiddenNames :: Array String
 forbiddenNames = ["Convert", "JsonConverter", "Type"]
 
+listOption :: Option Boolean
+listOption = booleanOption "use-list" "Use List<T> instead of T[]" false
```
dvdsgl: Perhaps `--sequence-type array|list` instead of `--use-list yes|no`? It might be nice to generalize a bit so this works naturally with other renderers and sequence types.

cli/quicktype.ts
```diff
@@ -100,7 +120,32 @@ const sections = [
...
+function optionDefinitionsForRenderer(renderer: Renderer): OptionDefinition[] {
+  return _.map(renderer.options, o => {
+    return {
+      name: o.name,
+      description: o.description,
+      typeLabel: o.typeLabel,
+      renderer: true,
+      type: String as any } as OptionDefinition;
+    });
+}
```
dvdsgl: Is there any reason we might want to prefix renderer options by name? For example, if the C# renderer specifies an option `use-lists`, might we want to present it as `--csharp-use-lists`?

src/Language/CSharp.purs
dvdsgl: I just don't like `--use-list no` because it's not clear what that implies, whereas `--sequence-type array` is explicit.

src/Language/CSharp.purs
```diff
@@ -85,7 +90,12 @@ renderTypeToCSharp = case _ of
     IRString -> pure "string"
     IRArray a -> do
         rendered <- renderTypeToCSharp a
-        pure $ rendered <> "[]"
+        useList <- getOptionValue listOption
```
dvdsgl: Nice!

src/Main.purs
```diff
@@ -29,8 +34,8 @@ renderers :: Array Doc.Renderer
 renderers = Renderers.all
 
 -- json is a Foreign object whose type is defined in /cli/src/Main.d.ts
-main :: Json -> Either Error SourceCode
-main json = do
+main :: StrMap String -> Json -> Either Error SourceCode
+main optionStrings json = do
```
dvdsgl: This breaks `quicktype.io` so please bump the major release number in `package.json`.

src/Options.purs
```diff
@@ -0,0 +1,74 @@
...
+        specMap :: StrMap OptionSpecification
+        specMap = SM.fromFoldable $ map (\spec -> Tuple spec.name spec ) optionSpecifications
+
+        makeOptionValue :: String -> String -> OptionValue
+        makeOptionValue name optionString =
+            -- Error handling here - we assume the option has a spec
+            let spec = unsafePartial $ fromJust $ SM.lookup name specMap
+            in
+                let l = String.toLower optionString
+                in BooleanValue $ l == "true" || l == "t" || l == "yes" || l == "y" || l == "1"
```
dvdsgl: ```l `elem` ["true", "t", "yes", "y", "1"]```

cli/Main.d.ts
```diff
@@ -1,15 +1,9 @@
...
-    aceMode: string;
-}
-
 type SourceCode = string;
 type ErrorMessage = string;
 
 interface Main {
     renderers: Renderer[];
-    main(config: Config): Either<ErrorMessage, SourceCode>;
+    main(options: {[name: string]: string}): ((config: Config) => Either<ErrorMessage, SourceCode>);
```
dvdsgl: Can you uncurry this? `Main.main` is always called from Type/JavaScript so we could just do the parameters as a record.

cli/quicktype.ts
dvdsgl: It's verbose but would it be worthwhile to make the option parsing simpler?

cli/quicktype.ts
```diff
@@ -298,13 +362,28 @@ class Run {
     }
   }
 
-  getOptions = (argv: string[]): Options => {
-    let opts: { [key: string]: any } = commandLineArgs(optionDefinitions, { argv });
-    let sane = _.mapKeys(opts, (v, k) => {
-      // Turn options like 'src-urls' into 'srcUrls'
-      return _.lowerFirst(k.split('-').map(_.upperFirst).join(''));
+  parseOptions = (optionDefinitions: OptionDefinition[], argv: string[], partial: boolean):
```
dvdsgl: Can you please add a comment explaining this function?

src/Options.purs
```diff
@@ -0,0 +1,74 @@
...
+    = BooleanValue Boolean
+
+type OptionSpecification =
+    { name :: String
+    , description :: String
+    , typeLabel :: String
+    , default :: OptionValue
+    }
+
+type OptionSpecifications = Array OptionSpecification
```
dvdsgl: I think you can define `OptionSpecifications` as an Array of records that contain `specification :: OptionSpecification`:

```purs
type OptionSpecifications = Array { specification :: OptionSpecification | r }
```

I'm not sure if you have to quantify the `r` (you might need a `forall r.`) but this should let renderers do:

```purs
, options: [listOption]
```

without having to add `.specification`.

dvdsgl: I can't quite get this to work but I think it's possible.

cli/Main.d.ts
dvdsgl: Or could `options` be a property of `Config`? That seems better.

cli/quicktype.ts
schani: Maybe, maybe not.  I don't like it.

src/Language/CSharp.purs
schani: Implemented.

src/Main.purs
schani: This is breaking my spirit already.

schani: Actually, I'll make a separate `mainWithOptions` function, so the API doesn't change at all.  And it's better if it's curried, obviously ;-)

src/Options.purs
schani: I don't understand how that works.

cli/Main.d.ts
```diff
@@ -1,15 +1,10 @@
...
-    aceMode: string;
-}
-
 type SourceCode = string;
 type ErrorMessage = string;
 
 interface Main {
     renderers: Renderer[];
     main(config: Config): Either<ErrorMessage, SourceCode>;
+    mainWithOptions(options: {[name: string]: string}): ((config: Config) => Either<ErrorMessage, SourceCode>);
```
dvdsgl: `Config` already is a bag of options that's designed to make `Main` easier to call from `quicktype.io` and our IDE extension. Why are you adding additional options outside of `Config`? It's totally fine to change `Config` and therefore `Main.main` – when I said it would 'break' `quicktype.io`, I didn't mean you should not do it!

Issue #155, opened 2017-09-01
Eliminate dead code from PureScript output


dvdsgl:
We're shipping 10mb of JavaScript output by the PureScript compiler. This is a pretty big increase in code size compared to the 500kb `bundle.js` we used to ship. Much of this is dead code that can be pruned with an option to `pulp build`.

schani:
No more PureScript in 1f91b59eadfcfd9d353aba43f874261c60539986.

PR #163, opened 2017-09-04
Adds DCE elimination for PureScript code.

damncabbage:
A bundle is now spat out at `output/purs/index.js`, with some shims in that directory for each of the exposed modules (just `Main` and `Samples` for now); the cli JS now looks there instead.

This is, unfortunately, a little horrific. It's both working around a problem where I couldn't add `--to` without having the other options be rejected (https://github.com/purescript-contrib/pulp/issues/208#issuecomment-326813274), and also trying to preserve the `Main` and `Samples` modules as separate files even after bundling (to keep the `Main/index.js` <--> `Main.d.ts` correspondence you have).

(Happy to treat this as an experiment that is quietly closed, if this is all a bit too much. 💦 )

### DIFF
--- a/cli/quicktype.ts
+++ b/cli/quicktype.ts
@@ -5,7 +5,7 @@ import * as Either from "./either";
 import tryRequire from "./try-require"
 import * as _ from "lodash";
 
-const Main: Main = tryRequire("../output/Main", "./Main");
+const Main: Main = tryRequire("../output/purs/Main", "./Main");
 const makeSource = require("stream-json");
 const Assembler = require("stream-json/utils/Assembler");
 const commandLineArgs = require('command-line-args');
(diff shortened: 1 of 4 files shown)


dvdsgl:
I appreciate the effort but I worry about making our already piecemeal tooling even more so.

Do you know how this interacts with the `psc-ide` auto-building? Do editor extensions need to be told to output to `./output/purs` instead?

I'd prefer to wait until we can accomplish this in a way that's easy to understand and well-supported by `purs` or `pulp`. I'm not so worried about the 10mb npm module that I'd want to introduce some potential headaches and increased maintenance cost right now.

dvdsgl:
Fixes #155

damncabbage:
Oof, understood; happy to close this and revisit later. (You're right about the issues.)

PR #171, opened 2017-09-09
Big list of naughty strings

schani:


schani:
Oh, that's nice! That also works with files not in the test directories?

dvdsgl:
Yes I think so.

PR #196, opened 2017-09-19
Use NSNull for null in Swift

schani:


schani:
Don't merge yet.  This might change for Swift 4.

schani:
This was part of Swift 4.

PR #203, opened 2017-09-20
Use the module name as the C# namespace

schani:


schani:
I forgot we have this module name thing.  That would allow us to have a global option for module/namespace name, instead of it being language-specific.

schani:
@dvdsgl are you ok with merging this?

dvdsgl:
I found that sometimes there are issues with the namespace and top-level class have the same name.

schani:
Then shouldn't we rename the namespace if the user wants the toplevel to be called `QuickType`?

schani:
We use an option for this now.

PR #204, opened 2017-09-22
Search renderers only in PureScript

schani:


schani:
@dvdsgl take a look, please.

cli/Prelude.d.ts
```diff
@@ -1 +1,2 @@
 type Either<L, R> = { value0: L | R };
+type Maybe<T> = { value0: T };
```
dvdsgl: Is it not `{ value0?: T }`?

cli/maybe.ts
```diff
@@ -0,0 +1,7 @@
+export function isJust<T>(maybe: Maybe<T>): boolean {
+  return maybe.constructor.name === "Just";
+}
+
+export function fromJust<T>(maybe: Maybe<T>): T {
+  return maybe.value0;
```
dvdsgl: Do you want to throw an exception if this is invalid?

cli/Main.d.ts
```diff
@@ -3,6 +3,7 @@ type ErrorMessage = string;
 
 interface Main {
   renderers: Renderer[];
+  rendererForLanguage(language: string): Maybe<Renderer>;
```
dvdsgl: Now that we distribute as normal modules, you can define `rendererForLanguage` in `Language/Renderers.purs` to avoid bloat in `Main`.

src/Main.purs
```diff
@@ -29,6 +31,9 @@ intSentinel = IRTypeable.intSentinel
 renderers :: Array Doc.Renderer
 renderers = Renderers.all
 
+rendererForLanguage :: String -> Maybe Doc.Renderer
+rendererForLanguage = Renderers.rendererForLanguage
```
dvdsgl: Ah, I see -- yes, you no longer need to do the re-export from Main trick, as you can require any of our modules from TS.

src/Language/Renderers.purs
```diff
@@ -22,3 +28,10 @@ all =
     , SimpleTypes.renderer
     , JsonSchema.renderer
     ]
+
+rendererForLanguage :: String -> Maybe Doc.Renderer
+rendererForLanguage language =
+    find match all
+    where
+        match :: Doc.Renderer -> Boolean
+        match ({ names }) = elem language names
```
dvdsgl: Extra parentheses.

src/Main.purs
schani: That means we can also remove `Main.renderers` in favor of `Renderers.all`?

Do I just have to name the definition file `Renderers.d.ts` and it'll work?

dvdsgl: Yes, I think that’s exactly how it works.

Issue #187, opened 2017-09-17
Run `list` test for all languages


schani:
This is now the `JSONSchemaFixture`.  It runs on all languages except for

- [x] Swift: In Swift we use `struct`s for everything, which doesn’t allow us to represent recursive types.  However, if we use `class` instead we can’t have the initializers the way we have them now.  In particular, we can’t have an initializer that does `self = ...`.  I think we should just use a static method `Foo.from(jsonString: ...)`.  That’s about the same amount to type, and it works with `struct`s and `class`es.

- [ ] Elm: First we're using `type alias` to define all our types, and you can't `alias`-define a recursive type. Removing the `alias` leads to some trouble in the definition of the encoders and decoders which I didn't investigate further.

schani:
We have the option to produce `struct`s or `class`es for Swift.  The `list` test runs with `class`es now.

We should automatically detect cycles in the IR and generate classes for those, and structs for the rest, at least by default.  Having this analysis might be useful for other languages, too.  One other use case would be forward declarations, eg for C++.

schani:
Duplicate of #36 and #273

PR #208, opened 2017-09-24
Refactor tests to abstract out the target language

schani:


schani:
@dvdsgl I don't want to merge this without you having taken a look first.  The purpose of this refactoring is to make it possible to run tests other than with JSON input across different languages.  In particular see #187.

dvdsgl:
Looks great, although at 741 lines I think it's time to split `test.ts` into multiple files.

PR #215, opened 2017-09-28
Option to disable inferring maps. Fixes #69

### LINKED ISSUE #69: Users should be able to determine maps vs objects
`dependencies` and `devDependencies` are `Map String String` but are treated as custom objects.

We'll have to expose some way for people to determine these.

schani:


schani:
@dvdsgl ?  This is not ideal, as it applies to all maps, but at least one user needed it, so in lieu of a better solution, which we won't have soon, I think this does the job.

dvdsgl:
Looked good to me.

PR #218, opened 2017-09-29
Trying to enable TypeScript strict null checks

schani:
I fixed all our code to work with strict null checks
(might have caught one or two unproblematic bugs),
but I don't know how to configure it such that only
our code, and not imported modules, are compiled
with strict null checks.  At least one module doesn't
compile.

### DIFF
--- a/cli/either.ts
+++ b/cli/either.ts
@@ -10,11 +10,11 @@ export function get<T, U>(either: Either<T, U>): T | U {
   return either.value0;
 }
 
-export function fromRight<T>(either: Either<string, T>): T {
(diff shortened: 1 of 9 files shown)


schani:
@dvdsgl If you know how to enable strict null checks just for our code, that would be really nice.

dvdsgl:
I enabled `strictNullChecks` in `cli/tsconfig.json` and was able to build without error. How did you build with errors?

schani:
I don't know, I'll try again ;-)

schani:
Yeah, weird, works for me, too.

schani:
@dvdsgl Shall we merge this?

PR #263, opened 2017-11-07
Swift in typescript

schani:
This leaves us with Swift 3 in PureScript and Swift 4 in TypeScript.

### DIFF
--- a/src-ts/ConvenienceRenderer.ts
+++ b/src-ts/ConvenienceRenderer.ts
@@ -197,9 +197,16 @@ export abstract class ConvenienceRenderer extends Renderer {
 
     protected forEachTopLevel = (
         blankLocations: BlankLineLocations,
-        f: (t: Type, name: Name) => void
+        f: (t: Type, name: Name) => void,
+        predicate?: (t: Type) => boolean
     ): void => {
-        this.forEachWithBlankLines(this.topLevels, blankLocations, (t: Type, name: string) =>
+        let topLevels: Collection<string, Type>;
+        if (predicate) {
+            topLevels = this.topLevels.filter(predicate);
+        } else {
+            topLevels = this.topLevels;
+        }
+        this.forEachWithBlankLines(topLevels, blankLocations, (t: Type, name: string) =>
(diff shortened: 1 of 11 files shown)


schani:
@dvdsgl How do you feel about this? We'd have two separate Swift languages with this PR, as opposed to one with an option.  Or should we just get rid of Swift 3 right away?

dvdsgl:
I'm fine if you remove Swift 3 entirely, but could you just jerryrig the PureScript renderer into the TypeScript renderer by calling the PureScript renderer?

schani:
That depends on how long we want to keep it around.  Here's a proposal:

1. Make a target language wrapper for Swift that uses either PS Swift or TS Swift, depending on the option passed.

2. Make PS target languages report an error when trying to generate code for an enum.

3. Implement enums in TS Swift.

4. Ship TS Swift and enums.

At that point we'd have enums in all TS renderers, and we'd fail with an error for PS renderers when enums are used.  That's not a big deal (yet), because enums can only can from JSON Schema.  I wouldn't talk much about the feature at that point.

Once we implement inferring enums and/or GraphQL we should move away completely from PS renderers, because those are high(er) profile features and I think we should support them on all target languages.

dvdsgl:
Let's just drop Swift 3 and reconsider if we hear anyone ask about it. Sounds like a lot of effort if we only have a handful of Swift 3 users, and they can always stick with an older version of quicktype.

schani:
That still leaves us with Elm and Java not supporting enums yet, and I don't have the motivation to port them to TS at this point, so really the only additional effort my proposal entails is the target language wrapper, which should be pretty trivial.

schani:
Basically, the options are:

1. Port the last two PS renderers to TS.  I don't have the motivation to do this right now, so you'll have to do it, soon.

2. Keep the PS renderers around until we have either GraphQL or enum inference, at which point my motivation to port renderers will probably be sufficient to do it.  In this case it costs us very little to keep Swift 3.

schani:
@dvdsgl If the checks pass I think this is ready to merge. Let me know if you have any objections.

PR #280, opened 2017-11-18
Inference in ts

schani:


dvdsgl:
In the Pokedex sample, `pokemon.weaknesses` is inferred as `Weakness[]`, but `pokemon.type` is inferred as `string[]`, when they have the same domain.

dvdsgl:
Also, there seems to be an opportunity to merge `enums` with the same cases. From `us-senators.json`:

```csharp
    public enum SenatorRankLabel { Junior, Senior };
    public enum SenatorRank { Junior, Senior };
    public enum SenatorClassLabel { Class1, Class2, Class3 };
    public enum SenatorClass { Class1, Class2, Class3 };
    public enum RoleTypeLabel { Senator };
    public enum RoleType { Senator };
    public enum GenderLabel { Female, Male };
    public enum Gender { Female, Male };
```

schani:
So I have this metric for when something is a enum:

* all its items are short strings
* the number is unique items is at most the square root of the number of its total items

For the `type`, the number of unique items is 15, and they're 213 items total, which is just short of 15^2, which is 225.  Suggestions?

And yes, once I have a better graph representation we can merge enums, too.

dvdsgl:
I think that's a good heuristic. I'm wondering if we can use the existence of an inferred enum to determine the types of samples that meet a weaker form of the heuristic.

schani:
I'm not entirely sure I understand.  You mean if we have an "established" enum, make any string type that conforms to that enum to be that enum type, too, whether or not it meets the counts requirement?

Note that this wouldn't save you here because `type` has a case that's not in weaknesses (`Normal`).

dvdsgl:
Yes, that's what I meant. Sorry I was typing while commuting in the Tokyo metro! Yes I see it wouldn't work in this case.

PR #287, opened 2017-11-24
Remove PureScript

schani:


wclr:
@schani 
just wonder, why did you decide to migrate from PureScript to TypeScript back then?

Hi-Angel:
@schani same question. Came here because this project was referred to in StackOverflow question about making TypeScript and PureScript coexist. Was surprised to see there's no PureScript, found this PR, but no explanation.

Hi-Angel:
The PR has no code-review either, but I see commits from @dvdsgl, perhaps was discussion happening elsewhere? Anyway, CCing David Siegel as one of the authors as well.

schani:
> @schani Just curious, why did you decide to migrate from PureScript to TypeScript back then?

Sorry, I missed this question way back when.

A couple of reasons:

- I found it very tedious to write purely functional code. That's just me.
- Performance was bad. Maybe because I'm bad at Purescript.
- Purescript was very immature. The straw that finally broke the camel's back here was [this issue](https://github.com/glideapps/quicktype/commit/618fb088ffd52f481709e17aceee2a0c1d2db374).

The switch ended up being a complete success. I do have to say that having Purescript's type checking was very nice, but it was not worth all the other downsides. Another upside we got, that I don't remember we even considered, was that going with Typescript allowed us to have many more outside contributors.

Hi-Angel:
Oh, I see, thanks for explanation!

> * Performance was bad.
> […]
> * Purescript was very immature. The straw that finally broke the camel's back here was [this issue](https://github.com/glideapps/quicktype/commit/618fb088ffd52f481709e17aceee2a0c1d2db374).

For the record, PS is indeed immature, and both issues you mention are due to that. PS has problems with tail-call optimization which is mandatory for such language *(even as of writing the words)*; and for some unknown reason optimizations are implemented as [a third-party project](https://github.com/aristanetworks/purescript-backend-optimizer) instead of being part of the compiler.

Both of these are the reasons for problems you met, so yeah… I understand what you're talking about.

Hi-Angel:
> * I found it very tedious to write purely functional code. That's just me.

This one is too due to compiler being immature. For example Haskell's has much better error-reporting and it even has "suggestions" where it sees certain patterns that can be replaced with lesser code. So yeah…

schani:
> This one is too due to compiler being immature. For example Haskell's has much better error-reporting and it even has "suggestions" where it sees certain patterns that can be replaced with lesser code. So yeah…

That's part of it, but I also think the purely functional programming model doesn't mesh with the way my brain works.  I've written a bunch of Ocaml and various Lisps, mostly pure, and loved it, but going all the way always feels like I'm working with one arm tied behind my back.  One of my hot takes is that monads are a bug, not a feature, of functional programming. As in: they're a workaround that make pure's limitations at least somewhat bearable. Sure, they also let you do some other neat things, but I think in the grand scheme of things those are vastly overinflated.

PR #288, opened 2017-11-24
Clean up build process now that PureScript is removed

dvdsgl:


script/build.ts
```diff
@@ -56,6 +36,10 @@ function makeDistributedCLIExecutable() {
   shell.chmod("+x", cli);
 }
 
+function bundleDistributables() {
+  shell.cp(["readme.md", "package.json"], "dist");
+}
```
schani: `README.md` is not all lower-case.  This will not work on Linux.

dvdsgl: Hmm, I changed the filename but git didn't notice. Will fix.

PR #290, opened 2017-11-25
Remove TypeScriptTargetLanguage

dvdsgl:


schani:
I don't like this `default` export, and [other people agree with me](https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html).  The default should be not to use `default`.  The only case in our codebase where I accept `default` as valid is for the actual target language implementations.  With everything else you new know whether you'll need to add a type to the exports, for example.  Plus you'll never be sure whether you need the default export or a specific one.

PR #311, opened 2017-12-03
Split CLI-specific functionality out of main module

dvdsgl:


dvdsgl:
@schani can you please look at the issue in `Types.ts` – there are multiple declarations of the same field. I did 95928017c5bc065be1b0197faa678ea16f38bb59 just to make it build but I'm not sure what's correct.

Issue #302, opened 2017-12-01
An acyclic, naming IR


schani:
Our IR allows recursive types the renderers don't support.  For example, it's possible to define, in a JSON Schema, a type like this:

    T = Map<string, T>

We currently distinguish between "named" and "non-named" types.  Named types get a name during type graph construction, and renderers usually produce explicit type definitions for them.  That's why a class like this works with most target languages:

    T = class { property: T | null }

Maps are not named types, but the proper solution isn't to make all potentially recursive types named (maps and arrays are the two missing ones), because in most cases you don't want separate type definitions for those types.

The proper solution is to detect where the type graph contains cycles, and to explicitly represent type definitions in cases where they're needed.  That would mean that those types have to carry name information, for the case that they need to be explicitly named, so in a way they would become named types, but named types wouldn't be super special anymore.

A target language could now specify which kinds of types it always wants definitions for.  TypeScript, for example, doesn't want definitions for unions.  The acyclic IR would represent all those type definitions, plus those required to resolve cycles.  The target language will additionally be able to specify whether it requires types to be defined before they're used, like C++, for example.  If that's required, and a cycle occurs, the acyclic IR will also add forwarding declarations.

Having cycles resolved will also allow target languages which use containment vs reference by default to only use the minimum amount of references, as opposed to having the user figure out that there's a cycle and then switch to all-references.  That applies to C++, for example.

schani:
There are at least two ways to solve the containment vs reference issue:

1. Solve it at the type level, i.e. mark types as either value or reference types. The works in languages like Swift, which distinguishes between those.  Languages like C++ or Go would have to "emulate" this by using reference types via pointers.

2. Solve it at the usage level, i.e. mark each use of a type within another type as being either containment or reference.  This has the disadvantage of being hard to implement in, say, C#, where value types are always contained.  Also, it seems like it might be too confusing, having to distinguish by context whether to dereference or not, vs having to do it on the type level.  In addition to that it's harder to implement in the IR.

schani:
An algorithm for figuring out which types need to be reference types, as well as arriving at an ordering for declarations for languages that require forward declarations, like C++:

```
DECLARED_TYPES = {}
FORWARDED_TYPES = {}
for each top level node TL:
    visit(TL, [])

visit(N, PATH):
    if N in DECLARED_TYPES:
        return
    if N in PATH:
        mark N as reference type
        emit forward declaration for N
        FORWARDED_TYPES.add(N)
        return
[...]

PR #313, opened 2017-12-04
quicktype v7

dvdsgl:
- Faster type inference (@schani can you quantify or add any details?)
- Allow recursive maps in JSON Schema
- Separate type inference and code rendering, allowing caching
- TypeScript enums
- GraphQL
- Simpler main module


schani:
I think we should distinguish between stuff that affects users and stuff that's just infrastructure when we write release notes/changelogs.

Stuff that affects users:

* Faster type inference - what it says on the box.  I'll blog about it at some point, but that's not interesting to users.
* TypeScript enums
* GraphQL

Stuff that doesn't affect users:

* Recursive maps in JSON Schema.  Doesn't affect users because no target language supports it yet.  See #302
* Caching.  Doesn't affect users yet because parsing is slow part, and we don't cache that.
* Simpler main module

PR #325, opened 2017-12-08
Nicer DependencyNamer

schani:


schani:
Already merged via #324

PR #328, opened 2017-12-10
[DO NOT MERGE] Use JSON-order for class properties

schani:


src/TargetLanguage.ts
```diff
@@ -21,7 +23,9 @@ export abstract class TargetLanguage {
         if (this._options === undefined) {
             return panic(`Target language ${this.displayName} did not set its options`);
         }
-        return this._options.map(o => o.definition);
+        const definitions = this._options.map(o => o.definition);
+        definitions.push(this._sortPropertiesOption.definition);
+        return definitions;
```
schani: @dvdsgl This sucks, too.  I'm adding the sort option to every single target language because we don't have a nice way to pass non-language specific options.  Should we just pass a subset of the `Options` to the target language, including the `rendererOptions` there, too?

dvdsgl: For what it's worth, I don't think property order should be an option—it should just always match the JSON order.

schani: For the time being we need it for testing. We had a test that does JSON->X vs JSON->Schema->X and compares the results. Since we're using the standard JSON parser to read schema we're not preserving the order there, so I have to use a canonical order. Ideally we would preserve the order from Schema, too, but for the time being this is easier.

I'll put something in that doesn't expose the option in the UI.

schani:
@dvdsgl It's a global option now, which I like better, and the CLI is nicer (the option doesn't show up for every single language).  Objections to merging?

dvdsgl:
I'm wondering, why should we support alphabetical ordering? If there isn't a good reason, we should make source ordering mandatory.

schani:
We need it for testing.  Once we can preserve order in JSON Schema output and from JSON Schema input we can do away with sorting.  In that context, sorting also makes sense if you generate from JSON Schema, since otherwise you basically get a random order.

dvdsgl:
So, the default behavior is to preserve order? The name of the option is ambiguous to me. Perhaps we could call it `--alphabetize-properties`?

schani:
Yes, JSON order is the default.  I'll change the name.

dvdsgl:
Looks good to me!

PR #329, opened 2017-12-10
Less boilerplate

schani:


schani:
@dvdsgl Please look at this.  We get rid of lots of boilerplate, but at the cost of some type safety.  I'm not super happy with it.  There are some TypeScript features that might make it possible to do this in a type-safe way, but I don't want to spend much time playing around with them.

In particular, there are [namespaces](https://www.sitepen.com/blog/2014/08/22/advanced-typescript-concepts-classes-types/) and [decorators](https://www.typescriptlang.org/docs/handbook/decorators.html).

dvdsgl:
Yes it's a bit funky but I'm not too concerned. I agree we should look for a better metaprogramming approach with decorators, but I'm fine with merging this if you want to.

PR #345, opened 2017-12-18
Mixed directory

schani:


schani:
@dvdsgl This might break the web UI.  Could you take a look?

dvdsgl:
I should enable GraphQL in the web as well.

dvdsgl:
If this changes the index.ts module exports (which it does by changing the source types), just bump the major version and merge regardless.

PR #349, opened 2017-12-20
blockchain.info API is not working - use Reddit

schani:


schani:
@dvdsgl Do you want to adapt this to a different URL?  If not I'll just close it.

dvdsgl:
Let's close it for now, I'll think about it later.

PR #354, opened 2017-12-23
Extension features

schani:
This does three things:

1. Add an API-only option to let the renderer emit leading comments.

2. Implement `just-types` for all renderers where it applies (it makes no sense for Simple Types and JSON Schema).

3. Not emitting packages/namespaces/whatever for `just-types`.

@dvdsgl Objections to 3?


dvdsgl:
Looks good to me!

PR #360, opened 2017-12-24
Swift string enum codables. Fixes #340

### LINKED ISSUE #340: Swift enums declaration
This proposal is about to make all generated enums to be raw representable, specifically String representable.

Currently when we consider JSON schema of enums, e.g. consider schema that defines all film genre types

```
{
  "description": "Genre types",
  "type": "object",
  "properties": {
    "id": {
      "type": "string"
    },
    "genreType": {
      "type": "string",
       "enum" : [
            "Action",
            "Adventure",
            .
            .
            .
        ]
    }
  }
}
```
The generated code looks like this:

```
// To parse the JSON, add this file to your project and do:
//
//   let genre = Genre.from(json: jsonString)!

import Foundation

struct Genre: Codable {
    let id: String?
    let genreType: GenreType?
}

enum GenreType: Codable {
    case action
    case adventure
 
}

// MARK: Top-level extensions -

extension Genre {
    static func from(json: String, using encoding: String.Encoding = .utf8) -> Genre? {
        guard let data = json.data(using: encoding) else { return nil }
        return from(data: data)
    }

    static func from(data: Data) -> Genre? {
        let decoder = JSONDecoder()
        return try? decoder.decode(Genre.self, from: data)
    }

    static func from(url urlString: String) -> Genre? {
        guard let url = URL(string: urlString) else { return nil }
        guard let data = try? Data(contentsOf: url) else { return nil }
        return from(data: data)
    }

    var jsonData: Data? {
        let encoder = JSONEncoder()
        return try? encoder.encode(self)
    }

    var jsonString: String? {
        guard let data = self.jsonData else { return nil }
        return String(data: data, encoding: .utf8)
    }
}

// MARK: Codable extensions -

extension Genre {
        case id = "id"
        case genreType = "genreType"
    }
}

extension GenreType {
    init(from decoder: Decoder) throws {
        let container = try decoder.singleValueContainer()
        if let x = try? container.decode(String.self) {
            switch x {
            case "Action":
                self = .action
                return
            case "Adventure":
                self = .adventure
                return

            default:
                break
            }
        }
        throw DecodingError.typeMismatch(GenreType.self, DecodingError.Context(codingPath: decoder.codingPath, debugDescription: "Wrong type for GenreType"))
    }

    func encode(to encoder: Encoder) throws {
        var container = encoder.singleValueContainer()
        switch self {
        case .action:
            try container.encode("Action")
        case .adventure:
            try container.encode("Adventure")
        }
    }
}

// Helpers

class JSONNull: Codable {
    public init() {
    }

    public required init(from decoder: Decoder) throws {
        let container = try decoder.singleValueContainer()
        if !container.decodeNil() {
            throw DecodingError.typeMismatch(JSONNull.self, DecodingError.Context(codingPath: decoder.codingPath, debugDescription: "Wrong type for JSONNull"))
        }
    }

    public func encode(to encoder: Encoder) throws {
        var container = encoder.singleValueContainer()
        try container.encodeNil()
    }
}
```

`GenreType` enum is not `RawRepresentable`, therefore quicktype needs to define it's own initializers for encoding and decoding. This creates lot of boiler plate code at the moment. 

## Proposed solution
The generated enum cases could be mapped to theirs respective string values from JSON, same way as it's already done for `CodingKeys` enum. At the same time they would directly conform to `Codable` protocol, which comes free for `RawRepresentable` types. There wouldn't be need to define custom decodable and encodable initializers. Codewise, the generated code would be much more clean and readable.

The awaited outcome could look similar to this:

```
import Foundation

struct Genre: Codable {
    let id: String
    let genreType: GenreType
}

enum GenreType: String, Codable {
    case action     = "Action"
    case adventure   = "Adventure"
    .
    .
    .
}

// MARK: Top-level extensions -

extension Genre {
    static func from(json: String, using encoding: String.Encoding = .utf8) -> Genre? {
        guard let data = json.data(using: encoding) else { return nil }
        return from(data: data)
    }

    static func from(data: Data) -> Genre? {
        let decoder = JSONDecoder()
        return try? decoder.decode(Genre.self, from: data)
    }

    static func from(url urlString: String) -> Genre? {
        guard let url = URL(string: urlString) else { return nil }
        guard let data = try? Data(contentsOf: url) else { return nil }
        return from(data: data)
    }

    var jsonData: Data? {
        let encoder = JSONEncoder()
        return try? encoder.encode(self)
    }

    var jsonString: String? {
        guard let data = self.jsonData else { return nil }
        return String(data: data, encoding: .utf8)
    }
}

// MARK: Codable extensions -

extension Genre {
        case id = "id"
        case genreType = "genreType"
    }
}
```

schani:


dvdsgl:
In the case of `just-types`, why is it correct to use non-string-valued enums?

schani:
Because we don't care about JSON-decodability. just-types is for generating types because you need those types, not because you need to decode JSON.

dvdsgl:
Sounds good, merge it!

PR #364, opened 2017-12-26
[NOMERGE] Swift improvements

### LINKED ISSUE #365: JSONNull is being emitted for optional values
```swift
// To parse the JSON, add this file to your project and do:
//
//   let coordinate = try? JSONDecoder().decode(Coordinate.self, from: jsonData)

import Foundation

struct Coordinate: Codable {
    let latitude, longitude: Double?
}

// MARK: Encode/decode helpers

class JSONNull: Codable {
    public init() {}

    public required init(from decoder: Decoder) throws {
        let container = try decoder.singleValueContainer()
        if !container.decodeNil() {
            throw DecodingError.typeMismatch(JSONNull.self, DecodingError.Context(codingPath: decoder.codingPath, debugDescription: "Wrong type for JSONNull"))
        }
    }

    public func encode(to encoder: Encoder) throws {
        var container = encoder.singleValueContainer()
        try container.encodeNil()
    }
}
```

dvdsgl:
# Prototype

https://app2.quicktype.io/#l=swift

---

- [x] Remove separation between type declarations and Codable extensions
- [x] Replace static `from` extensions with failable constructors within primary type definitions
- [x] Implement `dense` option
  - [x] Group contiguous fields with same types
  - [x] Omit `CodingKey` case strings when redundant (e.g. `case x = "x"`)
  - [x] Group contiguous redundant `CodingKey` cases
  - [x] Omit `CodingKeys` enum when all cases are redundant
- [ ] ~~Make `from(url:)` callback-based~~
- [x] Add Swift 4.1 auto-protocols
- [ ] Fix #365

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "quicktype",
-  "version": "7.0.0",
+  "version": "8.0.0",
   "license": "Apache-2.0",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
@@ -46,6 +46,8 @@
     "uglify-js": "^3.0.26",
     "watch": "^1.0.2"
   },
-  "files": ["dist/**"],
+  "files": [
+    "dist/**"
+  ],
   "bin": "dist/cli.js"
 }
--- a/src/ConvenienceRenderer.ts
+++ b/src/ConvenienceRenderer.ts
@@ -18,6 +18,8 @@ import { Renderer, BlankLineLocations } from "./Renderer";
 import { defined, panic, nonNull } from "./Support";
 import { Sourcelike, sourcelikeToSource, serializeRenderResult } from "./Source";
 
+import { trimEnd } from "lodash";
+
 export abstract class ConvenienceRenderer extends Renderer {
     protected globalNamespace: Namespace;
     private _topLevelNames: Map<string, Name>;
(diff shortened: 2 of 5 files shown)


dvdsgl:
@schani I want to emit `case x` rather than `case x = "x"` but I don't know how to compare `Name` to `string`--is this possible?

```swift
this.emitBlock("enum CodingKeys: String, CodingKey", () => {
    this.forEachClassProperty(c, "none", (name, jsonName) => {
        const escaped = stringEscape(jsonName);
        if (name === escaped) { // <-- how do I do this?
            this.emitLine("case ", name);
        } else {
            this.emitLine("case ", name, ' = "', escaped, '"');
        }
    });
});
```

schani:
@dvdsgl You can use `this.sourcelikeToString(name)` for now.

dvdsgl:
I think I may do the convenience constructors and from(url:) in a Decodable extension.

schani:
I have no strong opinions on that.  Code looks good, just remove the "never meant to be committed" commits before you merge.

dvdsgl:
Sadly the extension is a no-go based on the vicissitudes of failable initializers, and how they differ between class and struct. Basically, it's a nightmare and we have to generate a lot more code to make failable initializers work.

test/fixtures.ts
```diff
@@ -532,6 +532,7 @@ export const allFixtures: Fixture[] = [
   new JSONFixture(languages.CPlusPlusLanguage),
   new JSONFixture(languages.ElmLanguage),
   new JSONFixture(languages.SwiftLanguage),
+  new JSONFixture(languages.SwiftDenseLanguage),
```
schani: We have the `quickTestRendererOptions` which only test a subset of the tests for non-default options you can specify, which is probably what you want here.

dvdsgl:
Here's a preview: https://app2.quicktype.io/#l=swift

PR #375, opened 2017-12-28
Use strings for Go enums. Fixes #373

### LINKED ISSUE #373: Enums in Go don’t work under certain circumstances
It seems that if the enum is the value in a map, or structurally contained in the value in a map then it won’t marshal properly, and potentially not unmarshal properly, either.  I filed [an issue](https://github.com/golang/go/issues/23263#issuecomment-354212350) about this, and it’s a known issue but will most likely not be fixed.

Possible fixes:

1. Make enums always be via pointer.
2. Detect the cases where enums won’t work and make them be via pointer only there.
3. Use a reference type for enums, like strings.

Solution 3 would be source-compatible with our current code and very easy to implement.

schani:
Unfortunately we don't check the validity of
the enums anymore, both when marshalling
and when unmarshalling.  If we kept the
custom marshalling methods we could check
there, but they wouldn't get invoked in all
cases, like the integers before, so we'd have
checking in most, but not all cases, and a hard
job of explaining which case applies in which
situation.

### DIFF
--- a/src/Language/Golang.ts
+++ b/src/Language/Golang.ts
@@ -203,59 +203,14 @@ class GoRenderer extends ConvenienceRenderer {
     };
 
     private emitEnum = (e: EnumType, enumName: Name): void => {
-        this.emitLine("type ", enumName, " int");
+        this.emitLine("type ", enumName, " string");
         this.emitLine("const (");
-        let onFirst = true;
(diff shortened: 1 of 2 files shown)


schani:
@dvdsgl Opinions?

PR #381, opened 2017-12-29
[NOMERGE] Iterate objects and maps correctly

### LINKED ISSUE #383: TypeScript runtime typechecking has obvious bugs iterating maps and object keys
The `if (!!prop) continue;` guards are the opposite of what we want.

```ts
   function isValidMap(typ: any, val: any): boolean {
        // all values in the map must be typ
        for (const prop in val) {
            if (!!prop) continue;
            path.push(`["${prop}"]`);
            if (!isValid(typ, val[prop]))
                return false;
            path.pop();
        }
        return true;
    }

    function isValidObject(className: string, val: any): boolean {
        let typeRep = typeMap[className];

        for (const prop in typeRep) {
            if (!!prop) continue;
            path.push(`.${prop}`);
            if (!isValid(typeRep[prop], val[prop]))
                return false;
            path.pop();
        }

        return true;
    }
```

dvdsgl:
Fixes #383

### DIFF
--- a/src/Language/TypeScript.ts
+++ b/src/Language/TypeScript.ts
@@ -308,8 +308,8 @@ function isValidArray(typ: any, val: any): boolean {
 
 function isValidMap(typ: any, val: any): boolean {
     // all values in the map must be typ
-    for (const prop in val) {
-        if (!!prop) continue;
+    for (const prop of Object.keys(val)) {
+        if (prop === undefined) continue;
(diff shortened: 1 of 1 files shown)


schani:
@dvdsgl What's going on with this PR?

dvdsgl:
Sorry, have been too busy to do a thorough job on this. I want to write tests for the runtime typechecking before merging this, and I need to investigate the test failures this change causes.

PR #382, opened 2017-12-29
Fix map inference

### LINKED ISSUE #380: String type equality is broken
The Swift test suite started timing out on Travis. I narrowed it down to the test on `test/inputs/json/misc/ed095.json`, which contains a `[String: String]` that now is generated as a struct with 1.5k fields. This stalls the swift compiler indefinitely. Something about the nullability test in `shouldBeMap` is invalid now.

Run this stalled test with:

```bash
FIXTURE=swift npm test -- test/inputs/json/misc/ed095.json
```

dvdsgl:
- [x] Create a test that fails if map inference fails
- [ ] Fix map inference

Fixes #380


schani:
This is an awfully kludgy way of adding a test.  Couldn't we add a proper fixture that runs quicktype on one JSON file and then uses the generated code to process any number of other JSONs?  It wouldn't be language-specific and we could also compare the output to the input.

dvdsgl:
Yes I know, but I just wanted something immediate so we can fix this fast.

I think we actually want a TypeScript test driver that tests quicktype as a library. For this test I think I'll generate schema and use a schema validator make an assertion in the test driver process. I imagine there are many useful unit tests we could write without running external processes.

dvdsgl:
I've included a fix to TypeScript runtime checking to make the test work, but this must need more scrutiny since it's changing the outcome of other tests. I'll write a different test for map inference.

PR #410, opened 2018-01-05
Disallow property named "url" in Swift. Fixes #409

### LINKED ISSUE #409: String property with name "url" produces incorrect code
```
{
    "url": "abc"
}
```

The generated `struct` has an implicit initializer `init(url: String)`, but we generate an extension with an initializer `init?(url: String)`.

The easiest way to avoid this is to just disallow `url` for property names.  I'm baffled that not a single one of our test cases has such a property.

schani:


dvdsgl:
Would you mind changing the forbidden property name to `fromUrl` and the generated initializer to `init?(fromUrl: String)`? `url` seems like an extremely common property name that we don't want to alias (and don't want to explain why it gets aliased).

PR #425, opened 2018-01-10
Re-add option for classes in Swift. Fixes #424

### LINKED ISSUE #424: Class option has been removed for Swift
https://github.com/quicktype/quicktype/commit/9751ebe52a68dbda85857d5b89ff58d05aba21e4#diff-9eca0ecd13219ea2d5cab7e7fd0de38f has completely removed the option to select classes vs structs. 

Why does this matter? 
I think the choice wether the user wants to have classes vs structs should be with the User rather than only when it's needed to break up cycles.

To give a use case: My Swift Types are separate entities, but have a 1:1 connection to each other. If I extend the struct and later add the other type I can't do that thanks to the struct being a value type.
I think that's only one of many reasons though ...

schani:


caffeineflo:
Looks good to me ;-)

PR #461, opened 2018-01-28
Update Swift.ts error debugging 

Yoloabdo:
Just fixing init with data so it can be useful at debugging, I don't know much about the file syntax or language 😂💯  so you can fix syntax if anything is wrong

### DIFF
--- a/src/Language/Swift.ts
+++ b/src/Language/Swift.ts
@@ -524,11 +524,15 @@ class SwiftRenderer extends ConvenienceRenderer {
                 // 1. Two convenience initializers for Json string and data
                 this.emitBlock(["init?(data: Data)"], () => {
                     this.emitLine(
-                        "guard let me = try? JSONDecoder().decode(",
+                        "do {
+                            self = try JSONDecoder().decode(",
                         this.swiftType(c),
-                        ".self, from: data) else { return nil }"
+                        ".self, from: data)
+                        }catch {
+                            print(error)
+                            return nil
+                        }"
                     );
-                    this.emitLine("self = me");
                 });
                 this.ensureBlankLine();
                 this.emitBlock(["init?(_ json: String, using encoding: String.Encoding = .utf8)"], () => {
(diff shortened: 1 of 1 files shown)


dvdsgl:
Thank you for your PR! What are the goals here? Would you mind creating an issue describing the problems you're having and we can discuss a solution there?

From what I can tell here, it looks like a better approach than printing error information would be to not catch the exception so you could catch it from a higher stack frame and respond more appropriately.

dvdsgl:
Your changes also do not compile. You can compile and test your changes like this:

```bash
$ FIXTURE=swift npm test
```

Please make sure tests pass before submitting a PR.

Yoloabdo:
throwing error here will give you detail about what variable isn't correct when json parsed, sometimes backend misses the response and you stand there not knowing what's the issue since json is always hard to find errors in types or other by just eyes, something like this would appear: 
`DataModel.CodingKeys.facebook], debugDescription: "Expected String value but found null instead.", underlyingError: nil))`
if you're ignoring error you wouldn't easily find it I believe, or that's how I did 

for the compile issue I think I should've submitted an issue not a pull request, but I did anyway

dvdsgl:
Perhaps we should simply not catch exceptions when reading JSON? Do you think this is preferable in Swift, rather than returning `nil`?

Yoloabdo:
catching exception is preferable indeed, check https://developer.apple.com/swift/blog/?id=37 in Writing a JSON Initializer with Error Handling part

Yoloabdo:
I'm totally with adding it as an option like detect enums, it could be 'Detect error' option

dvdsgl:
I think the best option is to allow these initializers to throw exceptions. They should mimic the behavior of `JSONDecoder.decode`, which throws.

Yoloabdo:
I agree with that totally too, you're totally right

dvdsgl:
I've filed an issue to track this: #462

dvdsgl:
@Yoloabdo for now you can just do this:

```swift
let thing = try JSONDecoder().decode(Thing.self, from: jsonString.data(using: .utf8)!)
```

This just decodes the type directly rather than using our convenience initializers, so you will get the exception here.

Issue #462, opened 2018-01-28
Swift convenience initializers should not swallow exceptions from JSONDecoder.decode


dvdsgl:
See discussion at #461 

cc @yoloabdo

Yoloabdo:
Awesome, thank you!

dvdsgl:
@Yoloabdo all fixed, let me know how it goes!

Yoloabdo:
Looks good, but there's failable inits and it has throw too, it shouldn't have failable anymore I guess

Yoloabdo:
for url init I'd suggest to stick with throw 
`        guard let data = json.data(using: encoding) else { throw URLError.badURL }`

dvdsgl:
Would you like to try to implement it? 😄

Yoloabdo:
Sure, let me give it a try 💃 😄

PR #467, opened 2018-01-29
going all 'throw -> Error' instead of failable inits

Yoloabdo:
getting red of fail inits as discussed in [#462](https://github.com/quicktype/quicktype/issues/462#issuecomment-361121302)


dvdsgl:
Again @Yoloabdo, you must run (and pass!) the tests before we can consider your PR:

```bash
$ FIXTURE=swift npm test
```

Because you changed the types of the initializers, the Swift test fixture (`tests/fixtures/swift/main.swift`) no longer compiles, so none of the tests run. Please update this file as well and run the tests.

Also, I think you should update [the comment code](https://github.com/quicktype/quicktype/blob/master/src/Language/Swift.ts#L348) at the top of the Swift output to use `try?` instead of `try`.

src/Language/Swift.ts
```diff
@@ -528,13 +528,13 @@ class SwiftRenderer extends ConvenienceRenderer {
                     this.emitLine("self = try JSONDecoder().decode(", this.swiftType(c), ".self, from: data)");
                 });
                 this.ensureBlankLine();
-                this.emitBlock(["init?(_ json: String, using encoding: String.Encoding = .utf8) throws"], () => {
-                    this.emitLine("guard let data = json.data(using: encoding) else { return nil }");
+                this.emitBlock(["init(_ json: String, using encoding: String.Encoding = .utf8) throws"], () => {
+                    this.emitLine("guard let data = json.data(using: encoding) else { throw URLError(.cannotParseResponse) }");
```
dvdsgl: Can `URLError(.cannotParseResponse)` be written as `URLError.cannotParseResponse`? Why is `URLError.badURL` written differently?

src/Language/Swift.ts
```diff
@@ -512,13 +512,13 @@ class SwiftRenderer extends ConvenienceRenderer {
...
-    guard let data = json.data(using: encoding) else { return nil }
+                this.emitMultiline(`convenience init(_ json: String, using encoding: String.Encoding = .utf8) throws {
+    guard let data = json.data(using: encoding) else { throw URLError(.cannotParseResponse) }
     try self.init(data: data)
 }`);
                 this.ensureBlankLine();
-                this.emitMultiline(`convenience init?(fromURL url: String) throws {
-    guard let url = URL(string: url) else { return nil }
+                this.emitMultiline(`convenience init(fromURL url: String) throws {
+    guard let url = URL(string: url) else { throw URLError.badURL}
```
dvdsgl: Please put a space before the `}`

dvdsgl:
I've added some comments inline.

src/Language/Swift.ts
Yoloabdo: turns out that same applies for both, except that it didn't complain in playground but it did in a project, weird swift

Yoloabdo:
Tests passed 🎉

dvdsgl:
`URLError(.cannotParseResponse)` is not actually the correct error for when a string cannot be encoded as data.

dvdsgl:
@Yoloabdo I substituted `NSError()` for `URLError(.cannotParseResponse)` and fixed the comment and test fixture in #470. Sorry to recreate some of your work there, but thank you for pushing this forward and doing the initial research.

I think this is a big improvement except for `NSError()` containing little useful information` – please let me know if you can think of an improvement or other standard error to use here. I'd rather not define one, although we could do a `fileprivate` custom error if we needed to.

Yoloabdo:
I just found the perfect error for it:
`EncodingError.invalidValue(string, EncodingError.Context.init(codingPath: [], debugDescription: "Can't parse string"))`
a bit long but that's the best I could find internally without having to recreate custom error

PR #470, opened 2018-01-30
Remove failable initializers from Swift

dvdsgl:
Swift convenience initializers now only throw exceptions on error, rather than sometimes throwing and sometimes returning nil.

Supercedes #467


dvdsgl:
It looks like NSError can’t be initialized without parameters on whatever Swift we test on.

PR #481, opened 2018-02-01
[NOMERGE] Objective-C Date Support

dvdsgl:
- [x] Make date inference opt-in
- [x] Support marshaling `NSDate`
- [ ] Add tests

cc @davidhodge

### DIFF
--- a/src/Inference.ts
+++ b/src/Inference.ts
@@ -74,7 +74,11 @@ function canBeEnumCase(s: string): boolean {
 }
 
 export class TypeInference {
-    constructor(private readonly _typeBuilder: TypeBuilder, private readonly _inferEnums: boolean) {}
+    constructor(
+        private readonly _typeBuilder: TypeBuilder,
+        private readonly _inferEnums: boolean,
+        private readonly _inferDates: boolean
+    ) {}
 
     inferType = (
         cjson: CompressedJSON,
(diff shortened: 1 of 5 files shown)


davidhodge:
Hurray!

dvdsgl:
@davidhodge would you mind testing this branch?

dvdsgl:
@davidhodge here's a web preview: https://app2.quicktype.io/#l=objc

davidhodge:
Yes! If my in-flight WiFi works. About to take off.

src/Inference.ts
```diff
@@ -124,13 +128,25 @@ export class TypeInference {
                     unionBuilder.addArray(cjson.getArrayForValue(value));
                     break;
                 case Tag.Date:
-                    unionBuilder.addStringType("date");
+                    if (this._inferDates) {
+                        unionBuilder.addStringType("date");
+                    } else {
```
schani: This is not the way to do it.  We already have the string type mapping, which specifies which string types map to which.  That’s how, for example, in C# `data` and `time` maps to `date-time`, and for all other languages all date/time types map to `string`.  All you have to do is use a string type mapping that maps everything to `string`.

dvdsgl: Ah ok, so I just need to pass the date inference flag into the target language.

schani: I’ll have to take a look at the code. I’d rather not bother every target language with it.

dvdsgl: But then this has to be repeated for each target language? I want an easy way to turn off date inference for any language.

dvdsgl:
@davidhodge good thing quicktype's website works offline 😎

src/Inference.ts
schani: That's what I mean - I want to it to work across all target languages without having to put code into each target language.

schani: Here's where you can override the string type mapping without meddling with target languages:

https://github.com/quicktype/quicktype/blob/master/src/index.ts#L134

dvdsgl:
Closing until I have time to wrap this up.

PR #512, opened 2018-02-11
Rust language support added.

krk:
With help from @schani.

[Preview in browser](https://app2.quicktype.io/#l=rs&s=pokedex)


src/Language/Rust.ts
```diff
@@ -0,0 +1,312 @@
...
+            " = ",
+            this.rustType(t),
+            ";"
+        );
+    };
+
+    protected emitSourceStructure(): void {
+        this.emitLine("extern crate serde_json;");
+        this.emitLine("use std::collections::HashMap;");
+        this.emitLine();
```
schani: Is it a Rust convention to have two blank lines in this position?  If it is, please add a comment, because it's a bit unusual, and since that's why can't use `ensureBlankLine`.

src/Strings.ts
```diff
@@ -9,14 +9,14 @@ import * as _ from "lodash";
 const unicode = require("unicode-properties");
 
 function computeAsciiMap(
-    mapper: (codePoint: number) => string
+    mapper: (codePoint: number, index: number) => string
```
schani: This looks like debugging code that you forgot to take out.

src/Strings.ts
```diff
@@ -81,7 +81,7 @@ function isLowSurrogate(cc: number): boolean {
     return cc >= 0xdc00 && cc <= 0xdfff;
 }
 
-export function utf32ConcatMap(mapper: (codePoint: number) => string): (s: string) => string {
+export function utf32ConcatMap(mapper: (codePoint: number, index: number) => string): (s: string) => string {
```
schani: See comment on `combineWords` above.

src/Language/Rust.ts
```diff
@@ -0,0 +1,312 @@
...
+    const words = splitIntoWords(original);
+    return combineWords(
+        words,
+        legalizeName,
+        allLowerWordStyle,
+        allLowerWordStyle,
+        allLowerWordStyle,
+        allLowerWordStyle,
+        "_",
+        isLetterOrUnderscore
```
schani: `combineWords` can handle start characters, which I think you shouldn't need the index for the start character in `legalizeName`.  But maybe I'm missing something here.

src/Strings.ts
```diff
@@ -128,8 +128,8 @@ export function utf16LegalizeCharacters(isLegal: (utf16Unit: number) => boolean)
     return utf16ConcatMap(u => (isLegal(u) ? String.fromCharCode(u) : ""));
 }
 
-export function legalizeCharacters(isLegal: (codePoint: number) => boolean): (s: string) => string {
-    return utf32ConcatMap(u => (u <= 0xffff && isLegal(u) ? String.fromCharCode(u) : ""));
+export function legalizeCharacters(isLegal: (codePoint: number, index: number) => boolean): (s: string) => string {
```
schani: See `combineWords` comment.

src/Strings.ts
```diff
@@ -356,7 +360,7 @@ export function splitIntoWords(s: string): WordInName[] {
         return i - intervalStart;
     }
 
-    for (;;) {
+    for (; ;) {
```
schani: Whitespace damage.  Please remove.

test/fixtures.ts
```diff
@@ -50,7 +50,7 @@ function jsonTestFiles(base: string): string[] {
     jsonFiles.push(fn);
   }
   let i = 1;
-  for (;;) {
+  for (; ;) {
```
schani: Please remove.

test/fixtures.ts
```diff
@@ -248,7 +248,7 @@ class JSONFixture extends LanguageFixture {
       // Compare fixture.output to fixture.output.expected
       exec(
         `diff -Naur ${this.language.output}.expected ${
-          this.language.output
+        this.language.output
```
schani: Please remove.

src/Language/Rust.ts
```diff
@@ -0,0 +1,312 @@
...
+    private emitBlock = (line: Sourcelike, f: () => void): void => {
+        this.emitLine(line, " {");
+        this.indent(f);
+        this.emitLine("}");
+    };
+
+    private emitUnion = (u: UnionType, unionName: Name): void => {
+        const isMaybeWithSingleType = nullableFromUnion(u);
+
+        if (isMaybeWithSingleType) {
```
schani: Please never use `null` or `undefined` as a boolean - compare explicitly, i.e. `x === null`.  JavaScript will do stupid things like also treat empty strings as `false`, so you think you're checking whether it's `null`, but you're really checking whether it's `null`, or an empty string, or whatever else JS stupidly thinks are falsey.

src/Language/Rust.ts
```diff
@@ -0,0 +1,312 @@
...
+
+        const unionBody = () => this.forEachUnionMember(u, nonNulls, "none", null, (fieldName, t) => {
+            const rustType = this.breakCycle(t, true);
+            this.emitLine([fieldName, "(", rustType, "),"]);
+        });
+
+        this.emitBlock(["pub enum ", unionName], unionBody);
+    }
+
+    emitEnumDefinition = (e: EnumType, enumName: Name): void => {
```
schani: Don't bother fixing it, but for future reference: We don't declare methods like that anymore, i.e. with `=`.  We haven't changed all of them in the codebase yet, sorry.

Short explanation: `=` is slower and uses more memory because it allocates a closure, so in speed-sensitive circumstances we have to use the non-`=` way, and I don't want to have two different ways of defining methods, because that causes confusion is a pain to maintain.

src/Language/Rust.ts
```diff
@@ -0,0 +1,312 @@
...
+        this.emitLine("#[derive(Serialize, Deserialize)]");
+
+        const enumBody = () => this.forEachEnumCase(e, "none", (name, jsonName) => {
+            const escapedName = rustStringEscape(jsonName);
+
+            this.emitLine("#[serde(rename = \"", escapedName, "\")]");
+            this.emitLine([name, ","]);
+        });
+
+        this.emitBlock(["pub enum ", enumName], enumBody);
```
schani: Please put the `() =>` function right in here, to make the flow of control in line with the sequence in the code.

src/Language/Rust.ts
krk: Will remove the extra line here.

src/Language/Rust.ts
krk: I have removed the index parameter and limited the first char to a letter. How can I limit the first character to be a:

* letter or underscore if length(name) > 1
* letter if length(name) == 1

src/Language/Rust.ts
krk: Thank you for the explanation.

src/Language/Rust.ts
krk: Nevermind, I have added a new commit fixing this issue.

dvdsgl:
Please generate comments at the top of the output that explain how to read models from JSON.

src/Language/Rust.ts
```diff
@@ -0,0 +1,306 @@
...
+        this.emitLine(
+            "pub type ",
+            name,
+            " = ",
+            this.rustType(t),
+            ";"
+        );
+    };
+
+    protected emitSourceStructure(): void {
```
dvdsgl: Please generate comments at the top of the output that explain how to read models from JSON.

src/Language/Rust.ts
```diff
@@ -0,0 +1,306 @@
...
+        this.emitLine("extern crate serde_json;");
+        this.emitLine("use std::collections::HashMap;");
```
dvdsgl: Please only include `use std::collections::HashMap;` when needed.

src/Language/Rust.ts
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    private emitStructDefinition = (c: ClassType, className: Name): void => {
+        this.emitLine("#[derive(Serialize, Deserialize)]");
+
+        const structBody = () => this.forEachClassProperty(c, "none", (name, jsonName, prop) => {
+
+            const escapedName = rustStringEscape(jsonName);
+
+            this.emitLine("#[serde(rename = \"", escapedName, "\")]");
```
dvdsgl: I would omit this if `escapedname === jsonName`.

schani: The name of the Rust property is not `jsonName`, so that wouldn't be correct.

dvdsgl: Sorry, yes, I meant if the escapedName matches the property name.

schani: We don't do that in most other languages, either.  Like C#.

schani: What I'm trying to say is: get the code into a good condition, merge, and only then address cosmetics like this.

dvdsgl: Sounds good. I do have some trouble reading the Rust output because there are no empty lines between annotations and the preceding properties, although I’m not sure what the Rust convention is.

schani:
@krk Please see my fixes in https://github.com/quicktype/quicktype/commits/rust

krk:
Thanks for the fixes.

src/Language/Rust.ts
krk: I have addressed this one.

src/Language/Rust.ts
krk: Added usage comments.

src/Strings.ts
```diff
@@ -8,8 +8,7 @@ import * as _ from "lodash";
 
 const unicode = require("unicode-properties");
 
-function computeAsciiMap(
-    mapper: (codePoint: number) => string
+function computeAsciiMap(mapper: (codePoint: number) => string
```
schani: Remove.

src/Strings.ts
```diff
@@ -34,7 +33,7 @@ function precomputedCodePointPredicate(p: CodePointPredicate): CodePointPredicat
     for (let cp = 0; cp < 128; cp++) {
         asciiResults.push(p(cp));
     }
-    return function(cp: number) {
+    return function (cp: number) {
```
schani: Remove.

src/Strings.ts
```diff
@@ -129,7 +128,7 @@ export function utf16LegalizeCharacters(isLegal: (utf16Unit: number) => boolean)
 }
 
 export function legalizeCharacters(isLegal: (codePoint: number) => boolean): (s: string) => string {
-    return utf32ConcatMap(u => (u <= 0xffff && isLegal(u) ? String.fromCharCode(u) : ""));
+    return utf32ConcatMap((u) => (u <= 0xffff && isLegal(u) ? String.fromCharCode(u) : ""));
```
schani: Remove.

src/Strings.ts
krk: What should I remove in this block?

README.md
```diff
@@ -145,6 +145,7 @@ files, URLs, or add other options.
 * `golang` stack
 * `swift` compiler
 * `clang` and Objective-C Foundation (must be tested separately on macOS)
+# `rust` tools
```
Phrohdoh: This should be a bullet-point not a header, right?

schani: Good catch.

src/Strings.ts
schani: The change only reformats the code, unless I'm super blind.

README.md
krk: Certainly, thanks.

Phrohdoh:
Will someone give this the `rust` label for completeness, please?

PR #524, opened 2018-02-14
Make date-time inference optional

dvdsgl:


src/cli.ts
```diff
@@ -233,6 +234,7 @@ function inferOptions(opts: Partial<CLIOptions>): CLIOptions {
         topLevel: opts.topLevel || inferTopLevel(opts),
         noMaps: !!opts.noMaps,
         noEnums: !!opts.noEnums,
+        inferDates: opts.inferDates || false,
```
schani: Please use `!!` for consistency.

src/index.ts
```diff
@@ -96,6 +97,7 @@ const defaultOptions: Options = {
     findSimilarClassesSchema: undefined,
     inferMaps: true,
     inferEnums: true,
+    inferDates: false,
```
schani: Why should this be off by default?

Issue #521, opened 2018-02-13
Emit descriptions in all target languages


schani:
quicktype now supports types and class property descriptions, which should be emitted as comments or doc strings in the generated code. 
 [See how it's done in Go](https://github.com/quicktype/quicktype/blob/d2ede2aa52ca2bbf9829094b3c852aa2b28f99b3/src/Language/Golang.ts#L177).

- [x] C#
- [x] Swift
- [x] TypeScript
- [x] Rust
- [x] Java
- [x] C++
- [x] Objective-C
- [ ] JSON Schema
- [x] Elm
- [x] Simple types

schani:
@krk This shouldn't be more than a few lines of code, in case you're interested.

Dr-Emann:
1. Do we also want to do anything with `title`s as well?
2. As a note, for rust, the comments should be done with doc comments (triple slash), so that the types/fields will get docs based on the description provided.

schani:
@Dr-Emann We already use `title` to name the types.

schani:
JSON Schema is now held up by input rather than output, i.e. the JSON Schema input doesn't completely pull in all the descriptions, or something happens in between.

Elm is held up by Elm: https://github.com/elm-lang/elm-compiler/issues/1678#issuecomment-366843853

schani:
Elm is done, too.  I'm closing this and will open a separate issue for JSON Schema.

PR #529, opened 2018-02-14
Emit descriptions in RustRenderer.

krk:
Added type descriptions and class property descriptions,  https://github.com/quicktype/quicktype/issues/521.

### DIFF
--- a/src/Language/Rust.ts
+++ b/src/Language/Rust.ts
@@ -258,11 +258,21 @@ class RustRenderer extends ConvenienceRenderer {
         }
     }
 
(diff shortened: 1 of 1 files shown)


schani:
I think the question is rather, what's the coding standard for Rust?  It seems like it's standard in Rust to not have those newlines.  From a coding standards repo: https://github.com/rust-lang-nursery/fmt-rfcs/blob/master/example/lists.rs

dvdsgl:
Sorry, yes, that is the question. Let's do what's conventional for Rust.

Issue #516, opened 2018-02-13
Rust improvements


dvdsgl:
- [x] Fix invalid JSON in header comments
- [ ] Optionally emit top-level marshaling functions
- [x] Improve header comments
  - [x] Only show how to deserialize JSON (no need to show going back to string)
  - [x] Don't use extra empty lines between statements
  - [x] Use actual top-level type names in sample
- [x] Optionally omit redundant rename attributes when property name and JSON name match (dense mode)

cc @krk @Phrohdoh

Phrohdoh:
The invalid JSON mentioned is [here](https://github.com/quicktype/quicktype/blob/67635a6feb8af2fc11369bdf1a2ff4e4cc003295/src/Language/Rust.ts#L310) and should be `let json = r#"{"answer": 42}"#;`.

dvdsgl:
@Phrohdoh would you like to submit a PR to improve the header comments? There's a lot of easy wins there.

Ryman:
To add: 
- The serialization code part isn't copy-pasteable currently as it imports `serde_json` but that part actually requires `serde_derive`. It would work if they copied *all* of the code though so maybe that's okay, but I imagine some people may not know that they need to import `serde_derive` in their library code. [Playpen example of it compiling with the change](http://play.integer32.com/?gist=0c2c1aa09c612ce25797972b6b05b07b&version=stable)
- It looks like the header comment improvements can be ticked off

I'd submit a patch for the first point myself but when I cloned the repo and tried running tests without modification they took *way* too long on my machine (with only the rust fixtures). Unsure if that's a problem with my configuration or not but thought it's worth mentioning.

dvdsgl:
@Ryman `FIXTURE=rust npm test` will generate and test a couple hundred Rust programs. Do individual tests take longer than expected or the entire test suite or both?

schani:
@dvdsgl Is it possible that we're downloading and re-compiling all Rust dependencies for each test case?

Ryman:
> @Ryman FIXTURE=rust npm test will generate and test a couple hundred Rust programs. Do individual tests take longer than expected or the entire test suite or both?

It was the entire rust test suit afaik, the first 4 or so tests ran... relatively quick then 20 or so more would take many many minutes more, I don't have access to my laptop where I can run it right now to check though. It definitely felt like it was compiling more than necessary.

dvdsgl:
@Ryman the test slowness is fixed now

Ryman:
[...]

PR #538, opened 2018-02-15
Header comment compacted, fix invalid JSON in header

krk:
Ref https://github.com/quicktype/quicktype/issues/516

### DIFF
--- a/src/Language/Rust.ts
+++ b/src/Language/Rust.ts
@@ -20,6 +20,7 @@ import { UnionType, nullableFromUnion, Type, ClassType, matchType, removeNullFro
 import { Sourcelike, maybeAnnotated } from "../Source";
 import { anyTypeIssueAnnotation, nullTypeIssueAnnotation } from "../Annotation";
 import { EnumOption } from "../RendererOptions";
+import { defined } from "../Support";
 
 enum Density {
     Normal,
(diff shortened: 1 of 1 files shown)


src/Language/Rust.ts
```diff
@@ -332,26 +332,23 @@ class RustRenderer extends ConvenienceRenderer {
     };
 
     protected emitUsageExample(): void {
+        const topLevelName = this.topLevels.keys().next().value;
```
schani: If you want to get the first top level name, use `defined(this.topLevels.keySeq().first())`

src/Language/Rust.ts
```diff
@@ -332,26 +332,23 @@ class RustRenderer extends ConvenienceRenderer {
...
 extern crate serde;
-
 #[macro_use]
 extern crate serde_derive;
-
 extern crate serde_json;
 
-use generated_module::TopLevel;
+use generated_module::` +
+                topLevelName +
```
schani: You can use string interpolation here, i.e. `${topLevelName}`

dvdsgl:
@krk could the header comments be line comments, each starting with `// ` as we use in other languages? Would this be conventional for Rust?

krk:
Line comments are preferred in API docs according to this RFC: https://github.com/rust-lang/rfcs/blob/master/text/0505-api-comment-conventions.md#use-line-comments

PR #551, opened 2018-02-16
[NO MERGE] Refactor union builder

schani:


schani:
Incorporated into #562

PR #580, opened 2018-02-20
Include postman item names and request urls in descriptions

dvdsgl:
For example:

```csharp
    /// <summary>
    /// Response Status Code
    /// 
    /// GET https://postman-echo.com/status/200
    /// 
    /// This endpoint allows one to instruct the server which status code to respond with.
    /// 
    /// Every response is accompanied by a status code. The status code provides a summary of the
    /// nature of response sent by the server. For example, a status code of `200` means
    /// everything is okay with the response and a code of `404` implies that the requested URL
    /// does not exist on server.
    /// A list of all valid HTTP status code can be found at the [List of Status
    /// Codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) wikipedia article. When
    /// using Postman, the response status code is described for easy reference.
    /// 
    /// Note that if an invalid status code is requested to be sent, the server returns a status
    /// code of `400 Bad Request`.
    /// </summary>
    public partial class ResponseStatusCode
    {
        [JsonProperty("status")]
        public long Status { get; set; }
    }
```

### DIFF
--- a/src/PostmanCollection.ts
+++ b/src/PostmanCollection.ts
@@ -1,7 +1,6 @@
 "use strict";
 
 import { JSONTypeSource } from ".";
-import { isObject } from "lodash";
 
 function isValidJSON(s: string): boolean {
     try {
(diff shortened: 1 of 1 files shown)


src/PostmanCollection.ts
```diff
@@ -1,6 +1,7 @@
 "use strict";
 
 import { JSONTypeSource } from ".";
+import { isObject } from "lodash";
```
schani: `typeof x === "object"`

schani:
Please squash these before merging.

PR #587, opened 2018-02-21
Refactor TargetLanguage

schani:
This makes it easier to extend target languages.


schani:
Included in JavaScript

PR #591, opened 2018-02-22
Support nested namespaces in C++

schani:
The option can be used like `--namespace a::b::c`


kirbysayshi:
Looks good and generated code compiled!

PR #611, opened 2018-02-28
Fix duplicate types

schani:


schani:
I solved this differently.

PR #645, opened 2018-03-08
Schema fixes

schani:


schani:
The `vega-lite.schema` output has changed, but that seems to be an improvement.  Also new test cases, which I guess we should not mark as errors in the diff.

PR #653, opened 2018-03-10
Foreign schema refs

schani:


schani:
The diff is due to a new file.

Issue #655, opened 2018-03-10
Add option for public fields in Rust structs


jkcclemens:
Currently, structs like the one below are generated.

```rust
#[derive(Serialize, Deserialize)]
pub struct Report {
    fights: Vec<ReportFight>,
    lang: String,
    friendlies: Vec<Friendly>,
    enemies: Vec<Enemy>,
    #[serde(rename = "friendlyPets")]
    friendly_pets: Vec<Enemy>,
    #[serde(rename = "enemyPets")]
    enemy_pets: Vec<serde_json::Value>,
    phases: Vec<serde_json::Value>,
    title: String,
    owner: String,
    start: i64,
    end: i64,
    zone: i64,
}
```

However, it is often desirable to have each field public, like so:

```rust
#[derive(Serialize, Deserialize)]
pub struct Report {
    pub fights: Vec<ReportFight>,
    pub lang: String,
    pub friendlies: Vec<Friendly>,
    pub enemies: Vec<Enemy>,
    #[serde(rename = "friendlyPets")]
    pub friendly_pets: Vec<Enemy>,
    #[serde(rename = "enemyPets")]
    pub enemy_pets: Vec<serde_json::Value>,
    pub phases: Vec<serde_json::Value>,
    pub title: String,
    pub owner: String,
    pub start: i64,
    pub end: i64,
    pub zone: i64,
}
```

An option to toggle making all fields public would be great.

dvdsgl:
This would be a trivial option to add, and I estimate would take you less than 10 minutes to implement if you're interested.

1. [Here's the line to edit](https://github.com/quicktype/quicktype/blob/master/src/Language/Rust.ts#L278) in the Rust renderer.
2. Add an option just like the [density option](https://github.com/quicktype/quicktype/blob/master/src/Language/Rust.ts#L31).

jkcclemens:
Ooh, great! I peeked in, but TypeScript is not something I know very well, so I got lost. I'll take a crack at it.

dvdsgl:
Godspeed! If you can write Rust, you can definitely handle TypeScript 😆

PR #656, opened 2018-03-10
Add visibility option on Rust fields

jkcclemens:
#655 

This adds an option to control visibility modifiers on generated struct fields in Rust.

I don't know if I need to edit anything elsewhere or if this is good, but it appears to work from my tests!

### DIFF
--- a/src/Language/Rust.ts
+++ b/src/Language/Rust.ts
@@ -27,10 +27,22 @@ enum Density {
     Dense
 }
 
+enum Visibility {
+    Private,
+    Crate,
+    Public
+}
+
 export default class RustTargetLanguage extends TargetLanguage {
     private readonly _denseOption = new EnumOption("density", "Density", [
         ["normal", Density.Normal],
         ["dense", Density.Dense]
+    ]);
+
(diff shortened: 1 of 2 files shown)


src/Language/Rust.ts
```diff
@@ -266,6 +283,15 @@ class RustRenderer extends ConvenienceRenderer {
         }
     }
 
+    private visibility(): string {
```
dvdsgl: Great work! if you put `get` before the method name, this becomes a property and you can refer to it as `this.visibility`.

dvdsgl:
This looks great, @jkcclemens. I added one small comment.

src/Language/Rust.ts
jkcclemens: `get_visibility`, `getVisibility`, or `getvisibility`?

dvdsgl: `get visibility()`

dvdsgl:
I added the option to our test harness as well.

dvdsgl:
Thank you for your PR, @jkcclemens! This will be available npm momentarily, and live on app.quicktype.io in 10 minutes.

PR #660, opened 2018-03-12
Tidier Types module

dvdsgl:
* Omit synthetic StrictDouble when not used
* Align enum declarations


dvdsgl:
This is not even close to correct 🙃

PR #664, opened 2018-03-13
Add an option for Rust to derive Debug impls.

waywardmonkeys:


schani:
The `emitLine` that produces the `derive` should be put into its own method eventually, to avoid the repetition.

waywardmonkeys:
I thought about doing that, but would be great to derive `Default` for structs one day.

dvdsgl:
"Derive Debug impl" — is that the description a Rust user would recognize? "Impl" looks odd.

waywardmonkeys:
I think so ... since in Rust, it would be `impl Debug for ... { ... }`

schani:
@waywardmonkeys What does `Default` do, and why can't we derive it now?

waywardmonkeys:
@schani `Default` lets you easily construct an instance of a struct or enum:

```rust
let f = Whatever::default();
```

But for that to work, all fields of a struct must have a default value, which is fine for everything that is `Option` since that would just be `None`, but if you have a value that isn't an `Option`, then it needs to have `Default` implemented for it, like any enumerated value.

I'm not sure entirely yet what would be possible here on the generating side as I haven't dug in much deeper yet other than having tried it, seen how and why it failed, and then removing it for now.

PR #669, opened 2018-03-14
Convenience initializers disabled by default in when outputting Swift

rjstelling:
Convenience initializers are no longer enabled by default, you have to pass `--initializers` when outputting Swift.

This confused, me and I could not find any documentation on how to disable the convenience initializers.

This changes default behaviour so adding a `--no-initializers` option might be a better resolution.

### DIFF
--- a/src/Language/Swift.ts
+++ b/src/Language/Swift.ts
@@ -49,7 +49,7 @@ export default class SwiftTargetLanguage extends TargetLanguage {
     private readonly _convenienceInitializers = new BooleanOption(
         "initializers",
         "Convenience initializers",
-        true
+        false
     );
 
     private readonly _alamofireHandlers = new BooleanOption("alamofire", "Alamofire extensions", false);


dvdsgl:
By default, we want the code to be trivial to use for loading JSON. What is your reason for disabling this by default, beyond you not wanting to use the feature?

dvdsgl:
This also breaks the generated code for all current users, so I'm unlikely to accept this without a compelling explanation that considers the tradeoffs.

schani:
@dvdsgl I think the problem here is that you can only set boolean options on the command line, not unset them.  So a boolean option that's on by default can never be turned off.

rjstelling:
@dvdsgl, @schani Is correct there is no way to unset a boolean option on the command line and I need to use quicktype without the initializers.

Breaking the code for other users isn't really an option. Would you consider accepting a `--no-initializers` command line option?

dvdsgl:
Oh I see! We will fix that ASAP.

rjstelling:
I will close this PR as @dvdsgl has indicated this is a bug that will be fixed.

PR #674, opened 2018-03-15
CLI Telemetry

dvdsgl:
Telemetry is off by default and controlled with:

```bash
$ quicktype --telemetry enable|disable
```

Until the user enables or disables telemetry, this header is added to files produced by the quicktype CLI:

```
Please help improve quicktype by enabling anonymous telemetry with:

  $ quicktype --telemetry enable

You can also enable telemetry on any quicktype invocation:

  $ quicktype https://blockchain.info/latestblock -o Block.cs --telemetry enable

This helps us improve quicktype by measuring:

  * How many people use quicktype?
  * What features are popular or unpopular?
  * How does quicktype perform? Are there any errors?

quicktype does not collect:

  * Your filenames or input data
  * Any personally identifiable information (PII)
  * Anything not directly related to quicktype's usage

If you don't want to help us, you can dismiss this message with:

  $ quicktype --telemetry disable

For a full privacy policy, please see https://app.quicktype.io/privacy
```

## Information collected

- [x] Duration of quicktype invocation
- [x] Number and target language of invocations
- [ ] quicktype exceptions

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -17,7 +17,8 @@ import {
     isSchemaSource,
     quicktypeMultiFile,
     SerializedRenderResult,
-    TargetLanguage
+    TargetLanguage,
+    languageNamed
 } from "..";
 
 import { OptionDefinition } from "../RendererOptions";
@@ -32,13 +33,14 @@ import { getStream } from "../get-stream/index";
 import { train } from "../MarkovChain";
 import { sourcesFromPostmanCollection } from "../PostmanCollection";
 import { readableFromFileOrURL, readFromFileOrURL, FetchingJSONSchemaStore } from "./NodeIO";
+import * as telemetry from "./telemetry";
 
 const commandLineArgs = require("command-line-args");
 const getUsage = require("command-line-usage");
 const chalk = require("chalk");
 const wordWrap: (s: string) => string = require("wordwrap")(90);
 
-const packageJSON = require("../package.json");
+const packageJSON = require("../../package.json");
 
 export interface CLIOptions {
     lang: string;
@@ -69,6 +71,7 @@ export interface CLIOptions {
     quiet: boolean;
     version: boolean;
     debug?: string;
+    telemetry?: string;
 }
 
 async function sourceFromFileOrUrlArray(name: string, filesOrUrls: string[]): Promise<JSONTypeSource> {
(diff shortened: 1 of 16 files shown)


src/cli/telemetry.ts
```diff
@@ -0,0 +1,155 @@
...
+    event(category: string, action: string, label?: string, value?: string | number) {
+        this.analytics.event(category, action, label, value);
+    }
+
+    enable() {
+        Storage.analyticsState = "enabled";
+        this.analytics = new GoogleAnalytics();
+    }
+
+    disable() {
```
schani: Please declare functions that don't return values as `void`.

src/cli/telemetry.ts
```diff
@@ -0,0 +1,155 @@
...
+    static get<T>(name: string, def: T): T {
+        return storage.getItemSync(name) || def;
+    }
+
+    static set<T>(name: string, val: T) {
+        storage.setItemSync(name, val);
+    }
+
+    static get userId(): string {
+        return Storage.get("userId", uuid());
```
schani: This seems weird.  `uuid()` generates a new UUID each time, right?  So if the user ID isn't set, this will return a different one each time?

Also I'd like to use all-caps for acronyms.

src/cli/telemetry.ts
```diff
@@ -0,0 +1,155 @@
...
+
+export const telemetry = new Telemetry();
+
+export const TELEMETRY_HEADER = `Please help improve quicktype by enabling anonymous telemetry with:
+
+  $ quicktype --telemetry enable
+
+You can also enable telemetry on any quicktype invocation:
+
+  $ quicktype https://blockchain.info/latestblock -o Block.cs --telemetry enable
```
schani: Is it necessary to include this example?

src/cli/telemetry.ts
dvdsgl: Oops, I forgot that I removed the code that sets the default value in `get`, well restore.

schani: `get` must not have side effects. The debugger will feel free to call it anytime, and it will screw you up.

dvdsgl: The method is just called `get`, it's not a property.

src/cli/index.ts
```diff
@@ -221,11 +225,15 @@ export function inferCLIOptions(opts: Partial<CLIOptions>): CLIOptions {
         srcLang = withDefault<string>(srcLang, "json");
     }
 
+    const lang = opts.lang || inferLang(opts);
+    const language = languageNamed(lang);
+    assert(language !== undefined, `Unsupported output language: ${lang}`);
```
schani: Better:

```typescript
if (language === undefined) {
    return panic("blabla");
}
```

Then TS can infer that `language !== undefined`, and the `defined` below isn't necessary anymore.

src/cli/storage.ts
```diff
@@ -0,0 +1,31 @@
+import { getItemSync, setItemSync, init } from "node-persist";
+import { TelemetryState } from "./telemetry";
+
+
```
schani: Only one newline.  Please do `npm run tslint` to lint.

src/cli/storage.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export class Storage {
```
schani: What's the point of this class again?  If you want the functions in a namespace you can always do `import * as Storage from "Storage"`.

PR #676, opened 2018-03-17
[NOMERGE] Simplify how urijs types are included

dvdsgl:
- [x] Distribute the urijs.d.ts file as a simple typings file under `src/`
- [ ] Figure out the kooky `uri.URI` type in `JSONSchemaInput.ts` because it's confusing to TypeScript consumers of quicktype


dvdsgl:
I have a tenuous grasp of how these typing files get included in our distribution.

dvdsgl:
This didn't work.

PR #677, opened 2018-03-17
[NO MERGE] Add urijs types in the distribution

schani:


dvdsgl:
This feels really hacky.

dvdsgl:
It also does not work. Is there a different approach we can take? We've tried three ways to distribute these stupid types and none has worked.

schani:
Ok, I'll see what happens if I just put the `@types` directory into a repo with a `package.json`.

schani:
This is not it.

PR #681, opened 2018-03-18
Refactor CLI option parsing to make it externally accessible

schani:


schani:
@dvdsgl Are you ok with this?

PR #683, opened 2018-03-18
Boolean options are named 'x' by default

dvdsgl:
and get an automatic 'no-x' variant on the CLI

image

## Example

```bash
$ quicktype getting-started.json -l swift --alamofire --no-initializers
```

Produces:

```swift
import Foundation
import Alamofire

struct GettingStarted: Codable {
    let greeting: String
    let instructions: [String]
}

// MARK: - Alamofire response handlers

extension DataRequest {
    fileprivate func decodableResponseSerializer<T: Decodable>() -> DataResponseSerializer<T> {
        return DataResponseSerializer { _, response, data, error in
            guard error == nil else { return .failure(error!) }

            guard let data = data else {
                return .failure(AFError.responseSerializationFailed(reason: .inputDataNil))
            }

            return Result { try JSONDecoder().decode(T.self, from: data) }
        }
    }

    @discardableResult
    fileprivate func responseDecodable<T: Decodable>(queue: DispatchQueue? = nil, completionHandler: @escaping (DataResponse<T>) -> Void) -> Self {
        return response(queue: queue, responseSerializer: decodableResponseSerializer(), completionHandler: completionHandler)
    }

    @discardableResult
    func responseGettingStarted(queue: DispatchQueue? = nil, completionHandler: @escaping (DataResponse<GettingStarted>) -> Void) -> Self {
        return responseDecodable(queue: queue, completionHandler: completionHandler)
    }
}
```

### DIFF
--- a/src/Language/JavaScript.ts
+++ b/src/Language/JavaScript.ts
@@ -22,7 +22,7 @@ import { BooleanOption, Option } from "../RendererOptions";
 const unicode = require("unicode-properties");
 
 export class JavaScriptTargetLanguage extends TargetLanguage {
-    protected readonly runtimeTypecheck = new BooleanOption(
+    protected readonly _runtimeTypecheck = new BooleanOption(
         "runtime-typecheck",
         "Verify JSON.parse results at runtime",
         true
@@ -37,7 +37,7 @@ export class JavaScriptTargetLanguage extends TargetLanguage {
     }
 
     protected getOptions(): Option<any>[] {
-        return [this.runtimeTypecheck];
+        return [this._runtimeTypecheck];
     }
 
     get supportsOptionalClassProperties(): boolean {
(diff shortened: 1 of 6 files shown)


src/Language/JavaScript.ts
```diff
@@ -22,10 +22,10 @@ import { BooleanOption, Option } from "../RendererOptions";
 const unicode = require("unicode-properties");
 
 export class JavaScriptTargetLanguage extends TargetLanguage {
-    protected readonly omitRuntimeTypecheck = new BooleanOption(
-        "no-runtime-typecheck",
-        "Don't verify JSON.parse results at runtime",
-        false
+    protected readonly _runtimeTypecheck = new BooleanOption(
```
schani: Underscore only for `private` members.

PR #698, opened 2018-03-23
Remove a weird TypeReconstituter argument

schani:


schani:
I guess we need the weird argument.

PR #701, opened 2018-03-24
Emit superclass - needs testing and review

bitsol:
The magic sauce:
https://github.com/json-helpers/quicktype/blob/emit-superclass/src/Language/CSharp.ts#L271

### DIFF
--- a/src/Language/CSharp.ts
+++ b/src/Language/CSharp.ts
@@ -265,7 +265,11 @@ export class CSharpRenderer extends ConvenienceRenderer {
                 break;
         }
         this.emitDescription(description);
-        this.emitLine(declaration, " ", name);
+        if (superclass === undefined) {
+            this.emitLine(declaration, " ", name);
+        } else {
+            this.emitLine(declaration, " ", name, " : ", superclass);
+        }
         this.emitBlock(emitter);
     }
 
@@ -288,7 +292,7 @@ export class CSharpRenderer extends ConvenienceRenderer {
             AccessModifier.Public,
             "partial class",
             className,
-            " : " + this.superclassForClass(c),
+            this.superclassForClass(c),
             () => {
                 if (c.properties.isEmpty()) return;
                 const blankLines = this.needAttributes && !this.dense ? "interposing" : "none";
--- a/src/Language/CSharp.ts
+++ b/src/Language/CSharp.ts
@@ -243,7 +243,7 @@ export class CSharpRenderer extends ConvenienceRenderer {
             return csType;
         }
     }
-    protected superclassForClass(): Sourcelike | undefined {
+    protected superclassForType(_t: Type): Sourcelike | undefined {
         return undefined;
     }
     protected emitType(
@@ -292,7 +292,7 @@ export class CSharpRenderer extends ConvenienceRenderer {
             AccessModifier.Public,
             "partial class",
             className,
-            this.superclassForClass(),
+            this.superclassForType(c),
             () => {
                 if (c.properties.isEmpty()) return;
                 const blankLines = this.needAttributes && !this.dense ? "interposing" : "none";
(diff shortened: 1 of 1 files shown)


schani:
There are compile errors here.  Before submitting a PR, please make sure that it at least compiles without errors on your machine with `npm run build`.

src/Language/CSharp.ts
```diff
@@ -243,12 +243,15 @@ export class CSharpRenderer extends ConvenienceRenderer {
             return csType;
         }
     }
-
+    protected superclassForClass(): Sourcelike | undefined {
```
schani: This should take a `ClassType` as an argument (that's why it's called `xxxForClass`).  In fact, come to think of it, you might want superclasses for non-class types, too, like for unions.  So let's instead make it

```csharp
    protected superclassForType(_t: Type): Sourcelike | undefined
```

src/Language/CSharp.ts
```diff
@@ -280,51 +287,58 @@ export class CSharpRenderer extends ConvenienceRenderer {
...
-                    this.emitDescription(description);
-                    this.emitLine(property);
-                } else if (this.dense && attribute !== undefined) {
-                    const comment = description === undefined ? "" : ` // ${description.join("; ")}`;
-                    columns.push([attribute, " ", property, comment]);
-                } else {
-                    this.emitDescription(description);
-                    if (attribute !== undefined) {
-                        this.emitLine(attribute);
+        this.emitType(
```
schani: Any chance you could make this change without having to reformat the whole method? It makes it really hard to figure out what's going on in this commit if it contains changes that just reformat code.

src/Language/CSharp.ts
```diff
@@ -280,51 +287,58 @@ export class CSharpRenderer extends ConvenienceRenderer {
...
+            this.descriptionForType(c),
+            AccessModifier.Public,
+            "partial class",
+            className,
+            this.superclassForClass(),
```
schani: `this.superclassForType(c)`

src/Language/CSharp.ts
```diff
@@ -280,51 +287,58 @@ export class CSharpRenderer extends ConvenienceRenderer {
...
+                }
             }
-        });
+        );
     }
 
     private emitUnionDefinition(u: UnionType, unionName: Name): void {
         const nonNulls = removeNullFromUnion(u)[1];
-        this.emitType(this.descriptionForType(u), AccessModifier.Public, "partial struct", unionName, () => {
+        this.emitType(this.descriptionForType(u), AccessModifier.Public, "partial struct", unionName, undefined, () => {
```
schani: Now here we can use `this.superclassForType(u)`

src/Language/CSharp.ts
schani: We're using [prettier](https://github.com/prettier/prettier) to format TypeScript.

src/Language/CSharp.ts
```diff
@@ -522,7 +536,7 @@ export class NewtonsoftCSharpRenderer extends CSharpRenderer {
             typeKind = "class";
         }
         const csType = this.csType(t);
-        this.emitType(undefined, AccessModifier.Public, [partial, typeKind], name, () => {
+        this.emitType(undefined, AccessModifier.Public, [partial, typeKind], name, this.superclassForClass(), () => {
```
schani: `this.superclassForType(t)`

src/Language/CSharp.ts
```diff
@@ -537,6 +551,7 @@ export class NewtonsoftCSharpRenderer extends CSharpRenderer {
             AccessModifier.None,
             "static class",
             defined(this._enumExtensionsNames.get(enumName)),
+            this.superclassForClass(),
```
schani: `this.superclassForType(e)`

src/Language/CSharp.ts
```diff
@@ -699,7 +714,7 @@ export class NewtonsoftCSharpRenderer extends CSharpRenderer {
 
     private emitSerializeClass(): void {
         // FIXME: Make Serialize a Named
-        this.emitType(undefined, AccessModifier.Public, "static class", "Serialize", () => {
+        this.emitType(undefined, AccessModifier.Public, "static class", "Serialize", this.superclassForClass(), () => {
```
schani: This class does not correspond to any JSON type - it doesn't need a superclass, so just pass `undefined`.

src/Language/CSharp.ts
```diff
@@ -755,7 +770,7 @@ export class NewtonsoftCSharpRenderer extends CSharpRenderer {
         // FIXME: Make Converter a Named
         let converterName: Sourcelike = ["Converter"];
         if (jsonConverter) converterName = converterName.concat([": JsonConverter"]);
-        this.emitType(undefined, AccessModifier.Internal, "class", converterName, () => {
+        this.emitType(undefined, AccessModifier.Internal, "class", converterName, this.superclassForClass(), () => {
```
schani: Same here I think.

src/Language/CSharp.ts
bitsol: I'll switch the method around, then.

schani: Switch it around?  I'm not sure what that means.

bitsol: Sorry. I meant change the arguments and name.

PR #708, opened 2018-03-28
Csharp optional

schani:


schani:
@dvdsgl I have a feeling that some people might prefer our current way of not distinguishing between optional and nullable.  Should we make this into an option?

dvdsgl:
I can barely understand this distinction. I don't think we should make users reason about it.

schani:
This is the code we generate with this change:

```c#
        [JsonProperty("in", Required = Required.Always, NullValueHandling = NullValueHandling.Ignore)]
         public string In { get; set; }
```

PR #718, opened 2018-03-31
Messages API

schani:


schani:
@dvdsgl Please take a look.

src/Messages.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export function messageError(message: ErrorMessage): never;
+export function messageError(message: ErrorMessageWithInfo, info: string): never;
+export function messageError(message: ErrorMessageWithTwoInfos, info1: string, info2: string): never;
+export function messageError(message: string, ...infos: string[]): never {
+    for (const info of infos) {
+        // FIXME: this won't work if one of the infos, apart from the last, contains "%s"
+        message = message.replace("%s", info);
+    }
+    return panic(message);
```
dvdsgl: It doesn't look like I will be able to compare these messages to the enum cases and present custom UI based on that. I would prefer that the enum cases are recoverable from the caught error.

schani: Yes, I didn't implement the actual exception type yet, I focused on the API for reporting first.

We can use the enum reverse mapping to get to the name of enum case:

```typescript
enum Enum {
    A
}
let a = Enum.A;
let nameOfA = Enum[a]; // "A"
```

schani: We could have an exception class that has

* The name of the enum case.
* The "raw" message, with the formatting characters.
* The values to substitute.
* A method that returns the full substituted message.

Would that be enough?

dvdsgl: Yes that's great.

schani: The error should also contain an optional location.

schani:
Once we're on TypeScript 2.8 we can use a type-safe API like this.  For now we should use an API that looks the same but is not type-safe.

```typescript
type Errors =
    { kind: "foo"; props: { quux: number } } |
    { kind: "bar"; props: { frob: boolean } };
type Kind = Errors extends { kind: infer T } ? T : never;
type Props = Errors extends { props: infer P } ? P : never;
type KindFor<P> = Extract<Errors, { props: P }> extends { kind: infer K } ? K : never;

function error<P extends Props>(kind: KindFor<P>, props: P): void {}

error("foo", { quux: 123 });
```

PR #721, opened 2018-03-31
TypeScript as a source language

dvdsgl:
image


dvdsgl:
@schani I don't think this would ever work in the browser, but would love to ship it in the CLI and enable in the web app via the cloud API.

schani:
Awesome!  I was thinking about something like this.

About the web thing: The TS compiler API allows the user to specify where the source is coming from.  That can be just a string in memory.  The only problem then is to expose that in `typescript-json-schema`.  Maybe it's already exposed, but in any case, it's probably not a big deal, and we should send them a PR for that.

schani:
There should be at least one test case for TS input.

dvdsgl:
## Testing Strategy

1. From a JSON sample `top.json`, use quicktype to generate `top.ts` with it.
1. From `top.ts`, generate `top-ts-src.ts` with `quicktype -s typescript`.
1. Test that `top-ts-src.ts` can parse and emit `top.json`.

schani:
I checked the diff, and it's fine.  @dvdsgl merge if you're happy.

PR #727, opened 2018-04-01
Add TypeScriptTypeSource

dvdsgl:


src/cli/index.ts
```diff
@@ -109,14 +112,14 @@ async function samplesFromDirectory(dataDir: string, topLevelRefs: string[] | un
             }
 
             if (file.endsWith(".url") || file.endsWith(".json")) {
-                // FIXME: Why do we include the URI here?
                 sourcesInDir.push({
+                    kind: "json",
                     name,
-                    uri: fileOrUrl,
```
dvdsgl: @schani I have no idea if this is used. Seems like it was not.

src/cli/index.ts
```diff
@@ -152,25 +160,34 @@ async function samplesFromDirectory(dataDir: string, topLevelRefs: string[] | un
...
-        if (jsonSamples.length > 0 && schemaSources.length + graphQLSources.length > 0) {
-            return panic("Cannot mix JSON samples with JSON Schema or GraphQL in input subdirectory");
+
+        if (jsonSamples.length > 0 && schemaSources.length + graphQLSources.length + typeScriptSources.length > 0) {
+            return panic("Cannot mix JSON samples with JSON Schems, GraphQL, or TypeScript in input subdirectory");
         }
-        if (schemaSources.length > 0 && graphQLSources.length > 0) {
-            return panic("Cannot mix JSON Schema with GraphQL in an input subdirectory");
+
+        const oneUnlessEmpty = <T>(xs: T[]) => (xs.length === 0 ? 0 : 1);
```
schani: No reason for this to be generic.  Also, can be written as `Math.sign(xs.length)`

src/cli/index.ts
```diff
@@ -152,25 +160,34 @@ async function samplesFromDirectory(dataDir: string, topLevelRefs: string[] | un
...
+        if (oneUnlessEmpty(schemaSources) + oneUnlessEmpty(graphQLSources) + oneUnlessEmpty(typeScriptSources) > 1) {
```
schani: Actually, this can just be written as `schemaSources.length + graphQLSources.length + typeScriptSource.length > 1`

src/index.ts
```diff
@@ -49,37 +50,63 @@ export type RendererOptions = { [name: string]: string };
 export type StringInput = string | Readable;
 
 export interface JSONTypeSource {
+    kind: "json";
```
schani: This breaks the API, so the major version needs an increment.

src/cli/index.ts
```diff
@@ -152,25 +160,34 @@ async function samplesFromDirectory(dataDir: string, topLevelRefs: string[] | un
...
             if (isJSONSource(source)) {
                 jsonSamples = jsonSamples.concat(source.samples);
             } else if (isSchemaSource(source)) {
                 schemaSources.push(source);
-            } else {
+            } else if (isGraphQLSource(source)) {
                 graphQLSources.push(source);
+            } else if (isTypeScriptSource(source)) {
+                typeScriptSources.push(source);
             }
```
schani: If you'd be so kind you could add a final `else` branch with `return panic("Unrecognized source");`

schani:
Before merging I'd like to rename the `WIP` commits in the `typescript-input` branch, and look it over again real quick.

src/cli/index.ts
dvdsgl: So we always only have one of these sources?

src/cli/index.ts
schani: Weird.  It wasn't even in the type definition.

src/cli/index.ts
schani: Oh, sorry, ignore that comment.

src/cli/index.ts
dvdsgl: I either have to do this, or make `xs: any` as far as the compiler tells me.

schani: Really?  It can't be `any[]`?  That's weird.

dvdsgl: oh yes I think that will work

src/cli/index.ts
schani: Leave it for now, but in cases like this, where the compiler knows which values are possible, it's better to do

```
return assertNever(source.kind);
```

instead, which is a static check.

schani:
I prettied up the `typescript-input` branch so you'll have to rebase.  Other than that I'm happy to merge.  Please check that the diff only contains new files, just to be sure.  I really need to implement it not failing if there are only new files.

PR #749, opened 2018-04-04
 Make rust anyType optional, fixes #702

### LINKED ISSUE #702: Rust fails on optional "any" properties
Given this schema

```
{
  "type": "object",
  "properties": {
    "foo": {}
  }
}
```

and the input `{}`, Rust fails:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ErrorImpl { code: Message("missing field `foo`"), line: 1, column: 2 }', src/libcore/result.rs:906:4
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

Test case in this branch: https://github.com/quicktype/quicktype/tree/optional-any

krk:


schani:
Thank you!

PR #763, opened 2018-04-06
JavaScript optional vs nullable. Fixes #740

### LINKED ISSUE #740: Distinguish between optional and nullable in TypeScript output
Apart from being more correct, it would allow us to run stricter tests for TypeScript input, because we do JSON -> TypeScript -> C#, and we lose optional vs nullable in the first step.

schani:


schani:
@dvdsgl Any objections to this?

PR #768, opened 2018-04-07
Put error message name into exception. Fixes #766

### LINKED ISSUE #766: Make ErrorMessage reverse-lookup-able
Either figure out how it can be done with the enum, or make it into a class?

schani:


schani:
@dvdsgl Look at this, please.  The "symbolic" name of the exception is now in `messageName`.

I can still make this type-safe in TS 2.8.

PR #770, opened 2018-04-07
Multiple JSON Schema inputs.  Fixes #711

### LINKED ISSUE #711: Allow multiple schema inputs
If we allow any number of schema URLs, including wildcard ones, then we don't need the `add-schema-top-level` option anymore.  We could instead do:

```
quicktype --src-lang schema ./schema.json#/definitions/One ./schema.json#/definitions/Two
```

or, to include all definitions

```
quicktype --src-lang schema ./schema.json#/definitions/
```

schani:


dvdsgl:
Looks great.

PR #772, opened 2018-04-08
Full object support for JS languages.  Fixes #764

### LINKED ISSUE #764: Validate full object types in JavaScript output

schani:
Unify object type representation, and make a general
reference "type" so we can represent all kinds of cycles
later on.

### DIFF
--- a/src/Language/TypeScriptFlow.ts
+++ b/src/Language/TypeScriptFlow.ts
@@ -154,8 +154,22 @@ export abstract class TypeScriptFlowBaseRenderer extends JavaScriptRenderer {
         return "function cast<T>(obj: any, typ: any): T";
     }
 
-    protected get typeAnnotations(): { any: string; anyArray: string; string: string; boolean: string } {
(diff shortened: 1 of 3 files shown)


src/Language/TypeScriptFlow.ts
```diff
@@ -154,8 +157,22 @@ export abstract class TypeScriptFlowBaseRenderer extends JavaScriptRenderer {
...
+        anyArray: string;
+        anyMap: string;
+        string: string;
+        stringArray: string;
+        boolean: string;
+    } {
+        return {
+            any: ": any",
+            anyArray: ": any[]",
+            anyMap: ": {[k: string]: any}",
```
dvdsgl: I usually see this written as `{ [key: string]: any }`, with spaces inside the `{ }`

schani: Yes.

dvdsgl:
Looks good to me.

PR #782, opened 2018-04-11
Separate stuff so as not to have a contagious urijs dep

schani:


dvdsgl:
Not only does this fix the build, but our bundle sizes have reduced by 40% since `v10`.

schani:
Which changes did that?

dvdsgl:
It would take me a while to figure that out specifically...

schani:
Is it easy to figure out which dependencies we're not including anymore?

dvdsgl:
It looks like quicktype actually grew, but there's a 25% size reduction due to a deduplicated TypeScript import(?).

# v10

image

image

# v12

image

image

schani:
Oh shit, we had the whole of typescript twice??

PR #803, opened 2018-04-17
[Java] Fix bug that always uses 'instantiateMapper' in Converter.java even when there are multiple classes being converted.

Enaero:
## Issue
Converter.java would use 'instantiateMapper' method even when there were multiple classes being converted with Converter.java. This caused a compilation error because the methods available are 'instantiate{Name}Mapper' where {Name} is a name of a generated class, and instantiateMapper is not found.

## Testing
jsons contains multiple json files
`script/quicktype -o javas/* -l java jsons`
`javac ...`
> No compilation error

jsons contains a single json file
`script/quicktype -o javas/* -l java jsons`
`javac ...`
> No compliation error

### DIFF
--- a/src/Language/Java.ts
+++ b/src/Language/Java.ts
@@ -632,12 +632,12 @@ export class JavaRenderer extends ConvenienceRenderer {
                 );
                 this.ensureBlankLine();
                 this.emitBlock(["private static ObjectReader ", this.readerGetterName(topLevelName), "()"], () => {
-                    this.emitLine("if (", readerName, " == null) ", this.methodName("instantiate", "Mapper", topLevelName), ";");
+                    this.emitLine("if (", readerName, " == null) ", this.methodName("instantiate", "Mapper", topLevelName), "();");
                     this.emitLine("return ", readerName, ";");
                 });
                 this.ensureBlankLine();
                 this.emitBlock(["private static ObjectWriter ", this.writerGetterName(topLevelName), "()"], () => {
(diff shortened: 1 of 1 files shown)


schani:
@Enaero the test suite fails.  Did you test on your machine with `FIXTURE=java npm test`?  If a test fails it will leave the directory where it tried running and you can see what went wrong.  For example:

```
Fixture threw an exception
  | {
  | "error": {
  | "command": "mvn package",
  | "code": 1
  | },
  | "sample": {
  | "path": "test/inputs/json/priority/list.json",
  | "saveOutput": true
  | },
  | "cwd": "/app/test/runs/java-720a87"
  | }
```

It says it tried to do `mvn package`, which failed, in `/app/test/runs/java-720a87`

schani:
Yes, just run `FIXTURE=java npm test`.

Enaero:
I got docker working and I found the error. I forgot the parentheses at the end while copying over the change between workspaces (sorry). It should work now, all 193 java tests passed locally.

Should I add a test for this use case?

schani:
We don't have the infrastructure to do multiple top-level tests, but if you can figure out how to do it, please go ahead - it is pretty important overall, not just for your PR.  I'm happy to merge this without the test case.

Enaero:
Ah OK. Given that it's not a small task to support this type of test I would like to consider it out of scope of this change. Could you merge it without the test case?

schani:
Thank you!

PR #807, opened 2018-04-18
Use DateTimeOffset unqualified in C#

### LINKED ISSUE #801: DateTimeOffset is written fully qualified
We write `System.DateTimeOffset` rather than `DateTimeOffset`. Perhaps there's a good reason to do this, but `System.` is often left implicit.

dvdsgl:
Fixes #801


dvdsgl:
This also fixes an issue where we emitted:

```csharp
 if (System.DateTimeOffset .TryParse(str, out dt))
...
```

PR #826, opened 2018-04-26
Update dependencies

dvdsgl:
This is a regular update of our dependencies.


dvdsgl:
We have a warning about a [medium-severity](https://nvd.nist.gov/vuln/detail/CVE-2018-3728) vulnerability in an indirect dependency of `pkg`. This PR does not fix it, because `pkg` and its dependencies have not updated yet. I don't believe this could affect quicktype in production.

PR #828, opened 2018-04-26
Always emit a default initializer for classes

### LINKED ISSUE #827: Swift compiler error for immutable classes without initializers
https://app.quicktype.io/?share=tBq6iqohUMJ3pF4e8EqT

dvdsgl:
It's a compiler error to define an immutable class without an initializer (or initail values)

Fixes #827


dvdsgl:
This was not discovered by our tests because `--initializers` is always on.

We may want to expand our test suite to make sure that output code is valid, even if it cannot read JSON.

PR #832, opened 2018-04-27
Don't make an enum if all cases are number strings

### LINKED ISSUE #453: Don't infer stringified numbers as enums

dvdsgl:
Fixes #453


schani:
Looks like it does its job!

PR #833, opened 2018-04-29
Reorganize and update to TS 2.8

schani:


dvdsgl:
Looks good to me, although I think the convention is to name source directories with `kebab-case`

PR #838, opened 2018-05-02
Kotlin Support

dvdsgl:
- [x] Get fixture running in Docker
- [x] Refactor enums
- [x] Support empty classes when `--framework none`
- [x] Support unions
- [x] Support top-level non-objects


dvdsgl:
Must fix this upstream: https://github.com/cbeust/klaxon/issues/145

schani:
Does this bug affect all unions?

dvdsgl:
No, I think it is a general bug affecting `List<List<T>>` where `T` requires custom conversion (enums, unions).

schani:
I see.  Is there maybe some other list/vector/array type we can use that makes it work?

dvdsgl:
I tried switching to `Array` and got the same error.

PR #843, opened 2018-05-04
Scala support

JannikArndt:
I added scala as an output format.

### DIFF
--- a/src/language/All.ts
+++ b/src/language/All.ts
@@ -9,6 +9,7 @@ import ObjectiveCTargetLanguage from "./Objective-C";
 import { JavaTargetLanguage } from "./Java";
 import { JavaScriptTargetLanguage } from "./JavaScript";
 import { TypeScriptTargetLanguage, FlowTargetLanguage } from "./TypeScriptFlow";
+import { ScalaTargetLanguage } from "./Scala";
 import SwiftTargetLanguage from "./Swift";
 import ElmTargetLanguage from "./Elm";
 import JSONSchemaTargetLanguage from "./JSONSchema";
@@ -25,6 +26,7 @@ export const all: TargetLanguage[] = [
     new TypeScriptTargetLanguage(),
     new JavaScriptTargetLanguage(),
     new FlowTargetLanguage(),
+    new ScalaTargetLanguage(),
     new SwiftTargetLanguage(),
     new ElmTargetLanguage(),
     new JSONSchemaTargetLanguage(),
--- a/src/language/Scala.ts
+++ b/src/language/Scala.ts
@@ -1,9 +1,7 @@
-import { Map } from "immutable";
-
-import { TypeKind, Type, ArrayType, MapType, EnumType, UnionType, ClassType, ClassProperty } from "../Type";
-import { matchType, nullableFromUnion, removeNullFromUnion, directlyReachableSingleNamedType } from "../TypeUtils";
+import { Type, EnumType, UnionType, ClassType } from "../Type";
+import { matchType, nullableFromUnion, directlyReachableSingleNamedType } from "../TypeUtils";
 import { TypeGraph } from "../TypeGraph";
-import { Sourcelike, maybeAnnotated } from "../Source";
+import { Sourcelike } from "../Source";
 import {
     utf16LegalizeCharacters,
     escapeNonPrintableMapper,
@@ -12,19 +10,17 @@ import {
     isAscii,
     isLetter,
     isDigit,
-    capitalize,
     splitIntoWords,
     combineWords,
     allUpperWordStyle,
     firstUpperWordStyle,
     allLowerWordStyle
 } from "../Strings";
-import { Name, Namer, funPrefixNamer, DependencyName } from "../Naming";
+import { Name, Namer, funPrefixNamer } from "../Naming";
 import { ConvenienceRenderer, ForbiddenWordsInfo } from "../ConvenienceRenderer";
 import { TargetLanguage } from "../TargetLanguage";
 import { BooleanOption, StringOption, Option } from "../RendererOptions";
-import { anyTypeIssueAnnotation, nullTypeIssueAnnotation } from "../Annotation";
-import { defined, assert, assertNever } from "../Support";
+import { defined, assert } from "../Support";
 
 export class ScalaTargetLanguage extends TargetLanguage {
     private readonly _justTypesOption = new BooleanOption("just-types", "Plain types only", false);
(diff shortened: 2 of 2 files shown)


schani:
@JannikArndt That's amazing!  Thank you.

Do you have plans to support union types eventually?

schani:
There are a few compile errors:

```
src/language/Scala.ts(132,26): error TS6138: Property '_packageName' is declared but its value is never read.
src/language/Scala.ts(133,26): error TS6138: Property '_justTypes' is declared but its value is never read.
 ```

Also, would you mind adding Scala to the test suite?  See `languages.ts` in `test`, and the very bottom of `fixtures.ts`.  Basically you'll have to write a little driver program (in `test/fixtures/scala`) that deserializes JSON and then serializes it back again, and you'll have to tell the test suite how to build and run it.  The `Dockerfile` will also have to install Scala, I guess.  Or maybe Maven does that automatically?

**Comment by JannikArndt** [comment]
**Union Types**
There is no obvious and universal way to represent union types in Scala (that's the beauty of the language 😉). A colleague of mine suggested nested `Either` (okay for two types, ugly for more) or shapeless' [`hlist`](https://github.com/milessabin/shapeless/wiki/Feature-overview:-shapeless-2.0.0#heterogenous-lists), which would add another dependency.
So to answer your question, if I plan to support union types eventually: *If* I find an adequate solution, I will definitely add it, but I'd rather _not_ support them then add a somewhat broken support. And I am still tinkering with possible solutions.

**Tests**
Yeah, I did see that one coming when I looked at #838 😉 It's on my list.

**Compile warnings**
Fixed that 👍

**Comment by schani** [comment]
I thought cases classes is how you represent unions in Scala?  Isn't that the equivalent to discriminated unions in ML-like languages?

We'll wait for the test suite before merging.

**Comment by schani** [comment]
@JannikArndt What's the status on this?  Can we help?

**Comment by JannikArndt** [comment]
Hey there, sorry, I didn't have the time to write the tests, yet :(

**Comment by hderms** [comment]
@JannikArndt is there anything wrong with just having a sealed trait of one argument case classes to represent union types? I suppose it would play weird with singleton types like None or something so I suppose that partially answers the question. 

Dotty won't become Scala 3 until 2020 as far as I've heard, so I feel like there are precious limited options in the short term.

**Comment by hderms** [comment]
@JannikArndt any opposition to handling this like https://scalapb.github.io/generated-code.html

specifically the section on oneOf?

**Comment by JannikArndt** [comment]
@hderms Sounds good, I personally like the ScalaPB style. I still didn't find the time to implement anything further, though :(

**Comment by schani** [comment]
@JannikArndt @hderms I'd really like to get this closed.  I'm happy to take over the work if you can give a short summary of what still needs to be done, from your point of view.  If you could write a test driver (see `test/fixtures/*` for other language's drivers) I'd be much obliged, but I can figure it out myself if you don't have the time at all.

**Comment by hderms** [comment]
@schani I'm not sure what needs to be done still but I felt like imitating the way ScalaPB does union types would be a reasonable way to do oneOf

**Comment by schani** [comment]
Doesn't look like this is moving forward, so I'm closing it for now.  Please revive when/if there is new work.

**Comment by hderms** [comment]
@schani I'm working on this now. Can you point me in the direction of a json schema or whatever that we would be expected to produce unions from? I tried `test/inputs/schema/direct-union.1.json` and `test/inputs/schema/union.json` and neither seem to produce unions in either Kotlin or Rust. Maybe those languages don't implement unions properly or am i incorrect in thinking these files should produce union'd types?

**Comment by hderms** [comment]
@schani maybe I'm misinterpreting it but I don't get how this output is a union?

`npm start -- "--lang csharp test/inputs/schema/union.schema"`
```
> quicktype@15.0.0 start /Users/dermothaughey/projects/quicktype
> script/watch "--lang csharp test/inputs/schema/union.schema"

> Watching src/
// <auto-generated />
//
// To parse this JSON data, add NuGet 'Newtonsoft.Json' then do:
//
//    using QuickType;
//
//    var union = Union.FromJson(jsonString);
```
```csharp
namespace QuickType
{
    using System;
    using System.Collections.Generic;

    using System.Globalization;
    using Newtonsoft.Json;
    using Newtonsoft.Json.Converters;

    public partial class Union
    {
        [JsonProperty("$schema")]
        public Uri Schema { get; set; }

        [JsonProperty("type")]
        public string Type { get; set; }

        [JsonProperty("items")]
        public Items Items { get; set; }
    }

    public partial class Items
    {
        [JsonProperty("oneOf")]
        public OneOf[] OneOf { get; set; }
    }

    public partial class OneOf
    {
        [JsonProperty("type")]
        public string Type { get; set; }

        [JsonProperty("properties")]
        public Properties Properties { get; set; }

        [JsonProperty("required")]
        public string[] OneOfRequired { get; set; }
    }

    public partial class Properties
    {
        [JsonProperty("one", NullValueHandling = NullValueHandling.Ignore)]
        public One One { get; set; }

        [JsonProperty("two")]
        public One Two { get; set; }

        [JsonProperty("three", NullValueHandling = NullValueHandling.Ignore)]
        public One Three { get; set; }
    }

    public partial class One
    {
        [JsonProperty("type")]
        public string Type { get; set; }
    }

    public partial class Union
    {
        public static Union FromJson(string json) => JsonConvert.DeserializeObject<Union>(json, QuickType.Converter.Settings);
    }

    public static class Serialize
    {
        public static string ToJson(this Union self) => JsonConvert.SerializeObject(self, QuickType.Converter.Settings);
    }

    internal static class Converter
    {
        public static readonly JsonSerializerSettings Settings = new JsonSerializerSettings
        {
            MetadataPropertyHandling = MetadataPropertyHandling.Ignore,
            DateParseHandling = DateParseHandling.None,
            Converters =
            {
                new IsoDateTimeConverter { DateTimeStyles = DateTimeStyles.AssumeUniversal }
            },
        };
    }
}
```

or this:
```go
package main

import "encoding/json"

func UnmarshalUnion(data []byte) (Union, error) {
	var r Union
	err := json.Unmarshal(data, &r)
	return r, err
}

func (r *Union) Marshal() ([]byte, error) {
	return json.Marshal(r)
}

type Union struct {
	Schema string `json:"$schema"`
	Type   string `json:"type"`
	Items  Items  `json:"items"`
}

type Items struct {
	OneOf []OneOf `json:"oneOf"`
}

type OneOf struct {
	Type       string     `json:"type"`
	Properties Properties `json:"properties"`
	Required   []string   `json:"required"`
}

type Properties struct {
	One   *One `json:"one,omitempty"`
	Two   One  `json:"two"`
	Three *One `json:"three,omitempty"`
}

type One struct {
	Type string `json:"type"`
}
```

is `union.schema` expected to output an actual union type?

schani:
Sorry, you'll have to pass `--src-lang schema`

hderms:
@schani my bad. That's what I get for trying to work on a project I barely understand :)

PR #850, opened 2018-05-08
[NO MERGE] Split quicktype into CLI and Core packages

schani:


schani:
@dvdsgl Please check whether this makes any sense.

I still have to remove some dependencies that aren't needed anymore.  Also I want to make one `index.ts` in Core that exports everything, so that the stupid `/dist/` isn't needed anymore when importing.

schani:
Superseded by #868

PR #871, opened 2018-05-25
Improve C++ Target Language Extendability

daichij:
Changed code generation methods in CPlusPlus.ts from `private` to `protected` to enable to make them overridable.

Added a function in CPlusPlus.ts named emitClassMembers to emit the contents of a class.


schani:
Looks great!  If CI likes it we'll merge it.

schani:
Thank you!

PR #873, opened 2018-05-30
Add onFirst and onLast arguments to all forEach methods

schani:


dvdsgl:
How about a new parameter that's "first" "last" "middle" or "only"? That seems easier to work with.

PR #874, opened 2018-05-30
Moved cppNameStyle from CPlusPlus.ts to support/Strings.ts and rename…

daichij:
…d it to makeNameStyle.

Added 2 new parameters to makeNameStyle:
 - legalizeName: A language-specific function that legalizes a name to render in a given language
 - prefix: An optional parameter that specifies a prefix to prepend to names

Moved NamingStyle enumerator from CPlusPlus.ts to support/Strings.ts to support the above change

CPlusPlus.ts:
 - Removed multiple private naming function variables and implemented them in their corresponding `makeXxxNamer` methods so they may be modified by inheritors
 - Changed the constructor to have C++ properties as default parameters instead of being hardcoded so that inheritors can make their language targetable by quicktype
 - updated code to use makeNamStyle instead of cppNameStyle
Removed private parameter naming functions


package.json
```diff
@@ -12,8 +12,7 @@
     "test": "jest && script/test",
     "start": "script/watch",
     "pkg": "script/make-pkgs.sh",
-    "tslint":
-      "tslint --project src/cli --exclude 'src/__tests__/**' --exclude 'src/quicktype-core/get-stream/**'",
+    "tslint": "tslint --project src/cli --exclude 'src/__tests__/**' --exclude 'src/quicktype-core/get-stream/**'",
```
schani: These all seem to be the result of reformatting.

src/quicktype-core/support/Strings.ts
```diff
@@ -505,3 +507,60 @@ export function addPrefixIfNecessary(prefix: string, name: string): string {
...
+            legalizeName,
+            firstWordStyle,
+            restWordStyle,
+            firstWordAcronymStyle,
+            restAcronymStyle,
+            separator,
+            isLetterOrUnderscore
+        );
+
+        if (prefix) {
```
schani: Please check `prefix !== undefined` instead.

package.json
schani: These changes shouldn't be in here.

PR #875, opened 2018-05-31
Enabled the option to select between upper cased and capitalized acronyms in camel and pascal case for C++

daichij:


src/quicktype-core/support/Strings.ts
```diff
@@ -32,7 +32,7 @@ function precomputedCodePointPredicate(p: CodePointPredicate): CodePointPredicat
     for (let cp = 0; cp < 128; cp++) {
         asciiResults.push(p(cp));
     }
-    return function(cp: number) {
+    return function (cp: number) {
```
schani: Whitespace damage.

src/quicktype-core/support/Strings.ts
```diff
@@ -389,7 +389,7 @@ export function splitIntoWords(s: string): WordInName[] {
         return i - intervalStart;
     }
 
-    for (;;) {
+    for (; ;) {
```
schani: More whitespace damage.

src/quicktype-core/support/Strings.ts
```diff
@@ -519,18 +519,24 @@ export function makeNameStyle(
     let firstWordAcronymStyle: WordStyle;
     let restAcronymStyle: WordStyle;
 
-    if (namingStyle === "pascal" || namingStyle === "camel") {
+    if (namingStyle === "pascal" || namingStyle === "camel" || namingStyle === "pascal-upper-acronyms" || namingStyle === "camel-upper-acronyms") {
         separator = "";
-        restWordStyle = firstUpperWordStyle;
-        restAcronymStyle = allUpperWordStyle;
+        if (namingStyle.includes("upper-acronyms")) {
```
schani: I think there's some compatibility issue for why we can't use `.includes`.  It's not in IE or something.  I also dislike depending on the exact value of the string - it makes search-and-replace harder, for example.  Please just do a comparison with both values.

PR #886, opened 2018-06-08
WIP: Add superclass options to the NewtonsoftCSharpTargetLanguage

bitsol:


src/quicktype-core/language/CSharp.ts
```diff
@@ -617,8 +628,16 @@ export class NewtonsoftCSharpRenderer extends CSharpRenderer {
             this.emitUsing([denseRequiredEnumName, " = Newtonsoft.Json.Required"]);
             this.emitUsing([denseNullValueHandlingEnumName, " = Newtonsoft.Json.NullValueHandling"]);
         }
+        if (this._options.superclass === "EntityData") {
+            this.emitUsing("Microsoft.Azure.Mobile.Server");
+        }
+    }
+    protected superclassForType(_t: Type): Sourcelike | undefined {
+        if (this._options.superclass) {
```
schani: You can just `return this._options.superclass` here.  If it's undefined it'll return `undefined` ;-)

PR #887, opened 2018-06-08
Fix watch script directory

bitsol:
`npm run start` doesn't work for me unless I switch the directories.
Could either @schani or @dvdsgl let me know if, with https://github.com/quicktype/quicktype/commit/23080ec8debf82cad1608deed3363cf97e05f6ff, this is the correct behavior?

### DIFF
--- a/script/watch
+++ b/script/watch
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-watch "script/quicktype $@" src/quicktype-core/language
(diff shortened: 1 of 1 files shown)


schani:
Hmm, my feeling is that it should watch the top `src`.  I'm embarrassed to say I never use this.  I'd say if you can verify that it works with `src`, then that's what it should be.

schani:
Thanks!

PR #941, opened 2018-07-04
Fix the CLI on Linux.  Fixes #940

### LINKED ISSUE #940: Quicktype hangs after upgrading version
We use quicktype in docker containers and it seems that since we updated to a newer version, the quicktype cli just hangs. It seems the issue started on **quicktype@15.0.30** because **quicktype@15.0.29** works fine. 

It's very simple to reproduce, so i'm wondering if i'm missing dependencies outside of what npm resolves by itself.

you can basically run these set of commands to get a docker up and running and installing npm on it and afterwards trying to run it. it hangs.
```sh
# run docker image of ubuntu
docker run -it ubuntu:latest bash

# install everything you need

apt-get update
apt-get install npm
apt-get install git

# you can select latest or anything between latest and 15.0.30
npm install -g  quicktype@latest

quicktype --version

```

Am i missing something? are there additional dependencies needed after version 15.0.29?

schani:
Problem:

https://stackoverflow.com/questions/4303128/how-to-use-multiple-arguments-for-awk-with-a-shebang-i-e

Solution:

http://sambal.org/2014/02/passing-options-node-shebang-line/

### DIFF
--- a/script/build.ts
+++ b/script/build.ts
@@ -7,46 +7,43 @@ import * as _ from "lodash";
 
 const OUTDIR = "dist";
 
-function mapFile(
-  source: string,
-  destination: string,
-  transform: (content: string) => string
-) {
-  const content = fs.readFileSync(source, "utf8");
-  fs.writeFileSync(destination, transform(content));
+function mapFile(source: string, destination: string, transform: (content: string) => string) {
+    const content = fs.readFileSync(source, "utf8");
+    fs.writeFileSync(destination, transform(content));
 }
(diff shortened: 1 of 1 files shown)


SSW-SCIENTIFIC:
This commit breaks quicktype-cli on windows because this makes /bin/sh required.

schani:
I guess we can't have pretty things :-(

Ok, I'll change it back to what it was before until I figure out how to solve it portably.

PR #965, opened 2018-07-16
Swift enhancements

### LINKED ISSUE #899: Swift structs need explicit initializers when access level is public
The default struct initializers are internal.

### LINKED ISSUE #900: Option to emit `with` method for modifying immutable Swift types
It would be nice to support an optional `with` method:

```swift
struct A {
  let age: Double
  let name: String

  func with(age: Double? = nil, name: String? = nil) -> A {
    return A(
      age: age ?? self.age,
      name: name ?? self.name
  }
}
```

dvdsgl:
* Explicit initializers for public structs (fixes #899)
* `.with` mutator methods (fixes #900)

### DIFF
--- a/src/quicktype-core/language/Swift.ts
+++ b/src/quicktype-core/language/Swift.ts
@@ -519,7 +519,12 @@ export class SwiftRenderer extends ConvenienceRenderer {
 
             // this main initializer must be defined within the class
             // declaration since it assigns let constants
-            if (isClass) {
+            if (
+                isClass ||
+                // Public structs need explicit initializers
+                // https://github.com/quicktype/quicktype/issues/899
+                this._options.accessLevel === "public"
+            ) {
                 // Make an initializer that initalizes all fields
                 this.ensureBlankLine();
                 let properties: Sourcelike[] = [];
--- a/src/quicktype-core/language/Swift.ts
+++ b/src/quicktype-core/language/Swift.ts
@@ -31,7 +31,7 @@ const MAX_SAMELINE_PROPERTIES = 4;
 
 export const swiftOptions = {
     justTypes: new BooleanOption("just-types", "Plain types only", false),
-    convenienceInitializers: new BooleanOption("initializers", "Convenience initializers", true),
+    convenienceInitializers: new BooleanOption("initializers", "Convenient initializers & mutators", true),
     urlSession: new BooleanOption("url-session", "URLSession task extensions", false),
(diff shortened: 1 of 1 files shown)


src/quicktype-core/language/Swift.ts
```diff
@@ -952,6 +960,37 @@ ${this.accessLevel}class JSONAny: Codable {
...
 
+    private emitConvenienceMutator(c: ClassType, className: Name) {
+        this.emitLine(this.accessLevel, "func with(");
+        this.indent(() => {
+            this.forEachClassProperty(c, "none", (name, _, p, position) => {
+                this.emitLine(
+                    name,
+                    ": ",
+                    this.swiftPropertyType(p),
+                    "? = nil",
```
caffeineflo: @dvdsgl For an already optional argument this would add something like `:String?? = nil` is this intentional? If so, what would be the difference to `:String? = nil`? 
(I apologies if this seems out of the blue, discovered a sample today and wasn't sure what I'm seeing)

schani: I think that would be an bug.  Do you have a test case for that?  Could you open an issue?

dvdsgl: This is intentional. I'm open to other ways to specify nil for an optional property, but this is the most unadorned approach I could think of.

schani: Ah, I see - you can replace an optional value with `nil` by using `Optional.some(nil)`.  Ok, it's a feature, I guess :-)

dvdsgl: Or `Optional(nil)` — far from lovely but the simplest semantics I could think of…

It's still tricky so I'm open to improvements.

caffeineflo: I see. But isn't the output `with(test: String?? = nil)` technically invalid Swift and it should rather be `with(test: String? = nil)` which would be valid Swift syntax with the same result?

schani: It's valid Swift - it compiles in our test suite.

The functionality of `with` is to replace one or more property values with new values.  If you don't want to replace the value, you pass in `nil`.  But if the property is optional, `nil` is itself a valid value, so what if you want to replace the value with `nil`?  The ugly answer: If you don't want to replace it, you pass `nil`.  If you want to replace it with `nil`, you pass `Optional(nil)`.

caffeineflo: Thank you so much for the explanation! It makes total sense now!!

PR #982, opened 2018-07-28
Enums with values, qt-enum-values 

harsszegi:
Added enums with values (like used by C++), when someone needs to define the ordinal value of an enum member. E.g. enum Foo { Bar = 1, FooBar = 2 };


schani:
FYI: The build error is just because your PR produces new output files.  We always diff all outputs against master, and if there's a change, we report it as an error, because there is no "warning" category for tests.  I was too lazy so far to make it so that new files in the diff don't trigger an error.

src/quicktype-core/EnumValues.ts
```diff
@@ -0,0 +1,41 @@
...
+        super("enumValues");
+    }
+    makeInferred(_: AccessorNames) {
+        return undefined;
+    }
+}
+
+export const enumValuesTypeAttributeKind: TypeAttributeKind<AccessorNames> = new EnumValuesTypeAttributeKind();
+
+export function enumCaseNames(e: EnumType, language: string): Map<string, [string, boolean] | undefined> {
```
schani: This should be called `enumCaseValues`, I think.

src/quicktype-core/EnumValues.ts
```diff
@@ -0,0 +1,41 @@
...
+    const enumvalues = enumValuesTypeAttributeKind.tryGetInAttributes(e.getAttributes());
```
schani: `enumValues`

src/quicktype-core/language/CPlusPlus.ts
```diff
@@ -479,9 +481,18 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
...
-        this.forEachEnumCase(e, "none", name => {
+        const enumValues = enumCaseNames(e, this.targetLanguage.name );
+
+        this.forEachEnumCase(e, "none", (name, jsonName) => {
             if (caseNames.length > 0) caseNames.push(", ");
             caseNames.push(name);
+
+            if (enumValues !== undefined) {
+                const [ enumvalue, isFixed ] = getAccessorName(enumValues, jsonName);
+                if (enumvalue !== undefined && isFixed !== undefined) {
```
schani: `isFixed` is a boolean.  The comparison to `undefined` will always succeed.

The fixed thing only really applied to names - if a name is given for a specific target language then it's taken verbatim.

src/quicktype-core/language/CPlusPlus.ts
```diff
@@ -479,9 +481,18 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
...
         const caseNames: Sourcelike[] = [];
-        this.forEachEnumCase(e, "none", name => {
+        const enumValues = enumCaseNames(e, this.targetLanguage.name );
+
+        this.forEachEnumCase(e, "none", (name, jsonName) => {
             if (caseNames.length > 0) caseNames.push(", ");
             caseNames.push(name);
+
+            if (enumValues !== undefined) {
+                const [ enumvalue, isFixed ] = getAccessorName(enumValues, jsonName);
```
schani: `enumValue`

src/quicktype-core/language/CPlusPlus.ts
```diff
@@ -479,9 +481,18 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
...
+                if (enumvalue !== undefined && isFixed !== undefined) {
+                    caseNames.push(" = " + enumvalue);
```
schani: `caseNames.push(" = ", enumValue.toString());`

schani:
Superseded by #987

PR #992, opened 2018-07-31
Make Output Typescript Pass Strict Mode Compilation

Logikgate:
When using quicktype to generate a typescript file the file was missing a few explicit any designations.  This adds them so that the output file will pass the typescript compiler in strict mode.


schani:
This is fabulous!  Thank you so much!

Issue #975, opened 2018-07-27
CSharp support for public static implicit operator in struct


danielmeza:
If we add support for public static implicit operator for every member of the union we can asing any type of its values for example

```csharp
public partial struct IFontWeightUnion
    {
        public static implicit operator IFontWeightUnion(IFontWeightEnum fontWeight) => new IFontWeightUnion { Enum = fontWeight}; 
        public static implicit operator IFontWeightUnion(double fontWeight) => new IFontWeightUnion { Double= fontWeight}; 
        public static implicit operator IFontWeightUnion(long fontWeight) => new IFontWeightUnion { Integer= fontWeight}; 
        public double? Double;
        public IFontWeightEnum? Enum;
        public long? Integer;
        public bool IsNull => Double == null && Integer == null && Enum == null;
    }
```
then we can use like this
```csharp
  IFontWeightUnion font = IFontWeightEnum.Value
  IFontWeightUnion font2 = 1; 
  IFontWeightUnion font3 = 2.0D;
```

schani:
Seems like a great idea!  I wonder whether there are any corner cases where there would be overloading conflicts.

Would you be interested in contributing?  This should be fairly straightforward to implement.  Take a look at [`emitUnionDefinition`](https://github.com/quicktype/quicktype/blob/master/src/quicktype-core/language/CSharp.ts#L436).

danielmeza:
Yes I have made it but I can't figureout how to test it :D So i will make the pull request

schani:
Wow, that's amazing!

Yes, we currently don't have testing infrastructure to test features that go beyond (de)serializing.  We do take contributions for that, too, in case you're adventurous ;-)

The test suite will at least test whether your code compiles, so there's that.

schani:
Implemented in #995

PR #995, opened 2018-08-01
#975 Implementing implicit operator for c#

danielmeza:
#975 hi @schani  here is  👍


schani:
@danielmeza Excellent!

PR #1005, opened 2018-08-03
C++: Support for minimum, maximum, minLength, maxLength, pattern

harsszegi:
JSON schema supports constraints on properties (minimum value, maximum value, minimum length, laxmimum length, regexp pattern).

The change addresses this problem - with regards to C++.
Notes:
- This is obviously only works with getter-setter
- Currently regexp checking is NOT yet implemented (needs to decided whether using boost::regexp
is enough or an overkill, or regexp should be generally "optionalized")
- only works on number + string (I doubt that anything else is supported by the standard)
- NO tests are yet created, however constaint.json + constraint.schema is checked in, have zero
idea how to create negative tests (obviously I have tested them, an exception is thrown and printed to the stdout) in the testing framework.


schani:
I haven't looked at the C++ part much yet.  I assume this will also check the constraints on deserialization?

We still don't have support for negative test cases, unfortunately.  I'll see if I can find time this weekend. #1006

As for the attributes, there are a few issues:

1. The attributes have to be on the types themselves, not on properties' classes.  There are a bunch of reasons for this, including that the way you're collecting the attributes won't get all of them, for example a property type might be a `$ref`, but you're not following it.  Also the constraint might be for array items, which you're completely ignoring.  (It's ok for this PR not to check constraints on array items, but it's not ok for it not to collect them)

2. Don't store numbers as strings.

3. At least the min/max, min/maxLength constraints should support combining and intersecting.  The `combine` and `intersect` methods are used for `oneOf` and `allOf` schemas.  I'm fine if we only keep only interval, even if this isn't as constraining as it should be.  So, you could have a `oneOf` of intervals [1-2] and [3-4].  Technically that means (2-3) is not allowed, but I'm fine if we allow [1-4].  If `intersect` results in an empty interval, it's ok to fail with an error.  See `Messages.ts` on how to add a new one.

4. For pattern I'm fine with not doing anything smart implementing `combine` and `intersect`, but whatever the implementations do, they have to accept JSONs that conform to the constraints.  The easiest thing would be for both of them to just remove the constraint completely.  Unfortunately neither of those two are allowed to return `undefined` right now - I'll fix that.

schani:
Also, `combinations.json` fails with `with-getter-setter`:

```
In file included from main.cpp:6:0:
TopLevel.hpp:33:16: error: 'optional' in namespace 'boost' does not name a template type
         boost::optional<int> minValue;
                ^
TopLevel.hpp:34:16: error: 'optional' in namespace 'boost' does not name a template type
         boost::optional<int> maxValue;
                ^
TopLevel.hpp:35:16: error: 'optional' in namespace 'boost' does not name a template type
         boost::optional<int> minLength;
                ^
TopLevel.hpp:36:16: error: 'optional' in namespace 'boost' does not name a template type
         boost::optional<int> maxLength;
                ^
TopLevel.hpp:37:16: error: 'optional' in namespace 'boost' does not name a template type
         boost::optional<std::string> pattern;
                ^
TopLevel.hpp:41:28: error: expected ')' before '<' token
             boost::optional<int> MinValue,
                            ^
TopLevel.hpp: In member function 'void quicktype::ClassMemberConstraint::setMinValue(int)':
TopLevel.hpp:50:39: error: 'minValue' was not declared in this scope
         void setMinValue(int value) { minValue = value; }
                                       ^
TopLevel.hpp: In member function 'auto quicktype::ClassMemberConstraint::getMinValue() const':
TopLevel.hpp:51:43: error: 'minValue' was not declared in this scope
         auto getMinValue() const { return minValue; }
                                           ^
TopLevel.hpp: In member function 'void quicktype::ClassMemberConstraint::setMaxValue(int)':
TopLevel.hpp:53:39: error: 'maxValue' was not declared in this scope
         void setMaxValue(int value) { maxValue = value; }
                                       ^
TopLevel.hpp: In member function 'auto quicktype::ClassMemberConstraint::getMaxValue() const':
TopLevel.hpp:54:43: error: 'maxValue' was not declared in this scope
         auto getMaxValue() const { return maxValue; }
                                           ^
TopLevel.hpp: In member function 'void quicktype::ClassMemberConstraint::setMinLength(int)':
TopLevel.hpp:56:40: error: 'minLength' was not declared in this scope
         void setMinLength(int value) { minLength = value; }
                                        ^
TopLevel.hpp: In member function 'auto quicktype::ClassMemberConstraint::getMinLength() const':
TopLevel.hpp:57:44: error: 'minLength' was not declared in this scope
         auto getMinLength() const { return minLength; }
                                            ^
TopLevel.hpp: In member function 'void quicktype::ClassMemberConstraint::setMaxLength(int)':
TopLevel.hpp:59:40: error: 'maxLength' was not declared in this scope
         void setMaxLength(int value) { maxLength = value; }
                                        ^
TopLevel.hpp: In member function 'auto quicktype::ClassMemberConstraint::getMaxLength() const':
TopLevel.hpp:60:44: error: 'maxLength' was not declared in this scope
         auto getMaxLength() const { return maxLength; }
                                            ^
TopLevel.hpp: In member function 'void quicktype::ClassMemberConstraint::setPattern(const string&)':
TopLevel.hpp:62:54: error: 'pattern' was not declared in this scope
         void setPattern(const std::string & value) { pattern = value; }
                                                      ^
TopLevel.hpp: In member function 'auto quicktype::ClassMemberConstraint::getPattern() const':
TopLevel.hpp:63:42: error: 'pattern' was not declared in this scope
         auto getPattern() const { return pattern; }
                                          ^
TopLevel.hpp: In function 'void quicktype::checkConstraint(const string&, const quicktype::ClassMemberConstraint&, int64_t)':
TopLevel.hpp:92:32: error: 'none' is not a member of 'boost'
         if (c.getMinValue() != boost::none && value < *c.getMinValue()) {
                                ^
TopLevel.hpp:92:69: error: void value not ignored as it ought to be
         if (c.getMinValue() != boost::none && value < *c.getMinValue()) {
                                                                     ^
In file included from main.cpp:6:0:
TopLevel.hpp:93:130: error: void value not ignored as it ought to be
             throw ValueTooLowException ("Value too low for "+name+" (" + std::to_string(value)+ "<"+std::to_string(*c.getMinValue())+")");
                                                                                                                                  ^
TopLevel.hpp:96:32: error: 'none' is not a member of 'boost'
         if (c.getMaxValue() != boost::none && value > *c.getMaxValue()) {
                                ^
TopLevel.hpp:96:69: error: void value not ignored as it ought to be
         if (c.getMaxValue() != boost::none && value > *c.getMaxValue()) {
                                                                     ^
TopLevel.hpp:97:132: error: void value not ignored as it ought to be
             throw ValueTooHighException ("Value too high for "+name+" (" + std::to_string(value)+ ">"+std::to_string(*c.getMaxValue())+")");
                                                                                                                                    ^
TopLevel.hpp: In function 'void quicktype::checkConstraint(const string&, const quicktype::ClassMemberConstraint&, const string&)':
TopLevel.hpp:102:33: error: 'none' is not a member of 'boost'
         if (c.getMinLength() != boost::none && value.length() < *c.getMinLength()) {
                                 ^
TopLevel.hpp:102:80: error: void value not ignored as it ought to be
         if (c.getMinLength() != boost::none && value.length() < *c.getMinLength()) {
                                                                                ^
TopLevel.hpp:103:144: error: void value not ignored as it ought to be
             throw ValueTooShortException ("Value too short for "+name+" (" + std::to_string(value.length())+ "<"+std::to_string(*c.getMinLength())+")");
                                                                                                                                                ^
TopLevel.hpp:106:33: error: 'none' is not a member of 'boost'
         if (c.getMaxLength() != boost::none && value.length() > *c.getMaxLength()) {
                                 ^
TopLevel.hpp:106:80: error: void value not ignored as it ought to be
         if (c.getMaxLength() != boost::none && value.length() > *c.getMaxLength()) {
                                                                                ^
TopLevel.hpp:107:142: error: void value not ignored as it ought to be
             throw ValueTooLongException ("Value too long for "+name+" (" + std::to_string(value.length())+ ">"+std::to_string(*c.getMaxLength())+")");
                                                                                                                                              ^
 
Fixture threw an exception
{
  "error": {
    "command": "g++ -O0 -o quicktype -std=c++14 main.cpp",
    "code": 1
  },
  "sample": {
    "path": "test/inputs/json/priority/combinations.json",
    "additionalRendererOptions": {
      "code-format": "with-getter-setter"
    },
    "saveOutput": false
  },
  "cwd": "/app/test/runs/cplusplus-8ae702"
}
```

schani:
Given the way min/max and min/maxLength interact, each pair should definitely be in a single attribute.  That way, if we ever decide we want to support multiple ranges, we can.  And we can detect empty intervals.  Of course the intervals can be open, so `min` and `max` have to be `number | undefined`.

harsszegi:
Aok, thanks for the comments, will fix the singlesource stuff first, completely forgot to add the extra includes.

Aok will try refactor to work on types, though still have hard time what does
combine do :)

schani:
@harsszegi `combine` must produce a new type attribute for a union of types (`oneOf` or `anyOf`) given the attributes for the union elements.  Let's say you have:

```
{
  "oneOf": [
    { "type": "integer", "minimum": 5 },
    { "type": "integer", "minimum": 3, "maximum": 8 }
  ]
}
```

This is a union of two integer types, so `combine` will be called with those two types' constraints, and its result is the constraint for the combined type.  In this case, the correct result constraint is a minimum of 3 and no maximum.

Does that make sense?

harsszegi:
Ooooh noew I understand, thanks.
One more favour. Can't get my head around how to use schema[] and map rthe found values to Types.
E.g. in my json I have 2 properties (width, height), both have minimum(width=5,height=10) and maximum set(width=15, height=20). So if I simply
fetch schema["minimum"] I will have a Set with 5,15 I suppose.
How will it be mapped back to width and height?
Thanks,

schani:
@harsszegi I'm not sure I fully understand your question.  In the attribute producer you just return the attribute value with `{ forType: constraints }`.  quicktype takes care of keeping the attributes with the types they belong to.  In the renderer you get the attribute for a given type with a `TypeAttributeStoreView`.  That's the recommended way, at least, since the API is nice and type-safe.

schani:
No, there is a problem! qucktype will currently only keep one "copy" of each primitive type around, so all the constraints for numbers would get combined into one single constraint for all numbers :-(.

I don't know of a particular reason why keeping multiple copies around would cause issues, but it's never been tested, so it would almost certainly cause issues ;-)

I'll try to figure this out this weekend. Until then I'm afraid this is blocked. Or at least if you implement it like it's supposed to work it won't work properly ;-)

schani:
Proposal: I'll figure out and implement the min/max constraints for numbers. Then you can do the C++ part and the string constraints.

harsszegi:
Would you be so kind as to check it? It doesn't right now at all, objectValuesAttributeProducer doesn't find any attributes.
Thanks,

schani:
@harsszegi Take a look at #1009 .  This was much easier than I imagined - so far everything seems to work fine with multiple number types.  Could I ask you to work off that branch for now?  I still have to add a few things, such as test cases, JSON Schema output, maybe transformers with C#/Python output.  I won't rebase, it, so you're safe to merge or rebase on top of it, or restart from scratch.

You can use the type attribute kind I wrote unchanged for min/maxLength, btw.  And then add one more for `pattern`.  You'll have to make `inIdentity` return `true` there, too.

schani:
@harsszegi In case you didn't know: You don't have to close a PR and open a new one even if you want to throw everything away - you can just force-push on your branch and the PR will update.  That way the discussion is preserved better, so please do that if necessary.

For future PRs, please also split work up into smaller PRs.  In this case, for example, min/max would be one PR, another min/maxLength, then another one for pattern.  That makes reviewing easier and gets them merged quicker.  But you're fine here, no worries.

schani:
Ok, now it looks like it does what it's supposed to do.  Now the min/maxLength producer will have to return its attributes with `forString`, though, which I haven't implemented yet ;-).  Doing that now.

harsszegi:
Ok, pulled in your changed, one thing. What should the combine for pattern do? Throw an error? Or?

harsszegi:
Ok, appears to be working, the expected source is generated, also running the test program produces the expected exceptions for all constraints. kudos :)

Please note that I have also started to fix the stringGenerator stuff as well so it might break the build

schani:
For pattern like I said, right now I'm happy with a non-solution for union and intersection. You can just return undefined and write a FIXME comment.

Having said that, I think combine can be implemented just by parenthesising the individual regexes and joining them with |. That's the union of regexps, literally.

The technically correct solution if both unions and intersections are involved would be to build up the tree of unions and intersections, emit and traverse it at runtime.

schani:
@harsszegi The JSON input files for a Schema test file must be in `test/inputs/schema` and named `xxx.1.json`, etc.  This is in reference to `objectlimits.json`.

schani:
@harsszegi It seems like you didn't properly merge my branch into yours, so we get conflicts now.  I wouldn't bother resolving those constraints - you can just do a diff between quicktype's `master` and your branch and use that diff as a single commit.  If you need help with that, let me know.

harsszegi:
Naah don’t bother with it right now I am refactoring loads of things, running all tests, fixing the fond bugs

src/quicktype-core/Constraints.ts
```diff
@@ -0,0 +1,186 @@
...
+    combine(arr: string[]): string {
+        assert(arr.length > 0);
+
+        let patt:string="";
+        for (let i = 1; i < arr.length; i++) {
+            patt += "("+arr[i]+")";
+            if (i !== arr.length-1) {
+                patt += "|";
+            }
+        }
```
schani: You're not adding element `0`.  Also, shorter:

```
arr.map(p => `(${p})`).join("|")
```

src/quicktype-core/Constraints.ts
```diff
@@ -0,0 +1,186 @@
...
+        return undefined;
+    }
+
+    makeInferred(_: string): undefined {
+        return undefined;
+    }
+}
+
+export const patternTypeAttributeKind: TypeAttributeKind<string> = new PatternTypeAttributeKind(
+    "pattern"
```
schani: I think the `PatternTypeAttributeKind` should have a constructor that calls `super("pattern")`.  We only need one.

src/quicktype-core/Constraints.ts
```diff
@@ -0,0 +1,186 @@
...
+
+export function patternAttributeProducer(
+    schema: JSONSchema,
+    _ref: Ref,
+    types: Set<JSONSchemaType>
+): JSONSchemaAttributes | undefined {
+    if (!(typeof schema === "object")) return undefined;
+    if (!types.has("string")) return undefined;
+
+    let patt: string | undefined = undefined;
```
schani: Write this as

```
const patt = schema.pattern;
if (typeof part !== "string") return undefined;
return { forString: ... };
```

src/quicktype-core/Constraints.ts
```diff
@@ -0,0 +1,186 @@
...
+
+    if (typeof schema["pattern"] === "string") {
+        patt = schema["pattern"];
+    }
+
+    if (patt === undefined) return undefined;
+    return { forString: patternTypeAttributeKind.makeAttributes(patt) };
+}
+
+export function pattern(t: Type): string | undefined {
```
schani: Ok, let's call all these getter functions `xxxForType`, please.  So `patternForType`, `minMaxConstraintForType`, `minMaxLengthConstraintForType`.

src/quicktype-core/Messages.ts
```diff
@@ -11,6 +11,8 @@ export type ErrorProperties =
       }
     | { kind: "MiscReadError"; properties: { fileOrURL: string; message: string } }
     | { kind: "MiscUnicodeHighSurrogateWithoutLowSurrogate"; properties: {} }
+    | { kind: "MiscInvalidMinMaxConstraint"; properties: { min: number; max: number } }
+    | { kind: "MiscMustntIntersectStringAttribute"; properties: {} }
```
schani: Remove this unused message.

src/quicktype-core/Constraints.ts
schani: Wait, you're not even using this function?

src/quicktype-core/language/CPlusPlus.ts
```diff
@@ -616,29 +626,127 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
...
+            }
+        });
+
+        return res.size === 0 ? undefined : res;
+    }
+
+    protected emitStringGenerator(c: ClassType): void {
+        this.forEachClassProperty(c, "none", (name, _jsonName, property) => {
+            const [getterName, , ] = defined(this._gettersAndSettersForPropertyName.get(name));
+            if (property.type.kind === "array") {
```
schani: What if the property is an array of maps of enums?

schani: I'm fine if the code doesn't handle that case fully, but it shouldn't emit invalid code.

src/quicktype-core/language/CPlusPlus.ts
```diff
@@ -940,9 +1048,111 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
     }
 
     protected emitHelperFunctions(): void {
+
+        if (this._options.codeFormat) {
+            this.emitBlock(["class ClassMemberConstraint"], true, () => {
```
schani: Actually, these class and member names have to be `Name`s, conforming to the naming convention selected in the options.  Goes for `minValue`, etc, too.

src/quicktype-core/language/CPlusPlus.ts
```diff
@@ -940,9 +1048,111 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
...
+                this.ensureBlankLine();
+
+                this.emitBlock(["if (c.getMaxValue() != boost::none && value > *c.getMaxValue())"], false, () => {
+                    this.emitLine("throw ValueTooHighException (\"Value too high for \"+name+\" (\" + std::to_string(value)+ \">\"+std::to_string(*c.getMaxValue())+\")\");");
+                });
+                this.ensureBlankLine();
+            });
+            this.ensureBlankLine();
+
+            this.emitBlock(["void checkConstraint(const std::string & name, const ClassMemberConstraint & c, const std::string & value)"], false, () => {
```
schani: Also, our (non-optioned) C++ convention for functions is `snake_case`.

harsszegi:
I need a help. I need to differentiate between union where the "toplevel" type of the union has a name, e.g. nameForNamedType returns a Name. Is there a way to check when I can I safely call nameForNamedType (currently for unnamed unions "somewhere deep" it throws an error:
Error: Internal error: Defined value expected, but got undefined
Thanks,

src/quicktype-core/language/CPlusPlus.ts
harsszegi: Hmm help me with this. What should I call to generate the correct names?

schani:
@harsszegi Your "lame" way works, and I usually do that.  If you want to be super correct you can call `this.unionNeedsName(t)` which is [defined in `ConvenienceRenderer`](https://github.com/quicktype/quicktype/blob/master/src/quicktype-core/ConvenienceRenderer.ts#L171).

src/quicktype-core/language/CPlusPlus.ts
schani: Actually, for the members of this class you don't have use a `Name`, since there can't be any conflict with other names, but you will have to use the name styling function selected by the current options.

Actually, come to think of it, you have two options for `ClassMemberConstraint`, and any other top-level name you introduce (the problem here, apart from name style, is that they can conflict with generated types):

1. Make a `Name` for them.  We've never needed this yet, so there's no "nice" way to do it.  You'll have to overload `setUpNaming()`, call it on `super`, and then add your own name.  Make sure you return the value that the call to `super` returned.

2. Easier: You can add the properly styled named dynamically to the forbidden names.  You just need to add it to the array returned by `forbiddenNamesForGlobalNamespace()`.  I think this is what you should do.

harsszegi:
Ok I retreat on the stringgenerator stuff. This is waaay to much work, and the end result will not be a single bit better if I convert the structure to Json and I dump the Json string (for debugging). And obviously I can do that substructure level as well, and not to mention that the output will be a lot better anyhow.

harsszegi:
Ok, this will take a LOT more than I have though. If I want to be thorough I have to make the ClassMemberConstraints member variables and getters/setters also made by a Namer.

schani:
@harsszegi Please don't do this - you're not using names they're supposed to be used.  `sourcelikeToString` is a hack that I want to eventually get rid of.  Please do what I suggested above and just add the styled global names to the forbidden names.  The members of your helper classes shouldn't need names because they can't collide with anything.

harsszegi:
Ok, but then we are back at the issue that the generated code format will not be unified. Or? Is it not a problem then?

schani:
@harsszegi I'm saying apply the name styling function (what `makeNameStyle` returns) to your name strings so that they do conform to the selected convention.

You'll need to save the two name stylers that are made here and passed to `funPrefixNamer`, in instance variables:

```
        this._memberNamingFunction = funPrefixNamer("members", makeNameStyle(_options.memberNamingStyle, legalizeName));

    protected makeNamedTypeNamer(): Namer {
        return funPrefixNamer("types", makeNameStyle(this.typeNamingStyle, legalizeName));
    }
```

schani:
@harsszegi I've implemented failure test cases, as promised.  Here's what you'll have to do to add one:

1. In `test/languages.ts` add a `LanguageFeature` `"minmax"` at the top (if you're adding the test case for min-max).

2. Add `"minmax"` to the `features` of `CPlusPlusLanguage`.

3. Add a test case like `test/inputs/schema/minmax.1.fail.minmax.json`.  The `minmax` after `fail` indicates to only run this test case for languages that have the `minmax` feature.  It's expected to fail, i.e. the generated executable has to return with a non-null exit status.

harsszegi:
Ok, fixed the naming, added finally regexp support for the constraint, however I have not yet created a test case for it yet :( [time pressure in office]

harsszegi:
Can you tell me what fails? Thanks

schani:
No failures, and the generated code looks good.  I'll look at the code in a bit.

schani:
Actually, the code doesn't generate any constraints.  Here's the output for `objectlimits.json`:

```
//  To parse this JSON data, first install
//
//      Boost     http://www.boost.org
//      json.hpp  https://github.com/nlohmann/json
//
//  Then include this file, and then do
//
//     Objectlimits data = nlohmann::json::parse(jsonString);

#pragma once

#include "json.hpp"

namespace nlohmann {
    template <typename T>
    struct adl_serializer<std::shared_ptr<T>> {
        static void to_json(json& j, const std::shared_ptr<T>& opt) {
            if (!opt) j = nullptr; else j = *opt;
        }

        static std::shared_ptr<T> from_json(const json& j) {
            if (j.is_null()) return std::unique_ptr<T>(); else return std::unique_ptr<T>(new T(j.get<T>()));
        }
    };
}

namespace quicktype {
    using nlohmann::json;

    inline json get_untyped(const json &j, const char *property) {
        if (j.find(property) != j.end()) {
            return j.at(property).get<json>();
        }
        return json();
    }

    template <typename T>
    inline std::shared_ptr<T> get_optional(const json &j, const char *property) {
        if (j.find(property) != j.end()) {
            return j.at(property).get<std::shared_ptr<T>>();
        }
        return std::shared_ptr<T>();
    }

    struct Dimension {
        int64_t height;
        int64_t width;
    };

    struct Objectlimits {
        std::shared_ptr<bool> checked;
        std::shared_ptr<Dimension> dimensions;
        std::shared_ptr<int64_t> id;
        std::shared_ptr<std::string> name;
        std::shared_ptr<double> price;
        std::shared_ptr<std::vector<std::string>> tags;
    };
}

namespace nlohmann {
    inline void from_json(const json& _j, quicktype::Dimension& _x) {
        _x.height = _j.at("height").get<int64_t>();
        _x.width = _j.at("width").get<int64_t>();
    }

    inline void to_json(json& _j, const quicktype::Dimension& _x) {
        _j = json::object();
        _j["height"] = _x.height;
        _j["width"] = _x.width;
    }

    inline void from_json(const json& _j, quicktype::Objectlimits& _x) {
        _x.checked = quicktype::get_optional<bool>(_j, "checked");
        _x.dimensions = quicktype::get_optional<quicktype::Dimension>(_j, "dimensions");
        _x.id = quicktype::get_optional<int64_t>(_j, "id");
        _x.name = quicktype::get_optional<std::string>(_j, "name");
        _x.price = quicktype::get_optional<double>(_j, "price");
        _x.tags = quicktype::get_optional<std::vector<std::string>>(_j, "tags");
    }

    inline void to_json(json& _j, const quicktype::Objectlimits& _x) {
        _j = json::object();
        _j["checked"] = _x.checked;
        _j["dimensions"] = _x.dimensions;
        _j["id"] = _x.id;
        _j["name"] = _x.name;
        _j["price"] = _x.price;
        _j["tags"] = _x.tags;
    }
}
```

If you run with `--debug=print-graph` you can see that the constraint attributes are definitely on the types:

```
0: class objectlimits | children 1,4,8,10,12,14 | distance 1 objectlimits (objectlimits_class)
1: union The Checked Schema | children 2,3 | distance 1 The Checked Schema (checked_union,objectlimits_checked,objectlimits_checked_union)
2: bool The Checked Schema | distance 1 The Checked Schema (checked_bool,objectlimits_checked,objectlimits_checked_bool)
3: null checked | distance 10 checked,dimensions,id,name,price,tags (checked_null,dimensions_null,id_null,name_null,price_null,tags_null,objectlimits_checked,objectlimits_checked_null,objectlimits_dimensions,objectlimits_dimensions_null,objectlimits_id,objectlimits_id_null,objectlimits_name,objectlimits_name_null,objectlimits_price,objectlimits_price_null,objectlimits_tags,objectlimits_tags_null)
4: union dimension | children 5,3 | distance 3 dimension (dimensions_union,objectlimits_dimensions,objectlimits_dimensions_union)
5: class dimension | children 6,7 | distance 3 dimension (dimensions_class,objectlimits_dimensions,objectlimits_dimensions_class)
6: integer The Height Schema | 10-20 | distance 1 The Height Schema (height_integer,dimensions_height,dimensions_height_integer)
7: integer The Width Schema | 5-15 | distance 1 The Width Schema (width_integer,dimensions_width,dimensions_width_integer)
8: union The Id Schema | children 9,3 | distance 1 The Id Schema (id_union,objectlimits_id,objectlimits_id_union)
9: integer The Id Schema | distance 1 The Id Schema (id_integer,objectlimits_id,objectlimits_id_integer)
10: union The Name Schema | children 3,11 | distance 1 The Name Schema (name_union,objectlimits_name,objectlimits_name_union) | 3-10 | unrestricted
11: string The Name Schema | distance 1 The Name Schema (name_string,objectlimits_name,objectlimits_name_string) | 3-10 | unrestricted
12: union The Price Schema | children 13,3 | distance 1 The Price Schema (price_union,objectlimits_price,objectlimits_price_union)
13: double The Price Schema | distance 1 The Price Schema (price_double,objectlimits_price,objectlimits_price_double)
14: union tag | children 15,3 | distance 3 tag (tags_union,objectlimits_tags,objectlimits_tags_union)
15: array tag | children 16 | distance 3 tag (tags_array,objectlimits_tags,objectlimits_tags_array)
16: string The 0 Schema | distance 1 The 0 Schema (tag_element,tag_string,objectlimits_tag,objectlimits_tag_string) | unrestricted
```

I don't know where it goes wrong.  Does it work for you?  I can debug if you can't figure it out.

schani:
Oh, I see - it only works with `with-getter-setter`!  I guess it wouldn't be trivial to make it work with structs, too?

If that's the case then I say we should make `with-getter-setter` the default, since it's much more fully featured now, and the structs mode is kinda shitty in comparison.  Also that will make all the tests run with `with-getter-setter`.  When you do change the default for `code-format` (which we should also rename at this point - I'll figure out a better name when the PR is merged), make sure you also change the `quickTestRendererOptions` to test `with-struct` now.

test/inputs/schema/objectlimits.schema
```diff
@@ -0,0 +1,90 @@
+{
+  "$id": "http://example.com/example.json",
+  "type": "object",
+  "properties": {
+    "checked": {
+      "$id": "/properties/checked",
```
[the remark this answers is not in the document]
harsszegi: Have zero idea, not quiet great on draft-07, I was using https://jsonschema.net/ to generate / edit the schema

src/quicktype-core/language/CPlusPlus.ts
```diff
@@ -565,11 +666,44 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
         );
     }
 
-    protected emitClassMembers(c: ClassType): void {
+    /**
+     * similar to cppType, it practically gathers all the generated types within
+     * 't'. It also records, whether a given sub-type is part of a variant or not.
+     */
+    protected generatedTypes(isClassMember:boolean, isVariant:boolean, l: number, t: Type, result:Array<TypeRecord>): void {
```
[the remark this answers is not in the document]
harsszegi: Nope, isvariant / l is set initially, but as you go down below, the value is changed and finally the "actual" stuff is stored in the record

src/quicktype-core/language/CPlusPlus.ts
```diff
@@ -580,6 +714,10 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
                     name,
                     ";"
                 );
+                if (constraints !== undefined && constraints.has(jsonName)) {
+                    const cnst = this.lookupGlobalName(GlobalNames.ClassMemberConstraints);
+                    this.emitLine(cnst," ", this._memberNameStyle(jsonName+ "Constraint"), ";");
```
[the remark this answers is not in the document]
harsszegi: Added a FIXME

src/quicktype-core/language/CPlusPlus.ts
```diff
@@ -285,17 +335,68 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
...
+    protected addGlobalName(name:string, type:GlobalNames): void {
+        this._generatedGlobalNames.set(type, this._namedTypeNameStyle(name));
+    }
+
+    protected addMemberName(name:string, type:GlobalNames): void {
+        this._generatedGlobalNames.set(type, this._memberNameStyle(name));
+    }
+
+    protected setupGlobalNames(): void {
+        this.addGlobalName("ClassMemberConstraints", GlobalNames.ClassMemberConstraints);
```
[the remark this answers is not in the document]
harsszegi: Oopsie daisy, I need some further info on this - regardless that I checked the link you included :( Like I was a dumb guy from elementary :)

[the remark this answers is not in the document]
schani: Is the link info enough or do you need more info?  If it's too much bother, I can do it, too.

schani:
@harsszegi `objectlimits.schema` is a negative test?  How so?  The schema is invalid?  We don't include invalid schemas in the test suite.  And if it is a negative test, please include a valid schema for the new features, and a positive and negative JSON.  See `minmax.schema` and the corresponding JSONs.

harsszegi:
The issue is that objectlimits1.json is configured currently for SURE to fail, if you run the test you’ll see it throws an exception, ok will check the minmax test!

schani:
@harsszegi See my comment above on failure test cases.  Add three language features `minmax`, `minmaxlen`, `pattern`, and then make one success case for that schema and one failure for each of the three features.

src/quicktype-core/language/CPlusPlus.ts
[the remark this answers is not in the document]
schani: Sure, no problem.

test/languages.ts
```diff
@@ -9,7 +9,8 @@ export type LanguageFeature =
   | "date-time"
   | "integer-string"
   | "bool-string"
-  | "uuid";
+  | "uuid"
+  | "minmax";
```
schani: You'll have to add `minmaxlength` and `pattern`, too.

harsszegi: Strange, all three have executed, although I have zero idea, whether the negative tests were executed at all :)

harsszegi: My issue was exactly that when I manually executed the three tests from commandline I thought that the .fail part would also be executed, but I guess this is not the case, right

harsszegi:
A lame question, could have missed a part of the review, that’s why I am asking; am I required to fix something still?
Thanks,

test/languages.ts
schani: `something.1.fail.feature.json` will only run if the language supports `feature`.

harsszegi: Ah understood, thanks for clarifying it

schani:
@harsszegi The `minmax` and `pattern` fail test case files are named incorrectly.  Instead of `minmax.1.json.fail.minmax.json` it has to be `minmax.1.fail.minmax.json`.  Same for `pattern`.  Other than that I'll look in a bit.

src/quicktype-core/language/CPlusPlus.ts
```diff
@@ -614,30 +763,83 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
...
-                this.emitLine(className, "() = default;");
+                if (constraints === undefined) {
+                    this.emitLine(className, "() = default;");
+                } else {
+                    this.emitLine(className, "() :");
+                    let numEmits:number = 0;
+                    constraints.forEach((initializer: string, _propName: string) => {
+                        numEmits++;
+                        if (numEmits === constraints.size) {
+                            this.emitLine("    ", initializer);
```
schani: Never explicitly emit indentation.  Use `indent()` or `emitBlock()` or whatever applies.

src/quicktype-core/language/CPlusPlus.ts
```diff
@@ -972,23 +1269,33 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
 
         if (this.haveUnions) {
             this.emitBlock(
-                [`template <typename T>\ninline ${optionalType}<T> get_optional(const json &j, const char *property)`],
+                ["template <typename T>\ninline ", optionalType, "<T> get_optional(const json &j, const char *property)"],
```
schani: Never emit newlines in strings, unless it's via `emitMultiline()`.

src/quicktype-core/language/CPlusPlus.ts
```diff
@@ -1079,166 +1388,72 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
...
-        }
+        let propTypes:Array<TypeRecord> = new Array();
+        this.generatedTypes(false, isClassMember, false, 0, propertyType, propTypes);
 
-        directlyReachableTypes<string>(propertyType, t => {
-            const typeName = this.sourcelikeToString(
-                this.cppType(
-                    t,
+        for (const t of propTypes) {
+            if (t.name === undefined) {
```
schani: `t.name` can't be `undefined` anymore, so this case goes away.

src/quicktype-core/Constraints.ts
schani: Still not resolved.

test/languages.ts
```diff
@@ -308,7 +311,7 @@ export const CPlusPlusLanguage: Language = {
     "fcca3.json"
   ],
   allowMissingNull: false,
-  features: ["enum", "union", "no-defaults"],
+  features: ["minmax", "minmaxlength", "pattern", "enum", "union", "no-defaults"],
```
schani: Please also add `"code-format": "with-struct"` to `quickTestRendererOptions`.

schani:
@harsszegi I'm refactoring the code right now.  Please don't change `CPlusPlus.ts` until I'm done, or we'll have conflicts.  We can merge tomorrow.

harsszegi:
ACKed

schani:
@harsszegi I'm done for now.  Please merge the `better-cpp` branch into your branch and push it so the PR gets updated.  I don't think I can push onto your PR.

harsszegi:
Done

schani:
Thank you for your patience!

In the future, please only small, one-feature PRs ;-)

harsszegi:
Agreed :) thanks a lot for the enormous help!

PR #1014, opened 2018-08-07
Don't emit module exports is justTypes is true

ninaniiranen:
This fixes a issue where the `--just-types` option was ignored for flow.


schani:
Great catch, thank you so much!

PR #1020, opened 2018-08-08
Replace string-hash with a Java compatible version

hansl:
It is faster, removes a dependency that is unmaintained and license-less,
and doesnt make it harder to maintain.

See https://github.com/quicktype/collection-utils/pull/3

### DIFF
--- a/build/quicktype-core/package-lock.json
+++ b/build/quicktype-core/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "quicktype-core",
-    "version": "4.0.0",
+    "version": "5.0.29",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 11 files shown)


hansl:
I might need some help figuring out the `package.in.json` and their lock files in `build/`.

schani:
@hansl Could you point me to the corresponding Java code and/or some docs about this hash function?

I'm happy to merge this, but I'll rewrite it to just use the version in collection-utils.

As for the `build` directory, check out [PACKAGES.md](https://github.com/quicktype/quicktype/blob/master/PACKAGES.md#building).  I think/hope it's up-to-date.

hansl:
Hi @schani,

I repushed after running those scripts. Not sure why it added a `package.json` in the `build/quicktype-core` directory.

For reference on Java String `hashCode()` implementation, the doc has the algorithm; https://docs.oracle.com/javase/6/docs/api/java/lang/String.html#hashCode(). Using bitwise operation is faster than multiplying.

Please note that this is the same as the `string-hash` package implementation, except that `string-hash` uses a start hash of 5381 for some reason (see https://github.com/darkskyapp/string-hash/blob/master/index.js)...

hansl:
I'm fine with reusing the `collection-utils` implementation for simplicity. Just release the PR on npm and I'll update this PR to import it from there instead.

hansl:
@schani Updated. Please take another look.

build/quicktype-core/package.json
```diff
@@ -0,0 +1,37 @@
+{
```
schani: This file is automatically generated.

hansl:
Done. Nothing to squahs, there's only one commit :)

schani:
Thank you!

PR #1044, opened 2018-08-19
Delete `yarn.lock`

### LINKED ISSUE #1042: Use pnpm or yarn or npm for development
There is both a `package-lock.json` and a `yarn.lock` committed. The `yarn.lock` is 9 months and many commits affecting `package-lock.json` and even `package.json` old.
Should the `yarn.lock` be deleted and added to `.gitignore`?

qm3ster:
fixes #1042

But maybe switch to pnpm instead.


.gitignore
```diff
@@ -1,6 +1,8 @@
 node_modules/
 /test/awesome-json-results/
 npm-debug.log
+yarn.lock
+shrinkwrap.yaml
```
schani: Why would we add `yarn.lock` if it's removed?  And what's `shrinkwrap.yaml`?

qm3ster: shrinkwrap.yaml is what happens if you `pnpm i`.
It's removed because it is currently already in the repo, and it's added to `.gitignore` because contributors using `yarn` or `pnpm i` will create these and may easily commit them again.
(which they would be absolutely right to do if the repository's source of truth was `pnpm`'s `shrinkwrap.yaml` or `yarn`'s `yarn.lock` rather as the case seems to be `npm`'s `package-lock.json` )

schani: I see.  Thank you.

PR #1077, opened 2018-09-28
Support Sourcelike descriptions

colinking:
This PR switches the type of `emitCommentLines` and the various `emitDescription` methods from `string[]` to `Sourcelike[]`.

As an example, this is useful for emitting javadoc which may rely on rendering `Sourcelike` inputs. For example:

```
this.emitDescriptionBlock(['@param ', paramName, ' foos the bar'])
```


schani:
Wonderful, thank you so much!

PR #1078, opened 2018-10-02
Make emitMark available to extending Renderers

colinking:
It may be worth considering making most, if not all, `private` methods `protected` so that these renderers can be easily extended.


schani:
@colinking I wouldn't wholesale make everything `protected`, since any change in those interfaces requires a major version bump, but I do agree that many more should be `protected` than there are now.

PR #1082, opened 2018-10-02
Make various Objective C methods protected and exported

colinking:
As an extension to https://github.com/quicktype/quicktype/pull/1078, this PR opens up various methods to make it easier to extend the Objective C renderer.

Happy to discuss whether any of these API changes don't make sense.

### DIFF
--- a/src/quicktype-core/language/Objective-C.ts
+++ b/src/quicktype-core/language/Objective-C.ts
@@ -25,7 +25,7 @@ import { RenderContext } from "../Renderer";
 
 const unicode = require("unicode-properties");
 
-type MemoryAttribute = "assign" | "strong" | "copy";
+export type MemoryAttribute = "assign" | "strong" | "copy";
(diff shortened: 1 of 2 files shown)


src/quicktype-core/language/Objective-C.ts
```diff
@@ -210,7 +210,7 @@ function isPartCharacter(utf16Unit: number): boolean {
 
 const legalizeName = utf16LegalizeCharacters(isPartCharacter);
 
-function isAnyOrNull(t: Type): boolean {
+export function isAnyOrNull(t: Type): boolean {
```
schani: If this is exported, it should go to `TypeUtils.ts`.

colinking:
Good call, I just moved `isAnyOrNull`. It'd also make sense to squash this, too.

Thanks!

schani:
Thank you!

PR #1084, opened 2018-10-03
Convert Renderer closures into methods

colinking:
This updates all closure-based methods from all the Renderers to use methods. This allows subclasses to extend these functions, which otherwise would not be accessible via `super`.

This PR should be squashed if merged.

### DIFF
--- a/src/quicktype-core/language/Elm.ts
+++ b/src/quicktype-core/language/Elm.ts
@@ -269,10 +269,10 @@ export class ElmRenderer extends ConvenienceRenderer {
         }
     }
 
-    private decoderNameForNamedType = (t: Type): Name => {
+    private decoderNameForNamedType(t: Type): Name {
         const name = this.nameForNamedType(t);
         return defined(this._namedTypeDependents.get(name)).decoder;
-    };
+    }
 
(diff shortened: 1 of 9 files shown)


src/quicktype-core/language/Elm.ts
```diff
@@ -625,26 +625,26 @@ import Dict exposing (Dict, map, toList)`);
 
         this.forEachTopLevel(
             "leading-and-interposing",
-            this.emitTopLevelDefinition,
+            (t: Type, topLevelName: Name) => this.emitTopLevelDefinition(t, topLevelName),
```
schani: I don't usually declare argument types in these anonymous functions.  In some cases TypeScript can't infer them, but those are rare.  No big deal, doesn't need correction.  The general idea is that I want to declare types at interface boundaries to avoid type drift.  In this case here I just care that the types are compatible, not what they are in detail.

schani:
Thank you, and please excuse my tardiness.

src/quicktype-core/language/Elm.ts
colinking: Good to know!

colinking:
No worries, thanks for your quick feedback on all of these PRs this week.

PR #1091, opened 2018-10-17
Add support for GraphQL mutations.

ryancerium:


schani:
@ryancerium This is fantastic!

Could you please add at least one mutation test case to `test/inputs/graphql`, just so that we're checking that it at least doesn't crash and produces code that compiles?

ryancerium:
I figured that would be the first comment @schani :-D  Yeah, I'll work on it tonight after I get home and get the kids to bed.

src/quicktype-graphql-input/index.ts
```diff
@@ -160,6 +160,7 @@ function expandSelectionSet(selectionSet: SelectionSetNode, inType: GQLType, opt
 interface GQLSchema {
     readonly types: { [name: string]: GQLType };
     readonly queryType: GQLType;
+    readonly mutationType?: GQLType;
```
schani: I don't think it matters here, but `?` has different semantics than `X | undefined`.  `X | undefined` means the property must be present but its value can be `undefined`.  `?` means the property can be missing.

schani:
@ryancerium Thank you! :-)

src/quicktype-graphql-input/index.ts
ryancerium: I think "missing" is more appropriate in this case anyway, since a GraphQL endpoint without any mutation operations wouldn't have a `mutationType`.

ryancerium:
@schani Do you want me to test an array too?  I'm using them at work, but did a very fast test in `github9.graphql`

schani:
Wonderful!  Thank you so much!

PR #1101, opened 2018-10-30
Acronym style

orm-forks:
Solution for the comment around acronyms consistently relating to https://stackoverflow.com/questions/8277355/naming-convention-for-upper-case-abbreviations

Added a new CLIoption called "acronym-style" which supports originalWord, pascal, camel and capitalize as the styling for an acronym. I've added a fallback so by default we stay with the current wordstyle. 
It's has been implemented on both swift and Java to resolve the current fixme in the JavaRender 

**Example**
--acronym-style originalWord

### DIFF
--- a/src/quicktype-core/support/Acronyms.ts
+++ b/src/quicktype-core/support/Acronyms.ts
@@ -1115,7 +1115,8 @@ export const acronymOption = new EnumOption(
         ["camel", acronymStyleOptions.camel],
         ["capitalize", acronymStyleOptions.capitalize]
     ],
-    "pascal"
+    "pascal",
+    "secondary"
 );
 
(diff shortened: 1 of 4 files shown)


src/quicktype-core/support/Acronyms.ts
```diff
@@ -1095,3 +1098,34 @@ export const acronyms: string[] = [
...
 ];
+
+export enum AcronymStyleOptions {
+    Pascal,
+    Camel,
+    Capitalize,
+    OriginalWord
+}
+
+export const acronymOption = new EnumOption(
```
schani: Could you make this into a function that takes the default as an argument?  Not all languages will have `pascal` as their default.

orm-forks: Sure, I'll do the change now

PR #1111, opened 2018-11-09
Fix rename of method in string text.

waywardmonkeys:
In 59cd3db87, a method was renamed, but not in this string text.


schani:
Thank you!

Issue #1102, opened 2018-10-31
Python conversions can be simplified if we declare from_list, from_dict etc differently


Ark-kun:
Before:

```
def from_list(f: Callable[[Any], T], x: Any) -> List[T]:
    assert isinstance(x, list)
    return [f(y) for y in x]

class MyClass:
    def from_dict(obj: Any):
        command = from_union([lambda x: from_list(from_str, x), from_none], obj.get("command"))

```
After:

```
def from_list_of(f: Callable[[Any], T]) -> Callable[[Any], List[T]]:
    def _from_list(x: Any) -> List[T]:
        assert isinstance(x, list)
        return [f(y) for y in x]
    return  _from_list

class MyClass:
    def from_dict(obj: Any):
        command = from_union([from_list_of(from_str), from_none], obj.get("command"))

```

schani:
Looks great!  Please consider contributing.  We'd be happy to help.

schani:
We do not take feature requests anymore. Please consider making a pull request instead. We encourage the use of coding agents, but please use at least an Opus-class model.

PR #1116, opened 2018-11-17
[WIP] Python - Getting rid of lambdas

Ark-kun:
As suggested in https://github.com/quicktype/quicktype/issues/1102, Python conversion expressions can be simplified by defining curried functions instead of lambdas.

When fully implemented, this would simplify both the generated code and also the generator code.

    foo = from_union([lambda x: from_list(from_str, x), from_none], obj.get("foo"))  #before
    foo = from_union_of([from_list_of(from_str), from_none])(obj.get("foo"))         #after

    result["bar"] = from_union([lambda x: to_class(Bar, x), from_none], self.bar)    #before
    result["bar"] = from_union_of([to_class_of(Bar), from_none])(self.bar)           #after

This results in fully functioning converter functions where the value is only passed in the end.

I created the modified conversion functions, but my head hurts while I try to wrap it around the `serializer`/`deserializer` functions (https://github.com/quicktype/quicktype/blob/master/src/quicktype-core/language/Python.ts#L990). I feel like the change I propose is more compatible with older versions of Python generator (e.g. https://github.com/quicktype/quicktype/commit/89423cb15a26aa43819bba60824928e88c29fc1c#diff-9678b475f763b3fd561e5fb2f238e25e) when there was no complicated recursive value passing. We could do it similarly to how the only `serializerFn` function worked.

@schani If you know the easiest way to do this, I'd like a bit of your help.

P.S. This [WIP] PR *adds* the new variants of list/dict/class/union converters, but the final goal is to replace the existing ones.

### DIFF
--- a/src/quicktype-core/language/Python.ts
+++ b/src/quicktype-core/language/Python.ts
@@ -505,9 +505,13 @@ export type ConverterFunction =
     | "str"
     | "to-enum"
     | "list"
+    | "list-of"
     | "to-class"
+    | "to-class-of"
     | "dict"
+    | "dict-of"
     | "union"
+    | "union-of"
     | "from-datetime"
     | "from-stringified-bool"
     | "is-type";
@@ -709,6 +713,35 @@ export class JSONPythonRenderer extends PythonRenderer {
         );
     }
 
+    protected emitListOfConverter(): void {
+        const tvar = this.typeVar();
+        this.emitBlock(
+            [
+                "def from_list_of(f",
+                this.typeHint(": ", this.withTyping("Callable"), "[[", this.withTyping("Any"), "], ", tvar, "]"),
+                ")",
+                this.typeHint(" -> ", this.withTyping("Callable"), "[[", this.withTyping("Any"), "], ", this.withTyping("List"), "[", tvar, "]", "]"),
+                ":"
+            ],
+            () => {
+                this.emitBlock(
+                    [
+                        "def _from_list(",
+                        this.typingDecl("x", "Any"),
+                        ")",
+                        this.typeHint(" -> ", this.withTyping("List"), "[", tvar, "]"),
+                        ":"
+                    ],
+                    () => {
+                        this.emitLine("assert isinstance(x, list)");
+                        this.emitLine("return [f(y) for y in x]");
+                    }
+                );
+                this.emitLine("return _from_list");
+            }
+        );
+    }
+
     protected emitToClassConverter(): void {
         const tvar = this.typeVar();
         this.emitBlock(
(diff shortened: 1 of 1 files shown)


schani:
I'm adding a bunch of comments to make the code marginally easier to understand.

Due to the way transformers can be nested, your `from_list_of` solution wouldn't work in the general case, though it would probably to apply to most cases.  The `lambda x: from_list(from_str, x)` in your example might in another case be `lambda x: from_list(from_str, some_other_transformer(x))`, i.e. you can't assume that the second argument to `from_list` will always be `x`.

To handle the (very common) special case, look for example at this piece of code in `deserializer`:

```
            arrayType =>
                compose(
                    value,
                    v => [
                        this.conv("list"),
                        "(",
                        makeLambda(this.deserializer(identity, arrayType.items)).source,
                        ", ",
                        v,
                        ")"
                    ]
                ),
```

The case you have to handle is `value` being `identity`.  In that case, you can return `["from_list_of(", makeLambda(this.deserializer(identity, arrayType.items)).source, ")"]`.  To that end I'd introduce a new function `composeWithShortcut` or something, which checks whether `value` is identity, and in that case uses the shortcut form.

I'll follow up with the (slightly) commented code soon.

Ark-kun:
> The lambda x: from_list(from_str, x) in your example might in another case be lambda x: from_list(from_str, some_other_transformer(x)), i.e. you can't assume that the second argument to from_list will always be x.

I do not think I'm assuming that. 
In my second example I have `from_union_of([to_class_of(Bar), from_none])`. I think that `to_class_of(Bar)` is that `some_other_transformer` that you mention.

My idea is that you can always factor all transformations without needing `x` until the very end.
Simple transforms are simple functions that can take `x` (1 argument).
Complex transforms are transform factories that take other transforms and create a combined transform (which again is a function that can take `x` (1 argument)).

I cannot think of a case where it's not possible to factor the code in this way. In a way, I just move the lambdas from your `deserializer` code to the conversion functions.

schani:
Let's stick to the first example, please, or we're talking at cross-purposes.

I can see now how you can make

    lambda x: from_list(from_str, some_other_transformer(x))

into a `lambda`-free version, if you have an `_of` version of all the functions you might be calling, in this case `from_str`.  So this would be

    from_list_of(from_str_of(some_other_transformer))

This might actually simplify the code since you would always return a transforming function, instead of either a lambda or a value.  You'd still have to special-case the identity function, however, or you'd probably end up with something like

    from_list_of(from_str_of(identity))

I'm for it!  Let me know if you need any help.

Ark-kun:
> from_list_of(from_str_of(some_other_transformer))

Can you please give an example where this happens (where an object of basic type (string, int, float etc) can come from another transformer)? Does that happen during serialization, deserialization or both?

schani:
> Can you please give an example where this happens (where an object of basic type (string, int, float etc) can come from another transformer)? Does that happen during serialization, deserialization or both?

I'm not sure whether the current transformers will ever do this, but the transformer mechanic allows it, and future transformers might use it, so we shouldn't remove that functionality.

schani:
I just thought of a transformer that would go there: validation transformers for string length.  We already support `minLength`/`maxLength` for C++, and it would be nice to have it in Python (and C#).

PR #1120, opened 2018-11-27
fix: enum cases should not be different as other word in JS

alan-agius4:


schani:
Sorry, buildkite doesn't allow public access.  Nothing is technically breaking - the `diff` step just compares the output of the last build with this build's output, and since the casing changed, it complained.  Here's an example:

```
@@ -64,7 +64,7 @@
     condition:   Condition;
     description: string;
     forecast:    Forecast[];
-    guid:        GUID;
+    guid:        Guid;
     lat:         string;
     link:        string;
     long:        string;
```

I didn't have time to check your previous PR, but I just did, and it's a bit incomplete, which also relates to this one:

TypeScript and Flow have a `nice-property-names` option with its own naming style, which should also use the selected acronym style.  Example:

```
Marks-iMac:quicktype schani$ ./script/quickesttype --lang typescript --nice-property-names --just-types --acronym-style lowerCase /tmp/bla.json 
export interface Bla {
    myURL: Myurl;
}

export interface Myurl {
    fooURL: string;
}
Marks-iMac:quicktype schani$ cat /tmp/bla.json 
{
	"my_url": {
		"foo_url": "abc"
	}
}
```

Regarding this PR, it just seems to hard-code one particular acronym style, vs using the one the user selected via the option.  If they want acronyms to be all-upper, then you shouldn't override that.  There's a bit of an added difficulty in that if `upper` is set, the first word should be upper-case, even if the acronym style is all-lower, so you should probably call `capitalize` on the result of `acronymsStyle`.

alan-agius4:
@schani I updated the PR, hopefully it's what you had in mind.

src/quicktype-core/language/JavaScript.ts
```diff
@@ -83,7 +84,7 @@ export function nameStyle(
         legalizeName,
         upper ? firstUpperWordStyle : allLowerWordStyle,
         firstUpperWordStyle,
-        upper ? allUpperWordStyle : allLowerWordStyle,
+        upper ? s => capitalize(acronymsStyle(s)) : acronymsStyle,
```
schani: The lower option should always be `allLowerWordStyle`.  Otherwise, if you selected Pascal, you'd get `fooData` for `foo_data`, but `JSONData` for `json_data`.

alan-agius4: Pushed :)

schani:
Yes, that's what I had in mind, modulo the comment I left.

Would you also mind fixing the name style for TypeScript/Flow?  Seems you only have to pass in the `acronymsStyle` parameter to `nameStyle` [here](https://github.com/quicktype/quicktype/blob/bf899cb4eab8ef139867c46030c0a8093fb53dc8/src/quicktype-core/language/TypeScriptFlow.ts#L87).  I suggest making `nameStyle` a `protected` method in `JavaScriptRenderer`, so you don't even have to pass that argument.

Issue #1144, opened 2018-12-31
Accept JSON5 as input


andyhu:
JSON is a pretty strict format, even changing double quote to single quote fails validation. It happened many time to me that what I really need to input is not a strict JSON but a Javascript Object. In this regards I think it would be quite convenient if adding JSON5 as input source, which is 100% compatible with standard JSON, but with steroids to make most normal Javascript Objects valid as input.
Just for reference, here is the specification of JSON5 https://json5.org/

schani:
Would you consider contributing this feature?  We'd be happy to help.

andyhu:
Thanks for the reply. I've made an attempt, but it doesn't seem to work. What could be the issue?

schani:
We do not take feature requests anymore. Please consider making a pull request instead. We encourage the use of coding agents, but please use at least an Opus-class model.

PR #1145, opened 2018-12-31
Add JSON5 support as input

andyhu:
First attempt to switch from JSON to JSON5 as discussed in #1144


andyhu:
I accidentally deleted the original fork so I created a new one. I think the size might not a problem. I'll test it on local machine when I have time.

PR #1149, opened 2019-01-08
Pike

mkrawczuk:
This is a pull request I've been suggested to do to receive feedback. All kinds of criticism are more than appreciated.

Among other TODO's, the most important ones for now are:
* decoding JSON data into an object type,
* determining to which type should be the JSON data decoded into,
* finishing integration with the test suite.

### DIFF
--- a/src/quicktype-core/language/Pike.ts
+++ b/src/quicktype-core/language/Pike.ts
@@ -1,0 +1,118 @@
+"use strict";
+
+import { anyTypeIssueAnnotation, nullTypeIssueAnnotation } from "../Annotation";
+import { ConvenienceRenderer } from "../ConvenienceRenderer";
+import { Name, Namer, funPrefixNamer } from "../Naming";
+import { Option } from "../RendererOptions";
+import { RenderContext } from "../Renderer";
+import { Sourcelike, maybeAnnotated } from "../Source";
+import { TargetLanguage } from "../TargetLanguage";
+import { Type, ClassType } from "../Type";
+import { matchType, nullableFromUnion } from "../TypeUtils";
+
+import {
+  legalizeCharacters,
+  isLetterOrUnderscoreOrDigit,
+  stringEscape,
+  makeNameStyle
+} from "../support/Strings";
+
+export const pikeOptions = {
+};
+
+const legalizeName = legalizeCharacters(isLetterOrUnderscoreOrDigit);
+
+const namingFunction = funPrefixNamer("namer", makeNameStyle("underscore", legalizeName));
+const namedTypeNamingFunction = funPrefixNamer("namer", makeNameStyle("pascal", legalizeName));
+
+export class PikeTargetLanguage extends TargetLanguage {
+  constructor() {
+    super("Pike", ["pike", "pikelang"], "Pike");
+  }
+  protected getOptions(): Option<any>[] {
+    return [];
+  }
+
+  protected makeRenderer(renderContext: RenderContext): PikeRenderer {
+    return new PikeRenderer(this, renderContext);
+  }
+}
+
+export class PikeRenderer extends ConvenienceRenderer {
+
+  protected emitSourceStructure(): void {
+    this.forEachObject(
+      "leading-and-interposing",
+      (c: ClassType, className: Name) => this.emitClass(c, className));
+  }
+
+  protected makeEnumCaseNamer(): Namer {
+    return namingFunction;
+  }
+  protected makeNamedTypeNamer(): Namer {
+    return namedTypeNamingFunction;
+  }
+
+  protected makeUnionMemberNamer(): Namer {
+    return namingFunction;
+  }
+
+  protected namerForObjectProperty(): Namer {
+    return namingFunction;
+  }
+
+  protected forbiddenNamesForObjectProperties(): string[] {
+    return ["private", "protected", "public", "return"];
+  }
+
+  private emitBlock = (line: Sourcelike, f: () => void): void {
+    this.emitLine(line, " {");
+    this.indent(f);
+    this.emitLine("}");
+  };
+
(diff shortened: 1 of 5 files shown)


src/quicktype-core/language/Pike.ts
```diff
@@ -0,0 +1,178 @@
+"use strict";
```
schani: This is not necessary.

schani:
Thank you so much!  I haven't looked at the code yet - I'll do that later today.

In the mean time, for Pike to run in our CI you'll have to add two more things:

* Add fixtures for JSON, JSON Schema, and GraphQL in `test/fixtures.ts`.
* Add the JSON and JSON Schema fixtures to `.buildkite/pipeline.yml`, in the `csharp ruby golang` step.

schani:
@mkrawczuk It fails on a negative JSON Schema test (i.e. the input doesn't match the schema and is supposed to not "work", but it does).  You can run the schema tests on your machine with

```
FIXTURE=schema-pike npm test
```

Here is the failure message:

```
Expected failure on input /app/test/inputs/schema/class-map-union.1.fail.union.json
--
  | {
  | "result": "{\n  \"union\": {\n    \"foo\": {\n      \"quux\": \"foo\"\n    },\n    \"bar\": {\n      \"quux\": \"bar\"\n    }\n  }\n}\n",
  | "cwd": "/app/test/runs/schema-pike-9f2e73"
  | }
  | Fixture threw an exception
  | {
  | "error": {
  | "result": "{\n  \"union\": {\n    \"foo\": {\n      \"quux\": \"foo\"\n    },\n    \"bar\": {\n      \"quux\": \"bar\"\n    }\n  }\n}\n",
  | "cwd": "/app/test/runs/schema-pike-9f2e73"
  | },
  | "sample": {
  | "path": "test/inputs/schema/class-map-union.schema",
  | "saveOutput": true
  | },
  | "cwd": "/app/test/runs/schema-pike-9f2e73"
  | }
  | [object Object]
  | at /app/test/test.ts:68:17
  | at step (/app/test/test.ts:32:23)
  | at Object.throw (/app/test/test.ts:13:53)
  | at rejected (/app/test/test.ts:5:65)
  | at <anonymous>
  | at process._tickDomainCallback (internal/process/next_tick.js:228:7)
  | npm ERR! Test failed.  See above for more details.
```

schani:
@mkrawczuk Do you need help on this?

mkrawczuk:
I've finished integration with the test pipeline and consider basic Pike support done.
There are, however, some issues:
* Wide strings are not supported in Pike's `write` function. Not much can be done about that in quicktype's test suite scope, so I've disabled test cases containing them.
* Some trouble appear around JSONs with non-UTF8 contents. Disabled them also.
* Since Pike's enum values are directly available in the scope they've been declared in, there were some naming collisions in cases where there would be different enums with the same values. For example, `us-senators.json` fails because output contains enums `enum Gender {MALE = "Male", FEMALE = "Female}` and `enum GenderLabel {MALE = "Male", FEMALE = "Female}` . This causes a collision between `Gender`'s `MALE` and `GenderLabel`'s `MALE`, and between `Gender`'s `FEMALE` and `GenderLabel`'s `FEMALE` Disabled similar cases. 
One solution for this problem would be to prefix an enum value with an enum name, so the prevous example would look something like 
```
enum Gender {Gender_MALE = "Male", Gender_FEMALE = "Female"}
enum GenderLabel {GenderLabel_MALE = "Male", GenderLabel_FEMALE = "Female"}
```
How can I achieve such effect?

It is somehow comforting that the test cases disabled for Pike can be also seen disabled for other languages, so they probably display a more general problem.

test/languages.ts
```diff
@@ -830,3 +830,88 @@ export const DartLanguage: Language = {
...
+    return `pike main.pike \"${sample}\"`;
+  },
+  diffViaSchema: false,
+  skipDiffViaSchema: [],
+  allowMissingNull: true,
+  features: ["union"],
+  output: "TopLevel.pmod",
+  topLevel: "TopLevel",
+  skipJSON: [
+    "github-events.json",
```
schani: Why are all of these skipped?

src/quicktype-core/language/Pike.ts
```diff
@@ -0,0 +1,290 @@
...
+        this.forEachClassProperty(c, "none", (name, jsonName, p) => {
+            const pikeType = this.sourceFor(p.type).source;
+
+            table.push([[pikeType, " "], [name, "; "], ['// json: "', stringEscape(jsonName), '"']]);
+        });
+        this.emitTable(table);
+    }
+
+    private emitInformationComment() {
+        this.emitLine(["// This source has been automatically generated by quicktype."]);
```
schani: Please use `emitCommentLines` for this.

schani:
Regarding the disabled JSONs: Please at least add comments describing why those JSONs are disabled.  The number of disabled JSONs suggests there are still some fundamental bugs that will show up in practice often.

Regarding the enums: What you probably want is [this](https://github.com/quicktype/quicktype/blob/master/src/quicktype-core/language/Elm.ts#L209).

mkrawczuk:
Thanks, @schani ! Exactly what I needed.
I've been able to get rid of most of the bugs, and enabled most of previously disabled tests. Also did what you asked me for.

test/languages.ts
mkrawczuk: Un-skipped most of them, described reasons for remaining ones.

mkrawczuk:
Also thank you @schani for all the help! Adding Pike support to quicktype was just a step in achieving my main objective, but I so fell in love with this project that I'll be really glad to spend some more free time on contributing to it later in the future.

schani:
I can't wait to see what's next!

PR #1163, opened 2019-02-03
Java: Option to "Use T[] or List<T>" (similar to C#)

### LINKED ISSUE #1161: Java: Option to "Use T[] or List<T>" (similar to C#)
For C# there's already an option `--array-type array|list` to control how JSON-Schema type `array` should be translated. Having this option also available for Java would be great.

about-code:
Closes #1161

### DIFF
--- a/src/quicktype-core/language/Java.ts
+++ b/src/quicktype-core/language/Java.ts
@@ -1,29 +1,15 @@
-import { TypeKind, Type, ArrayType, MapType, EnumType, UnionType, ClassType, ClassProperty } from "../Type";
-import { matchType, nullableFromUnion, removeNullFromUnion, directlyReachableSingleNamedType } from "../TypeUtils";
-import { Sourcelike, maybeAnnotated } from "../Source";
-import {
-    utf16LegalizeCharacters,
-    escapeNonPrintableMapper,
-    utf16ConcatMap,
-    standardUnicodeHexEscape,
-    isAscii,
-    isLetter,
-    isDigit,
-    capitalize,
-    splitIntoWords,
-    combineWords,
-    allUpperWordStyle,
-    firstUpperWordStyle,
-    allLowerWordStyle
-} from "../support/Strings";
-import { Name, Namer, funPrefixNamer, DependencyName } from "../Naming";
-import { ConvenienceRenderer, ForbiddenWordsInfo } from "../ConvenienceRenderer";
-import { TargetLanguage } from "../TargetLanguage";
-import { BooleanOption, StringOption, EnumOption, Option, OptionValues, getOptionValues } from "../RendererOptions";
 import { anyTypeIssueAnnotation, nullTypeIssueAnnotation } from "../Annotation";
-import { defined, assert, assertNever } from "../support/Support";
+import { ConvenienceRenderer, ForbiddenWordsInfo } from "../ConvenienceRenderer";
+import { DependencyName, funPrefixNamer, Name, Namer } from "../Naming";
 import { RenderContext } from "../Renderer";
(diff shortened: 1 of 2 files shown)


src/quicktype-core/language/Java.ts
```diff
@@ -303,7 +309,7 @@ export class JavaRenderer extends ConvenienceRenderer {
     }
 
     protected emitPackageAndImports(imports: string[]): void {
-        const allImports = ["java.util.Map"].concat(this._options.justTypes ? [] : imports);
+        const allImports = ["java.util.*"].concat(this._options.justTypes ? [] : imports);
```
schani: Is this because we might need `List`, too?  Is this a standard thing to do in Java?

src/quicktype-core/language/Java.ts
```diff
@@ -350,7 +362,11 @@ export class JavaRenderer extends ConvenienceRenderer {
 
     protected javaTypeWithoutGenerics(reference: boolean, t: Type): Sourcelike {
         if (t instanceof ArrayType) {
-            return [this.javaTypeWithoutGenerics(false, t.items), "[]"];
+            if (this._options.useList) {
+                return [this.javaTypeWithoutGenerics(true, t.items)];
```
schani: This seems weird.  Shouldn't this be `List`?

about-code: Yeah you're right. I probably wrote it in [line 467](https://github.com/quicktype/quicktype/blob/47dd21307560cd515a306a40fb0eab03b224d2a5/src/quicktype-core/language/Java.ts#L467) because of the name of them method *javaTypeWithoutGenerics()*. `List<T>` is called a generic type in Java. But I will see what changes if I put it into the method. If everything is fine I can change it. Just to be sure, so you propose to change it to the following?
```ts
return ["List<", this.javaTypeWithoutGenerics(true, t.items), ">"]
```

schani: No, it seems that this should just be `List`, like it is for `Map` a few lines further down.  Maybe that doesn't work in Java?

src/quicktype-core/language/Java.ts
about-code: Just to add: another reason why the wildcard wasn't much of a concern to me is because there have already been wildcard imports, e.g. `com.fasterxml.jackson.annotation.*` in the code. So if its the wildcard which you would like being avoided, just let me know. But getting rid of wildcard imports as a whole would then probably be a different issue.

schani:
@about-code Do you need help to move this forward?

src/quicktype-core/language/Java.ts
about-code: @schani Sorry for being a bit unresponsive. I am only able to work on this on weekends and unfortunately I crashed my local dev environment when trying to replace docker with podman. Your proposal to just return `List` doesn't work. E.g. in `Deserializer` code we'd end up with

```
jsonParser.readValueAs(new TypeReference<List>() {});
```
instead of
```
jsonParser.readValueAs(new TypeReference<List<Bar>>() {});
```
However, Jackson requires `Bar` to being able to infer the type of the elements of a list with the correct type. I am not sure why returning just `"Map"` works but in general there are a few differences between the array cases and the Map cases. Eventually I now changed the code to 

```java
return ["List<", this.javaTypeWithoutGenerics(true, t.items), ">"];
```
and would submit this as eventual PR from my side. I would leave any further modifications up to you or the community.

schani:
Thank you so much!

about-code:
You're welcome. Has been a pleasure to contribute. Thank you for your feedback, merging and maintaining.

Issue #1189, opened 2019-03-22
Refactor Typescript generation to not use Namespaces


wgottschalk:
Babel 7 parses typescript and can confirm that they refuse to implement namespaces due to the fact that they are legacy. https://github.com/babel/babel/issues/8244

When parsing from JSON schema to typescript the `Convert` function is wrapped in a namespace.

Solution:

Refactor the Convert function to be a const instead of a namespace. Do not export any of the helpers still.

before:
```ts
export namespace Convert {
  export function toThing(data: string): Thing {...}
  export function thingToJSON(data: Thing): string {...}
}
```

after:

```ts
export const Convert = {
  toThing() {...}
  thingToJSON() {...}
}

// rest of the helpers go here and are not exported
```

schani:
Would you consider contributing a PR that does this?  We'd love to help you along.  The code that emits TypeScript is [here](https://github.com/quicktype/quicktype/blob/master/src/quicktype-core/language/JavaScript.ts) and [here](https://github.com/quicktype/quicktype/blob/master/src/quicktype-core/language/TypeScriptFlow.ts).

wgottschalk:
Sounds good to me. I'll talk a look at this

PR #1192, opened 2019-03-30
updates typescript renderer to have babel 7 compliant output

wgottschalk:
## Background
The current output is not Babel 7 compliant.
https://github.com/quicktype/quicktype/issues/1189

## Changes
Pull the `_jsOptions.runtimeCheck` output into a helper method. This is needed to decouple the output from the actual convert functions.

Converts the `namespace` block in the TS renderer to a class and updates the convert methods to be static methods. I think I like this better than `const` actually. It's also slightly simpler because of the commas needed when defining methods on an object literal.

Below is a screen shot of a sample output:
image

### DIFF
--- a/src/quicktype-core/language/JavaScript.ts
+++ b/src/quicktype-core/language/JavaScript.ts
@@ -273,6 +273,9 @@ export class JavaScriptRenderer extends ConvenienceRenderer {
                 }
             });
         });
+    }
+
+  protected emitConvertModuleHelpers(): void { 
         if (this._jsOptions.runtimeTypecheck) {
             const {
                 any: anyAnnotation,
@@ -414,7 +417,7 @@ function r(name${stringAnnotation}) {
 `);
             this.emitTypeMap();
         }
-    }
+  }
 
     protected emitConvertModule(): void {
         this.ensureBlankLine();
@@ -478,6 +481,8 @@ function r(name${stringAnnotation}) {
         this.emitTypes();
 
         this.emitConvertModule();
+
+        this.emitConvertModuleHelpers() 
 
         this.emitModuleExports();
     }
--- a/src/quicktype-core/language/JavaScript.ts
+++ b/src/quicktype-core/language/JavaScript.ts
@@ -1,6 +1,13 @@
 import { arrayIntercalate } from "collection-utils";
 
-import { TransformedStringTypeKind,  PrimitiveStringTypeKind, Type, ClassProperty, ClassType, ObjectType } from "../Type";
+import {
+    TransformedStringTypeKind,
+    PrimitiveStringTypeKind,
+    Type,
+    ClassProperty,
+    ClassType,
+    ObjectType
+} from "../Type";
 import { matchType, directlyReachableSingleNamedType } from "../TypeUtils";
 import { acronymOption, acronymStyle, AcronymStyleOptions } from "../support/Acronyms";
 import {
@@ -49,10 +56,7 @@ export class JavaScriptTargetLanguage extends TargetLanguage {
     }
 
     protected getOptions(): Option<any>[] {
-        return [
-            javaScriptOptions.runtimeTypecheck,
-            javaScriptOptions.acronymStyle
-        ];
+        return [javaScriptOptions.runtimeTypecheck, javaScriptOptions.acronymStyle];
     }
 
     get stringTypeMapping(): StringTypeMapping {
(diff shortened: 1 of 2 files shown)


schani:
Will this change break existing code if its types are generated with `class` instead of `namespace`?

@colinking Any opinions about this wrt Typewriter?

colinking:
@schani we don't emit any JSON conversion logic in our JS clients, so this shouldn't affect us! Thanks for the heads up

wgottschalk:
@schani looks like the build fails because (my guess) is that there's an output miss match. Is there a place I need to go to update the output to get the tests to pass?

schani:
No, you're emitting incorrect code for Flow:

```
Error ----------------------------------------------------------------------------------------------- TopLevel.js:406:16
  |  
  | Unexpected token :
  |  
  | 406\| toTopLevel(json: string): TopLevel {
  | ^
  |  
  |  
  |  
  | Found 1 error
  |  
  | Command failed
  | {
  | "command": "flow check 1>&2 && flow-node main.js \"/app/test/inputs/json/priority/combinations1.json\"",
  | "code": 2,
  | "cwd": "/app/test/runs/flow-227877"
  | }
```

wgottschalk:
@schani got the languages to build correctly. Any guidance on how to update the diff? I tried searching around the test suite and couldn't find any information.

schani:
There's nothing that can be done about the diff signaling a build failure - Buildkite doesn't support an intermediate "Warning" level for builds.  It just means there's a diff we have to look at before we merge.  In this case it looks like the change introduces weird changes in the output.  Boilerplate code is added to all files, it seems.  Here's the diff:

https://s3.amazonaws.com/quicktype-outputs/diffs/8b6c014210f0585daf21ce64f57a06f8510ebef5-6187fb0508d87ae39bd521b9dc28efe712bf930f.diff

wgottschalk:
Ah. I see. That's actually expected. If you look at a subset of the output currently:
```ts
export namespace Convert {
    export function toCoordinate(json: string): Coordinate {
        return cast(JSON.parse(json), r("Coordinate"));
    }

    export function coordinateToJson(value: Coordinate): string {
        return JSON.stringify(uncast(value, r("Coordinate")), null, 2);
    }

    function invalidValue(typ: any, val: any): never {
        throw Error(`Invalid value ${JSON.stringify(val)} for type ${JSON.stringify(typ)}`);
    }

    function jsonToJSProps(typ: any): any {
        if (typ.jsonToJS === undefined) {
            var map: any = {};
            typ.props.forEach((p: any) => map[p.json] = { key: p.js, typ: p.typ });
            typ.jsonToJS = map;
        }
        return typ.jsonToJS;
    }
  //additional fields...
}
```

`invalidValue`, `jsonToJSProps` are indendented because the `namespace` keyword just generates an IFFE and returns all of the exported values. The unexported values are kept in the closure.

I wanted to make sure those addition functions didn't get exported with the class so I moved them outside of the classes' definition. Only the class and public methods are exported and the rest remain in the module. This means we have to dedent all of that output so that's why the diff is so huge.

```ts
export class Convert {
    public static toCoordinate(json: string): Coordinate {
        return cast(JSON.parse(json), r("Coordinate"));
    }

    public static coordinateToJson(value: Coordinate): string {
        return JSON.stringify(uncast(value, r("Coordinate")), null, 2);
    }
}

// none of these functions are exported. They remain private to the module
// however, they get dedented so that's why the diff is so large
function invalidValue(typ: any, val: any): never {
    throw Error(`Invalid value ${JSON.stringify(val)} for type ${JSON.stringify(typ)}`);
}

function jsonToJSProps(typ: any): any {
    if (typ.jsonToJS === undefined) {
        var map: any = {};
        typ.props.forEach((p: any) => map[p.json] = { key: p.js, typ: p.typ });
        typ.jsonToJS = map;
    }
    return typ.jsonToJS;
}
//additional fields...
```

schani:
Yes, but that's not the issue.  The issue is that that boilerplate code is added where it wasn't present before.  Just look at the first file in that diff.  There are no deletions, only additions.

wgottschalk:
Yeah. There definitely shouldn't be additions without the deletions. I'm trying to repro this locally. I'm running the cli in my fork of the repo and comparing the outputs.
```
$ ts-node --project src/cli/tsconfig.json ./src/cli/index.ts --src-lang json -o my_output.ts test/inputs/json/misc/00ec5.json 

$ quicktype --src-lang json -o current_output.ts test/inputs/json/misc/00ec5.json 
```

running `diff current_output.ts my_output.ts` and I am seeing deletions. Am I missing something from the CLI arguments?

schani:
This is extremely odd.  I don't understand what's going on there.  It seems like the output we have from `master` is completely missing the verification code.  I'll merge this, and revert if bad things happen.

PR #1193, opened 2019-04-02
exposes options for all languages so that the renders can be extended

wgottschalk:
I read the article about [customizing quicktype](http://blog.quicktype.io/customizing-quicktype/) and I found the guide really helpful. I want to be able to extend other languages but the options object was not exposed. I could hack around this but defining my own but then I'd have to keep them in sync.

This PR just makes sure that all languages with options export them through the index file so that renderers/language targets can be extended


schani:
Thank you so much!

Issue #892, opened 2018-06-10
imports not followed when quicktyping TypeScript input in autotune


dvdsgl:
```ts
// This file contains models that serialize to JSON
// We use quicktype to generate validators.

export type SerializedState = {
    /** @TJS-type integer */
    lastInitialized: number;
    experimentPicks: { [experiment: string]: string };
    outcomes: Outcomes;
};

// Unfortunately this is duplicated from ClientAPI until
// quicktype can make this work:
//
//  import { Outcomes } from "./ClientAPI";
//
export type Outcomes = {
    [experimentName: string]: {
        epsilon: number;
        bestOption: string;
    };
};
```

kasperisager:
Just hit this as well. An approach similar to https://github.com/YousefED/typescript-json-schema would be great: Specify a TypeScript input in addition to a top-level type for which a schema is then generated. This way, imports could be resolved without potentially generating unintended definitions.

PR #1195, opened 2019-04-03
[fix: #892] fix: enable to parse import modules using typescript-source-mode

euxn23:
Refered in #892 ,
typescript source mode can't use `import` statement to use other file,
throws `Error: TypeScript error: Cannot find module 'module-name'..`

I fixed this problem using `file full path` **instead of `baseName` and source file text** to TS Compiler API resolve module, not to readFile myself and delegate TS Compiler API.

I'm using typescript source mode in production, so I'd be grad if you merge as soon as possible.


euxn23:
Dear @schani ,

Would you check this p-r please? I'm using this in production.

Kind regard.

thegoldenmule:
I'm not terribly familiar with TS, but tests pass. Is this a breaking change @euxn23 ?

euxn23:
No, this is not breaking change.

mfrtrifork:
+1 I would like to have this feature.

mihilmy:
This is a 1 year old PR. We really need this!

mfrtrifork:
Dear @thegoldenmule and @untitaker 
What is the status of this PR? Is there someting preventing this from being merged?
Best regards, Michael

src/quicktype-typescript-input/index.ts
```diff
@@ -18,93 +16,14 @@ const compilerOptions: ts.CompilerOptions = {
...
-    if (Array.isArray(sources)) {
-        fileNames = sources;
-        host = ts.createCompilerHost(compilerOptions);
-    } else {
-        fileNames = Object.getOwnPropertyNames(sources);
-        host = new CompilerHost(compilerOptions, sources);
-    }
-
-    const program = ts.createProgram(fileNames, compilerOptions, host);
+export function schemaForTypeScriptSources(sourceFileNames: string[]): JSONSchemaSourceData {
```
schani, line 25: Removing the `sources: { [fileName: string]: string }` argument overload breaks the quicktype web app.

schani (changes requested):
This change breaks the quicktype API, in particular it will break the quicktype web app.

src/quicktype-typescript-input/index.ts
dvdsgl: No longer used in web app

PR #1196, opened 2019-04-04
Wrap function definitions in a fixed namespace with include guards

cartwrightluke:
Wrapped the optional type helpers in an include guard, since they're in a fixed namespace.

Previously these types caused conflicts when trying to include multiple generated code files, as they were exactly identical, even when namespaces differs. Now, they will only be included once, regardless of how many code files are included.

### DIFF
--- a/src/quicktype-core/language/CPlusPlus.ts
+++ b/src/quicktype-core/language/CPlusPlus.ts
@@ -1631,21 +1631,28 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
     }
 
     protected emitOptionalHelpers(): void {
-        this.emitLine("template <typename T>");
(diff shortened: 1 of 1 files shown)


cartwrightluke:
### Example
Generate: 
one.schema -> one.h
two.schema -> two.h
(the contents of the schema doesn't matter, so long as both have optional types)

In mycode.cpp add:
```
#include "one.h"
#include "two.h"
```

Previously, this resulted in build break. Now it will succeed, and both can be used.

cartwrightluke:
Diffs to the existing files are expected, so long as those changes are only the addition of include guards. There's no downside to adding them, so I don't think that they need to be behind a configuration flag.

schani:
Thank you! :-)

PR #1198, opened 2019-04-05
Fix Swift 5 compiler warnings

jimisaacs:
- We do not have a way to disable these warnings, so this is just being nice and silencing them.

### DIFF
--- a/src/quicktype-core/language/Swift.ts
+++ b/src/quicktype-core/language/Swift.ts
@@ -694,7 +694,7 @@ encoder.dateEncodingStrategy = .formatted(formatter)`);
 
         this.emitBlockWithAccess(["extension ", className], () => {
             if (isClass) {
-                this.emitBlockWithAccess("convenience init(data: Data) throws", () => {
+                this.emitBlock("convenience init(data: Data) throws", () => {
                     this.emitLine("let me = try newJSONDecoder().decode(", this.swiftType(c), ".self, from: data)");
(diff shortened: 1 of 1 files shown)


schani:
I'm sorry, I don't understand the purpose of this PR.  It seems to generate exactly the same code with the default settings, and make the access-level setting not emit the access level anymore.  Could you please explain what the intended purpose is?  What are the warnings that it fixes, and how does it fix them?

jimisaacs:
In extensions, members take on the default access of that extension, so when an extension is public, the methods it provides are already public unless specified otherwise.

So in Xcode 10.2, there’s a new warning that catches specifying repeated access and calls out the redundancy.

It’s actually a warning that SwiftLint has had for  while. The difference being that SwiftLint can be configured to ignore, while Xcode cannot.

We have a pretty large file generated from quicktype. One that now contains around 50-60 of these warnings. As I said before, there is no way to turn these warnings off. Not in Swift anyway, so trying to be nice and have quicktype  just not generate the redundancies.

lobianco:
Related: https://github.com/quicktype/quicktype/issues/1199

jimisaacs:
I think I should be clear too, this is unrelated to Swift 5 specifically, but Xcode 10.2, as it still generates the warnings even for Swift 4.2

lobianco:
@jimisaacs thanks, updated the issue 👍

schani:
Thank you so much!

Issue #1204, opened 2019-04-11
Proposal: Change Dart function names and move into class. 


ozum:
# Situation

Currently generated Dart code produces 2 top level functions and 1 static method and 1 method. They have nearly identical names, but accept and output different data types.

For example, when I see `fromJson`, I intuitively think both function and method accepts JSON string.

```dart
Category itemFromJson(String str) {}
String itemToJson(Item data) {}

class Item {
    factory Item.fromJson(Map<String, dynamic> json) => new Item();
    Map<String, dynamic> toJson() => {};
}
```

* `itemFromJson` accepts `string` and produces `Item`
* `Item.fromJson` accepts `map` and produces `Item`
* `itemToJson` accepts `Item` and produces `string`
* `item.toJson` produces `Map`

```dart
itemFromJson('{"name": "someItem"}');
Item.fromJson('{"name": "someItem"}'); // ERROR
```

# Proposal

I propose to combine all methods in Class and name it appropriately based on what they accept and output for a more consistent experience.

For example:

```dart
class Item {
    factory Item.fromRawJson(string json) => new Item();
    factory Item.fromJson(Map<String, dynamic> json) => new Item();

    String toRawJson() => {};
    Map<String, dynamic> toJson() => {};
}
```

```dart
Item.fromRawJson('{"name": "someItem"}'); // OK
Item.fromJson({"name": "someItem"}); // OK
```

If it is OK, I could try implementing this and create a PR.

EDITED: Naming convention is edited according to official JSON examples in flutter.dev website.

# Options

Proposed solution could be provided as optional. For example:

**Put encoder & decoder in Class**: If set true, functions are put into class. Otherwise behaves as before.

**Use method names fromMap() & toMap()**: If set true, methods are named as "fromMap()" and "toMap()" instead of "fromJson()" & "toJson()". See: https://github.com/flutter/flutter/issues/30956

PR #1206, opened 2019-04-14
[dart] optionally move functions into class and provide alternative names

ozum:
As proposed in #1204, this PR optionally moves functions into class and provides alternative names.

## Example

```json
{
  "name": "Pencil"
}
```

### DEFAULT: codersInClass: false, methodNamesWithMap: false

Same result as before:

```dart
Item itemFromJson(String str) => Item.fromJson(json.decode(str));

String itemToJson(Item data) => json.encode(data.toJson());

class Item {
    factory Item.fromJson(Map<String, dynamic> json) => new Item(...);

    Map<String, dynamic> toJson() => {...};
}
```

### codersInClass: true, methodNamesWithMap: false

```dart
class Item {
    factory Item.fromRawJson(String str) => Item.fromJson(json.decode(str));
    
    String toRawJson() => json.encode(toJson());

    factory Item.fromJson(Map<String, dynamic> json) => new Item(...);

    Map<String, dynamic> toJson() => {...};
}
```

### codersInClass: true, methodNamesWithMap: true

```dart
class Item {
    factory Item.fromJson(String str) => Item.fromMap(json.decode(str));

    String toJson() => json.encode(toMap());

    factory Item.fromMap(Map<String, dynamic> json) => new Item(...);

    Map<String, dynamic> toMap() => {...};
}
```

### DIFF
--- a/src/quicktype-core/language/Dart.ts
+++ b/src/quicktype-core/language/Dart.ts
@@ -468,7 +468,7 @@ export class DartRenderer extends ConvenienceRenderer {
                 this.ensureBlankLine();
                 this.emitLine(
                     "String ",
-                    this._options.methodNamesWithMap ? "toJson() => " : "toRawJson() =>",
+                    this._options.methodNamesWithMap ? "toJson() => " : "toRawJson() => ",
                     "json.encode(",
                     this._options.methodNamesWithMap ? "toMap" : "toJson",
                     "());"
--- a/src/quicktype-core/language/Dart.ts
+++ b/src/quicktype-core/language/Dart.ts
@@ -36,10 +36,11 @@ import { anyTypeIssueAnnotation, nullTypeIssueAnnotation } from "../Annotation";
 import { defined } from "../support/Support";
 import { RenderContext } from "../Renderer";
 import { arrayIntercalate } from "collection-utils";
-import { pythonOptions } from "./Python";
 
 export const dartOptions = {
-    justTypes: new BooleanOption("just-types", "Types only", false)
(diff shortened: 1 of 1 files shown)


ozum:
Hi @schani,

My dart tests in local setup pass, but unlike my previous PR, even `buildkite/quicktype-pr/swift-rust-elm-dart` automatic dart test seems to fail.

Could you please check?

```
script/dev
FIXTURE=dart npm test
```

Kind regards,

schani:
This is some weird package installation issue.  I'll look into it.

schani:
@ozum Everything works now.  [Here's the diff](https://s3.amazonaws.com/quicktype-outputsffs/d215e55c42f3d7b746401f8d9956617cb1eb986e-50de7abce8c7178d5897000f23b5530feecdfea4.diff).

Shouldn't these two options just combined into one "nicer code" option?  I'd like to make it the default, but that would break backward compatibility which I'd rather not do.

ozum:
> Shouldn't these two options just combined into one "nicer code" option?

I deliberately made it two separate options. I prefer single option (third example), but most docs, including official Dart docs, use `fromJson` convention. So even after putting methods into class, some may still prefer `fromJson` and `fromRawJson` (second example). And of course there is first example for backward compatibility.

ozum:
@schani diff link seems broken.

schani:
@ozum Fixed.  Sorry.

ozum:
Thanks @schani. Seems OK.

petrpan26:
Thanks

Issue #1148, opened 2019-01-04
Allow adding @objcMembers and NSObject conformance to generated Swift classes


acruis:
For a project in transition from Objective-C to Swift, there are scenarios where the generated model classes (which preferably are in Swift) have to be used by new code in legacy files (which are in Objective-C). Unless the class is appended with `@objcMembers` and inheriting `NSObject`, I believe the generated class will be unavailable for use by the Objective-C side.

PR #1213, opened 2019-04-25
Add the ability to generate Swift code with Objective-C compatibility

MrRooni:
## Reason to Be

From the issue:
> For a project in transition from Objective-C to Swift, there are scenarios where the generated model classes (which preferably are in Swift) have to be used by new code in legacy files (which are in Objective-C). Unless the class is appended with @objcMembers and inheriting NSObject, I believe the generated class will be unavailable for use by the Objective-C side.

Resolves https://github.com/quicktype/quicktype/issues/1148

## Thought Process

There are two main components to using Swift code in Objective-C:

1. Swift classes must inherit from NSObject.
2. Swift classes must expose classes, properties, and functions to Objective-C via `@objcMembers` or `@objc` 

While `@objcMembers` is only available in Swift 5 or greater, it was the simpler option to implement and ultimately the one I chose to use.

## How To Test

When generating Swift code you can declare `--objective-c-support` to enable this feature. It defaults to "false":

```bash
quicktype --struct-or-class class --objective-c-support -s schema schema.json -o src/ios/models.swift
```

Right now `--objective-c-support` is dependent on `--struct-or-class` being set to `class`, however there is no error reporting if you declare `--objective-c-support` without ` --struct-or-class class`. I'd be open to suggestions on how to handle this issue.

## Possible Impacts

None?

### DIFF
--- a/src/quicktype-core/Renderer.ts
+++ b/src/quicktype-core/Renderer.ts
@@ -138,6 +138,10 @@ export abstract class Renderer {
         this._emitContext.preventBlankLine();
     }
 
+    emitItem(item: Sourcelike): void {
+        this._emitContext.emitItem(item);
+    }
+
     emitLine(...lineParts: Sourcelike[]): void {
         if (lineParts.length === 1) {
             this._emitContext.emitItem(lineParts[0]);
--- a/src/quicktype-core/language/Swift.ts
+++ b/src/quicktype-core/language/Swift.ts
@@ -496,7 +496,7 @@ export class SwiftRenderer extends ConvenienceRenderer {
         this.emitLine("typealias ", name, " = ", this.swiftType(t, true));
     }
 
-    protected getProtocolsArray(_t: Type, isClass: Boolean): string[] {
+    protected getProtocolsArray(_t: Type, isClass: boolean): string[] {
         const protocols: string[] = [];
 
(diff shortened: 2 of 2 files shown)


schani:
Thank you for your PR!

It currently fails with a compile error:

```
src/quicktype-core/language/Swift.ts(811,65): error TS2554: Expected 1 arguments, but got 0.
```

MrRooni:
Cheers @schani . I'm not sure how I have it working here if it has a compile error, but I'll check it out. Thanks!

MrRooni:
I'm not sure why those tests are failing. I'll try and get the tests running locally so I can see what's up.

schani:
Stupid question, but are you sure you committed and pushed all the code?

MrRooni:
Yup, just checked. I'll spend some time on Monday getting tests up and running and see what's up. Thanks for being communicative @schani .

schani:
Now it's a TSlint issue:

```
> quicktype@15.0.0 tslint /app
  | > tslint --project src/cli --exclude 'src/__tests__/**' --exclude 'src/cli/get-stream/**'
  |  
  | no-unused-variable is deprecated. Since TypeScript 2.9. Please use the built-in compiler checks instead.
  |  
  | ERROR: /app/src/quicktype-core/language/Swift.ts:503:13 - This type is not allowed in the operand for the '&&' operator because it is always truthy. Only booleans are allowed.
```

MrRooni:
@schani Quick, merge it! 😄

schani:
Thank you!

PR #1214, opened 2019-04-26
Add support for yaml based json-schema

madmo:
since the yaml parser is yaml 1.2 compatible it is also able to parse json which is a subset of yaml 1.2


schani:
Thank you so much!

PR #1215, opened 2019-04-28
Allow custom protocol declaration in Swift

Ecarrion:
## Summary

This PR aims to allow `quicktype-core` users to add custom protocol conformance to generated swift code.

## Implementation Details

- I've added the `getProtocolsArray` method which takes care of assembling an array with all protocol declarations. This new method is marked as `protected` so consumers can override it.

- The `getProtocolString` method keeps being private and now only takes care of turning `getProtocolsArray` output into some swift valid string.

- Additionally I'm passing the `type` of the "class/union" being analyzed to allow for conditional customization.

## Usage

```typescript
class CustomRenderer extends SwiftRenderer {
    protected getProtocolsArray(t: Type): String[] {
        const protocols = super.getProtocolsArray(t);
        protocols.push("CustomProtocol");
        return protocols;
    }
}
```

### DIFF
--- a/src/quicktype-core/language/Swift.ts
+++ b/src/quicktype-core/language/Swift.ts
@@ -490,7 +490,7 @@ export class SwiftRenderer extends ConvenienceRenderer {
         this.emitLine("typealias ", name, " = ", this.swiftType(t, true));
     }
 
-    private getProtocolString(): Sourcelike {
+    protected getProtocolsArray(_t: Type): string[] {
         const protocols: string[] = [];
         if (!this._options.justTypes) {
             protocols.push("Codable");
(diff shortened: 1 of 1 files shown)


src/quicktype-core/language/Swift.ts
```diff
@@ -490,7 +490,7 @@ export class SwiftRenderer extends ConvenienceRenderer {
         this.emitLine("typealias ", name, " = ", this.swiftType(t, true));
     }
 
-    private getProtocolString(): Sourcelike {
+    protected getProtocolsArray(_t: Type): String[] {
```
schani: The return type needs to be lowercase `string[]`.

Ecarrion: Done! Didn't knew about that distinction 😅

schani:
Thank you!

PR #1218, opened 2019-05-03
Allow custom property definition in Swift

Ecarrion:
## Summary

This PR aims to allow `quicktype-core` users to add custom class property definitions to generated swift code, giving users the ability to add features such as default values.

## Implementation Details

I've extracted the property generation logic(only for non-dense mode) that was inside the `renderClassDefinition` method to expose it in a new protected method called `propertyLinesDefinition`.

## Usage

```typescript
class CustomRenderer extends SwiftRenderer {
    protected propertyLinesDefinition(name: Name, parameter: ClassProperty): Sourcelike {
        
        // Get the code for original property definition
        const originalDefinition = super.propertyLinesDefinition(name, parameter);

        // Return a modified version if required
        return [originalDefinition, " = ",  "Some Value"];
    }
}
```

### DIFF
--- a/src/quicktype-core/language/Swift.ts
+++ b/src/quicktype-core/language/Swift.ts
@@ -558,6 +558,10 @@ export class SwiftRenderer extends ConvenienceRenderer {
             : this._options.accessLevel + " ";
     }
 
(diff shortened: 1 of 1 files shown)


schani:
Great, thank you! :-)

PR #1219, opened 2019-05-04
Upgrade Elm support to 0.19

### LINKED ISSUE #1165: Is Elm 0.19 and latest Json.Decode.Pipeline supported?
I got a few compilation errors using Elm 0.19, and I'm not sure if this is just me or what Elm and `Json.Decode.Pipeline` package version is currently supported?

Would be nice to have that specifically mentioned somewhere.

If I have time I'll try and figure out what changes are required for Elm 0.19 support...

jonathanjouty:
Fixes #1165.
This will result in _only_ supporting Elm 0.19.

I have not commited my local changes to `package-lock.json`, should I?

Also, I did a quick hack around the Elm compiler's new attitude towards name shadowing, but this could cause an issue if your JSON has fields with my newly chosen variable name: `v`!

I'm sure there's a better way to solve that, but I'm not familiar with TypeScript or the `quicktype` code base, so would appreciate some help for that :)

Here's the Elm compiler error message:
```
-- SHADOWING --------------------------------------------------- ./QuickType.elm

The name `x` is first defined here:

72| x =
    ^
But then it is defined AGAIN over here:

66| encodeBar x =
              ^
Think of a more helpful name for one of them and you should be all set!

Note: Linters advise against shadowing, so Elm makes “best practices” the
default. Read <https://elm-lang.org/0.19.0/shadowing> for more details on this
choice.
```


jonathanjouty:
Also, I've left my `fixup` commits here for now, to be squashed later.

schani:
@jonathanjouty Is Elm 0.19 widely enough adopted that 0.18 support won't be missed?  I'm assuming quicktype without this PR doesn't support 0.19?

Failing with a property name `v` is not acceptable.  The proper way to fix this is to make a `SimpleName` for each class type for the variable that's currently `v`, and use that.  I think the proper place to do it is in `makePropertyDependencyNames`.  That's called for every property of every class, so you'll have to make sure that you only return the `SimpleName` for each class once.  There will also be the corner case of a class not having any properties, in which case you won't have a `SimpleName` for it.  Feel free to add to `ConvenienceRenderer` a method `makeClassDependencyNames` or something like that, to fix both problems, if you care.

icidasset:
Hey folks, would love to see this merged. I can help with the necessary changes. Most people have switched to v0.19 by now. Does your comment on the code still apply @schani ?

dvdsgl:
Moved to #1653

PR #1229, opened 2019-05-15
Fix source URLs CLI option

### LINKED ISSUE #1228: Source URLs CLI option not being observed
I like the idea of creating a JSON file with a list of schema URLs to generate. However I observed that the `--src-urls` CLI option simply doesn't work today.

On closer inspection, I can see that the option is being lost after parsing arguments.

I will have a PR open for this shortly.

benquarmby:
Resolves #1228

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -267,6 +267,7 @@ function inferCLIOptions(opts: Partial<CLIOptions>, targetLanguage: TargetLangua
     /* tslint:disable:strict-boolean-expressions */
     const options: CLIOptions = {
         src: opts.src || [],
+        srcUrls: opts.srcUrls,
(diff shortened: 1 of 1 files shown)


src/cli/index.ts
```diff
@@ -267,6 +267,7 @@ function inferCLIOptions(opts: Partial<CLIOptions>, targetLanguage: TargetLangua
     /* tslint:disable:strict-boolean-expressions */
     const options: CLIOptions = {
         src: opts.src || [],
+        srcUrls: opts.srcUrls,
```
benquarmby: Tiny change. It was being parsed correctly, but not assigned to the new `options` object.

schani:
Thank you!

PR #1231, opened 2019-05-15
Observe HTTP headers for all requests

### LINKED ISSUE #1230: Observe HTTP header option for all requests
The existing `--http-header` option is currently restricted to GraphQL introspection, but it would be useful for any HTTP request quicktype makes. For example, requesting raw JSON schema from a private GitHub repository (requiring an `Authorization: token X` header).

While it might be nice to specify exactly which requests get headers and which do not, the configuration overhead would be significant. I recommend attaching specified HTTP headers to all requests.

I will have a PR up for this shortly.

benquarmby:
Resolves #1230

### DIFF
--- a/src/cli/NodeIO.ts
+++ b/src/cli/NodeIO.ts
@@ -3,18 +3,49 @@ import { Readable } from "stream";
 
 import { getStream } from "./get-stream";
 
-import { messageError, JSONSchemaStore, JSONSchema, parseJSON } from "../quicktype-core";
+import { messageError, JSONSchemaStore, JSONSchema, parseJSON, panic } from "../quicktype-core";
 
 // The typings for this module are screwy
 const isURL = require("is-url");
 const fetch = require("node-fetch");
 
-export async function readableFromFileOrURL(fileOrURL: string): Promise<Readable> {
+interface HttpHeaders {
+    [key: string]: string;
+}
+
+function parseHeaders(httpHeaders?: string[]): HttpHeaders {
+    if (!Array.isArray(httpHeaders)) {
+        return {};
+    }
+
+    return httpHeaders.reduce(
+        function(result: HttpHeaders, httpHeader: string) {
+            if (httpHeader !== undefined && httpHeader.length > 0) {
+                const split = httpHeader.indexOf(":");
+
+                if (split < 0) {
+                    return panic(`Could not parse HTTP header "${httpHeader}".`);
+                }
+
+                const key = httpHeader.slice(0, split).trim();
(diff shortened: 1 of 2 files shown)


src/quicktype-core/input/JSONSchemaStore.ts
```diff
@@ -3,6 +3,8 @@ import { StringMap, assert } from "../support/Support";
 export type JSONSchema = StringMap | boolean;
 
 export abstract class JSONSchemaStore {
+    constructor(protected headers?: string[]) {}
```
benquarmby: I could see moving this to the `FetchingJSONSchemaStore` constructor. Might be better there.

benquarmby: Done.

schani:
This is failing TSLint.  The buildkite pipeline is now public - you should be able to see the errors.

benquarmby:
Thanks, will take a look.

src/cli/NodeIO.ts
```diff
@@ -9,12 +9,38 @@ import { messageError, JSONSchemaStore, JSONSchema, parseJSON } from "../quickty
...
+
+function parseHeaders(httpHeaders?: string[]): HttpHeaders {
+    if (!Array.isArray(httpHeaders)) {
+        return {};
+    }
+
+    return httpHeaders.reduce(
+        function (result: HttpHeaders, httpHeader: string) {
+            if (httpHeader !== undefined && httpHeader.length > 0) {
+                const [key, value] = httpHeader.split(":");
```
schani: This will not work if the input string is, say `Header-With-Colons: Foo:Bar:Baz`.  Maybe that's not allowed?

benquarmby: Yes you're right, this is a bit naive. I'll take another crack at making it more robust.

benquarmby: Updated, let me know what you think @schani.

schani: Looks good!

schani:
Sorry, we're having some build server issues at the moment, which might or might not be our fault.

schani:
Thank you!

benquarmby:
🎉 

I'm using this feature from my local build right now. What is the cadence for rolling out updates to NPM?

schani:
The cadence is it was published to NPM 50 minutes ago ;-)

https://www.npmjs.com/package/quicktype

benquarmby:
Booyah! Thank you very much.

PR #1233, opened 2019-05-16
Adds ExplicitCodingkey option to SwiftOptions

Ecarrion:
# Summary

When generating swift code without the `--just-types` option from the schema:
```json
"recipe.visit":{
  "$id":"#/events/recipe.visit",
  "type":"object",
  "description":"When a user successfully views / visits a recipe.",
  "properties":{
    "event":{
      "type":"string",
      "const":"recipe.visit"
    },
    "event_time":{
      "type":"string"
    },
  },
  "required":[
    "event",
    "event_time"
  ]
}
```
The output looks like this:
```swift
public struct RecipeVisit: Codable {
    public let event: String
    public let eventTime: String

        case event = "event"
        case eventTime = "event_time"
    }
}
```

Most of the time this is correct and works fine in swift expect when we specify the `KeyDecodingStrategy.convertFromSnakeCase` in swift's `JSONDecoder` class, which converts the incoming json to `camelCase` and then swift can't match the camel-cassed json to the generated snake-cased `CodingKeys` enum.

This PR aims to solve ☝️by providing the user a way to opt-out from explicit `CodingKeys` values.

# Usage

By running `quicktype` with the `--no-coding-keys` option
```
quicktype -s schema "schema.json" -o out.swift  --no-coding-keys
```

The swift generated code will look like
```swift
public struct RecipeVisit: Codable {
    public let event: String
    public let eventTime: String

        case event
        case eventTime
    }
}
```

Which will work fine with the `KeyDecodingStrategy.convertFromSnakeCase` swift's `JSONDecoder` option

### DIFF
--- a/src/quicktype-core/language/Swift.ts
+++ b/src/quicktype-core/language/Swift.ts
@@ -47,6 +47,7 @@ const MAX_SAMELINE_PROPERTIES = 4;
 export const swiftOptions = {
     justTypes: new BooleanOption("just-types", "Plain types only", false),
     convenienceInitializers: new BooleanOption("initializers", "Generate initializers and mutators", true),
+    explicitCodingKeys: new BooleanOption("coding-keys", "Explicit CodingKey values in Codable types", true),
     urlSession: new BooleanOption("url-session", "URLSession task extensions", false),
     alamofire: new BooleanOption("alamofire", "Alamofire extensions", false),
     namedTypePrefix: new StringOption("type-prefix", "Prefix for type names", "PREFIX", "", "secondary"),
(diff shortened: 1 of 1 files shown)


schani:
I don't understand what you mean by

> swift can't match the camel-cassed json to the generated snake-cased CodingKeys enum

I thought the whole point of the `CodingKeys` enum is to make it so that Swift can match it?

Ecarrion:
> I thought the whole point of the CodingKeys enum is to make it so that Swift can match it?

Yes, in a vanilla environment, that `CodingKey` enum will bridge the snake-cased json to the camel-cased swift file.

But in most projects(like mine 😅) where most of your models aren't auto-generated  you can use the `KeyDecodingStrategy.convertFromSnakeCase` to avoid writing those `CodingKeys` enums manually.
That thing basically turns the snaked-cased input json into a camel_cased json. So if you leave the auto-gen `CodingKeys` enum. Swift will try to find `event_time ` in a json where the field has been converted to `eventTime`.

Makes sense?

schani:
Thanks!

PR #1236, opened 2019-05-17
Support generating converters for all objects

### LINKED ISSUE #1235: Create converters for all types, not just top level
Unless `--just-types` is specified, a handy `Convert` utility is created to cast raw JSON to well typed data models. However, these are only created for the top level types.

Support creating conversion routines for all types described in a schema.

This will support the common scenario where REST APIs return nested resources. As a fabricated example:
* The JSON schema `car.json` describes the top level `Car` type and a nested `CarManufacturer` type.
* The API provides a top level `car/:carId` resource. This maps to the quicktype `Convert.toCar` helper ✅ 
* The API provides a nested `car/:carId/manufacturer` resource. While quicktype created the type for us, there is no matching converter ❌

benquarmby:
Resolves #1235

### DIFF
--- a/src/quicktype-core/language/JavaScript.ts
+++ b/src/quicktype-core/language/JavaScript.ts
@@ -10,6 +10,8 @@ import {
 } from "../Type";
 import { matchType, directlyReachableSingleNamedType } from "../TypeUtils";
 import { acronymOption, acronymStyle, AcronymStyleOptions } from "../support/Acronyms";
+import { convertersOption, ConvertersOptions } from "../support/Converters";
+
 import {
     utf16LegalizeCharacters,
     utf16StringEscape,
@@ -33,7 +35,8 @@ import { isES3IdentifierPart, isES3IdentifierStart } from "./JavaScriptUnicodeMa
 
 export const javaScriptOptions = {
     acronymStyle: acronymOption(AcronymStyleOptions.Pascal),
-    runtimeTypecheck: new BooleanOption("runtime-typecheck", "Verify JSON.parse results at runtime", true)
+    runtimeTypecheck: new BooleanOption("runtime-typecheck", "Verify JSON.parse results at runtime", true),
+    converters: convertersOption()
 };
 
 export type JavaScriptTypeAnnotations = {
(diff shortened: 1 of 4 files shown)


src/quicktype-core/language/JavaScript.ts
```diff
@@ -273,7 +276,17 @@ export class JavaScriptRenderer extends ConvenienceRenderer {
                     this.emitLine("return JSON.stringify(uncast(value, ", typeMap, "), null, 2);");
                 }
             });
-        });
+        };
+
+        switch (this._jsOptions.converters) {
+            case ConvertersOptions.All:
+                this.forEachObject("interposing", converter);
```
benquarmby: #naming-is-hard

Is `All` the right name for this? My assumption is this wont include enums?

src/quicktype-core/support/Converters.ts
```diff
@@ -0,0 +1,16 @@
+import { EnumOption } from "../RendererOptions";
+
+export enum ConvertersOptions {
+    Top = "top",
```
schani: I'd call this `TopLevel = "top-level"`.

schani:
Yes, this can ship with just JS/TS/Flow.

I left a comment on naming.

Right now the only easy test you can have is by adding the new option to the `quickTestRendererOptions`.  When the quicktype tests run, all input files are tested with default options, but then a small subset of input files is also tested with all of the `quickTestRendererOptions`.  In this case none of the additional generated code would run, but it would at least have to compile.  To test more than that you'd have to write custom tests, for which we don't have a nice framework yet.

schani:
There are [some differences in the output](https://s3.amazonaws.com/quicktype-outputs/diffs/53fc1c6bbba6ce2b38bd404278e2fae8b53901c8-7e40e8fc2ffb6afe8884a6bb4e512284205005ce.diff) of GraphQL tests.  Any idea why they happen?

benquarmby:
Will rename the `TopLevel` option and try out the `quickTestRendererOptions`.

Can't explain that diff. Will comb over the code changes again.

src/quicktype-core/language/JavaScript.ts
benquarmby: Renamed to `AllObjects`.

src/quicktype-core/support/Converters.ts
```diff
@@ -0,0 +1,19 @@
+import { EnumOption } from "../RendererOptions";
+
+export enum ConvertersOptions {
+    TopLevel = "top-level",
```
benquarmby: Renamed to `TopLevel`.

schani:
Fantastic, thank you!

PR #1237, opened 2019-05-19
Multi-file Golang Output

MrRooni:
## Reason to Be

When running quicktype to generate Go code from a directory of schema files the output currently dumps everything into a single `.go` file. This PR breaks each named type out into its own file. 

## CHANGELOG Entry

> [IMPROVED] Generating Go files from a directory of inputs can now output one Go file per schema object.

## Thought Process

After having been through the gauntlet with the Swift renderer, this one came together pretty fast.

## How To Test

```
quicktype . -l Golang -s schema -o .
```

## Possible Impacts

None.

### DIFF
--- a/src/quicktype-core/Renderer.ts
+++ b/src/quicktype-core/Renderer.ts
@@ -149,21 +149,26 @@ export abstract class Renderer {
         this._emitContext.emitItem(item);
     }
 
-    emitItemOnce(item: Sourcelike): void {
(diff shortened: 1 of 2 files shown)


schani:
There are some weird extra newlines in the output.  Make sure you use `ensureBlankLine()`, vs emitting newlines.

MrRooni:
@schani Sorted, nice catch.

schani:
Thank you!

PR #1238, opened 2019-05-20
Multi-file Rust Output

MrRooni:
## Reason to Be

When running quicktype to generate Rust code from a directory of schema files the output currently dumps everything into a single .rs file. This PR breaks each named type out into its own file.

## CHANGELOG Entry

[IMPROVED] Generating Rust files from a directory of inputs can now output one Rust file per schema object.

## Thought Process

After having been through the gauntlet with the Swift renderer and then the Go renderer, this one came together pretty fast.

## How To Test

`quicktype . -l Rust -s schema -o .`

## Possible Impacts

None.


schani:
This seems to add a lot of comments for every emitted type.

MrRooni:
After learning a bit more about how Rust handles files, multi-file output isn't really a good fit for this language right now.

PR #1251, opened 2019-06-21
[WIP] Make Python from_dict arg type more specific

gricey432:
Functions called "from_dict" and "to_dict" marked as having arg type `Any` rather than `dict`.
These eventually get down to each class's `from_dict` which immediately does
```
assert isinstance(obj, dict)
```


gricey432:
Making a newer one with more changes

PR #1254, opened 2019-06-24
Python improvements

gricey432:
Made a bunch of changes to the Python renderer. Replaces #1251 

- Add some more reserved words to the forbidden list (possibly breaking for people who have used them)
- Make all classes come off `object`
- Removed a lot of `Any`
- Removed all `assert` since they [might not do anything at runtime](https://stackoverflow.com/a/5143044/950099). I'm not sure if that was intended, but since this is for parsing user input I figured it was desired to always throw exceptions on bad data instead of continuing and failing elsewhere

### DIFF
--- a/src/quicktype-core/language/Python.ts
+++ b/src/quicktype-core/language/Python.ts
@@ -75,6 +75,7 @@ const forbiddenPropertyNames = [
     "for",
     "from",
     "global",
+    "id",
     "if",
     "import",
     "in",
@@ -83,6 +84,7 @@ const forbiddenPropertyNames = [
     "lambda",
     "nonlocal",
     "not",
+    "open",
     "or",
     "pass",
     "print",
@@ -356,7 +358,7 @@ export class PythonRenderer extends ConvenienceRenderer {
 
     protected declarationLine(t: Type): Sourcelike {
         if (t instanceof ClassType) {
-            return ["class ", this.nameForNamedType(t), ":"];
+            return ["class ", this.nameForNamedType(t), "(object):"];
         }
         if (t instanceof EnumType) {
             return ["class ", this.nameForNamedType(t), "(", this.withImport("enum", "Enum"), "):"];
@@ -609,7 +611,7 @@ export class JSONPythonRenderer extends PythonRenderer {
         snakeNameStyle(this.pyOptions.features.version, s, false)
     );
     private readonly _topLevelConverterNames = new Map<Name, TopLevelConverterNames>();
-    private _haveTypeVar = false;
+    private _typeVars = new Set<string>();
     private _haveEnumTypeVar = false;
     private _haveDateutil = false;
 
@@ -620,8 +622,8 @@ export class JSONPythonRenderer extends PythonRenderer {
         this.emitLine(tvar, " = ", this.withTyping("TypeVar"), "(", this.string(tvar), constraints, ")");
     }
 
-    protected typeVar(): string {
-        this._haveTypeVar = true;
+    protected typeVar(name: string = "T"): string {
+        this._typeVars.add(name);
         // FIXME: This is ugly, but the code that requires the type variables, in
         // `emitImports` actually runs after imports have been imported.  The proper
         // solution would be to either allow more complex dependencies, or to
@@ -629,7 +631,7 @@ export class JSONPythonRenderer extends PythonRenderer {
         // gather-emit is a bit buggy with blank lines, and I can't be bothered to
         // fix it now.
         this.withTyping("TypeVar");
-        return "T";
+        return name;
     }
 
     protected enumTypeVar(): string {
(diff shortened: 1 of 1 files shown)


henribru:
Does it make sense to type the `from_T` methods as taking T as input instead of Any? They are after all intended to take any possible type as input and fail if the type is wrong. The function doesn't actually do anything when passed T.

It might be better to only inherit from `object` in Python 2.7, since this is a no-op in Python 3.

dvdsgl:
Great!

Here's the error message from the build failure:
```
* CLI sanity check
--
  | * Running 1263 tests between 9 fixtures
  | * Setting up json-ts-csharp fixture
  | * Forking 8 workers 👷🏾 👷🏽 👷🏼 👷🏾 👷🏼 👷🏻 👷🏾 👷🏿
  | * [5/1263] python(python-version: 2.7) test/runs/python-41dfd6/combinations1.json: 2538 ms
  | * [1/1263] python(python-version: 3.5) test/runs/python-da6a3f/combinations1.json: 2659 ms
  | * [2/1263] python(python-version: 3.5) test/runs/python-b044d2/combinations2.json: 2662 ms
  | * [6/1263] python(python-version: 2.7) test/runs/python-832b0b/combinations2.json: 2665 ms
  | * [7/1263] python(python-version: 2.7) test/runs/python-fb8e81/combinations3.json: 2725 ms
  | * [3/1263] python(python-version: 3.5) test/runs/python-abb9c2/combinations3.json: 2768 ms
  | * [8/1263] python(python-version: 2.7) test/runs/python-3c7a8d/combinations4.json: 2826 ms
  | * [4/1263] python(python-version: 3.5) test/runs/python-280e89/combinations4.json: 2852 ms
  | quicktype.py:70: error: Argument 2 to "from_list" has incompatible type "Optional[Any]"; expected "List[Any]"
  | quicktype.py:71: error: Argument 1 to "from_bool" has incompatible type "Optional[Any]"; expected "bool"
  | quicktype.py:72: error: Argument 1 to "from_int" has incompatible type "Optional[Any]"; expected "int"
  | quicktype.py:73: error: Argument 1 to "from_str" has incompatible type "Optional[Any]"; expected "str"
  | quicktype.py:74: error: Argument 1 to "from_int" has incompatible type "Optional[Any]"; expected "int"
  | quicktype.py:75: error: Argument 1 to "from_int" has incompatible type "Optional[Any]"; expected "int"
  | quicktype.py:76: error: Argument 1 to "from_int" has incompatible type "Optional[Any]"; expected "int"
  | quicktype.py:77: error: Argument 1 to "from_int" has incompatible type "Optional[Any]"; expected "int"
  |  
  |  
  | Fixture threw an exception
  | {
  | "error": {
  | "command": "mypy quicktype.py",
  | "code": 1
  | },
  | "sample": {
  | "path": "test/inputs/json/priority/bug790.json",
  | "saveOutput": true
  | },
  | "cwd": "/app/test/runs/python-d4d17e"
  | }
  | [object Object]
  | at /app/test/test.ts:68:17
  | at step (/app/test/test.ts:32:23)
  | at Object.throw (/app/test/test.ts:13:53)
  | at rejected (/app/test/test.ts:5:65)
  | at <anonymous>
  | at process._tickDomainCallback (internal/process/next_tick.js:229:7)
  | npm ERR! Test failed.  See above for more details.
```

gricey432:
> Does it make sense to type the `from_T` methods as taking T as input instead of Any? They are after all intended to take any possible type as input and fail if the type is wrong. Some of them don't even do anything when passed T.

Yeah I thought about this, I think it makes sense. I'd argue that a function which throws an exception on anything except `from_T` should be marked as only taking `from_T`. I think the type hints should be used to hint the developer that any other argument type is probably going to cause problems.

> It might be better to only inherit from `object` in Python 2.7, since this is a no-op in Python 3.

I initially did this, but the reduced readability of the TS didn't seem worth it to avoid the no-op, especially since it's still perfectly valid.

> Thanks for replacing the assignments in `to_dict` with a literal, this has irked me!

:+1: I got very tired of Pycharm highlighting these yellow

RE The failing tests:
I tried to get them to run locally with the docker script and it failed while building the container. Didn't have time to debug the docker issue so I just left it. I'll have to come back to it at some point.

henribru:
> Yeah I thought about this, I think it makes sense. I'd argue that a function which throws an exception on anything except `from_T` should be marked as only taking `from_T`. I think the type hints should be used to hint the developer that any other argument type is probably going to cause problems.

Your reasoning makes sense, but the problem is that it doesn't match how the generated code actually uses these functions and this will cause complaints from a type checker. As an example, say Quicktype generates this:
```py
@dataclass
class Klass:
    attribute: int

    @staticmethod
    def from_dict(obj: dict) -> 'Klass':
        if not isinstance(obj, dict):
			raise TypeError()
		attribute = from_int(obj.get("attribute"))
		return Klass(attribute)
```
`obj.get("attribute")` has the type `Optional[Any]`, but is being passed to a function that expects an integer. This is an error. In fact, it's exactly the same error that the tests are showing!

gricey432:
Yep, I understand what you mean.

I guess the most correct fix would be to replace `from_dict(obj: dict)` with `from_dict(obj: KlassDict)` which is a `TypedDict` generated alongside `Klass` but that's a whole new endeavour. I might do that if you agree?

henribru:
That's probably the best solution to the issue, but `TypedDict` won't be in the standard library until Python 3.8. Adding a dependency on typing-extensions should probably be avoided.

gricey432:
Ah I'm so used to just having it installed on everything I work on I've forgotten it's not std :(

Ok looks like the best we can do is to leave the `from_T` the way they were

dvdsgl:
Sorry you’re having trouble running the tests. Can you run just the Python tests locally without Docker?

PR #1260, opened 2019-07-02
fix: GraphQL Introspection Query

benmccallum:
Out-of-the-box, schema introspection wasn't working with my GraphQL server as the default HTTP method was GET, but the request was using a body with the serialized content. This doesn't align with the recommended serving (in its various forms) of a graphql api over HTTP as per: https://graphql.org/learn/serving-over-http/

I was tempted to just make this `"POST"` and not wrap in `withDefault`, but I guess some people's graphql APIs might work on a GET with a body... albeit strange.


schani:
Thank you!

PR #1275, opened 2019-07-25
[Swift] Add more dictionary support

MrRooni:
## Reason to Be

At 1Password we have run into a few instances where we have generic dictionary objects that we want to initialize into strongly typed objects. This PR adds the ability to do that. It also allows us to dump a strongly typed object out to a dictionary. 

## CHANGELOG Entry

> [IMPROVED] Swift classes can now be initialized using a generic dictionary (`[AnyHashable:Any]`).
> [IMPROVED] You can now get a dictionary representation of a Swift class.

## Thought Process

This one was pretty straightforward. The code to do this is the same for every class, so I just added a new initializer and function.

## How To Test

```
quicktype . -l Swift -s schema -o .
```

## Possible Impacts

None.

### DIFF
--- a/src/quicktype-core/language/Swift.ts
+++ b/src/quicktype-core/language/Swift.ts
@@ -867,12 +867,12 @@ encoder.dateEncodingStrategy = .formatted(formatter)`);
             this.ensureBlankLine();
 
             this.emitBlock([convenience, "init(dictionary: [AnyHashable:Any]) throws"], () => {
-                this.emitMultiline(`    guard JSONSerialization.isValidJSONObject(dictionary) else {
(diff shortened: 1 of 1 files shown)


MrRooni:
@schani Looks like just the diff failed (as expected). Otherwise I think we’re good to go on this one.

MrRooni:
Discussing this more internally with the 1Password team, let's throw a `wait` tag on this one for now.

dvdsgl:
@MrRooni want me to close for now?

MrRooni:
@dvdsgl Indeed. Will reopen as needed. Thanks.

PR #1283, opened 2019-08-01
Python optional properties gets None default value

### LINKED ISSUE #1216: python 3.7: Optional w/o default, description placement
Hi!

I've got an API that can describe itself with something that looks like JSON Schema. I write "looks like", because I'm not sure if it does conform with everything.
(A quick romp through the schema spec does not state if a non-required field must provide a default.)

In my case, the result is a generated python dataclass with something like this (v15.0.174):

```python
@dataclass
class BackupCollection:
    """Backup Solr indexes and configurations for a specific collection. One copy of the indexes
    will be taken from each shard, and the config set for the collection will also be copied.
    """
    """Defines a request ID that can be used to track this action after it's submitted. The
    action will be processed asynchronously.
    """
    backup_collection_async: Optional[str]
    """The name of the collection to back up."""
    collection: str
    """A location on a shared drive for the backup-collection command to write to. Alternately,
    it can be set as a cluster property with the cluster endpoint, which also supports
    setting a location.
    """
    location: str
    """The name of the backup."""
    name: str
```

I have to gripe about some of that:

1. Note the missing default on `backup_collection_async`. 
The spec did not include a default, so there is none in the dataclass.
In this case, it's fine as it's just one parameter, but I have more complex classes that take 1 required and 15 optional parameters.
w/o a default value, this get's very tedious to use, when you don't want to use the `from_dict`.

Proposed solution: for fields that have no default specified in the schema, we use `[name]: Optional[T] = None`.

2. the docstrings are placed above the fields.
This is quite unusual in python. At least for me, since I'm usually expecting docstrings below the declaration.

Proposed solution: we put field descriptions in the top level docstring, i.e.

```python
@dataclass
class CreateAlias:
    """
    Create an Alias

    :param create_alias_async: Defines a request ID that can be used to track this action after it's submitted. The action will be processed asynchronously.
    :param collections: The list of collections to be known as this alias. Incompatible with any of the routing parameters. Either this parameter or a complete set of routing parameters is required.
    :param create_collection: The settings to use to create a collection for each new time partition. Most options from the collection create command are available, except for 'name', 'async' and 'waitForFinalState'.
    :param name: The alias name to be created.
    :param router: Routing specific properties to define a time routed alias.  Do not specify 'collections' when creating a time routed alias.
    :param tz: Optional timezone for use with any date math that may exist in other parameters. Defaults to UTC.
    """
    create_alias_async: Optional[str]
    collections: Optional[List[str]]
    create_collection: Optional[Dict[str, Any]]
    name: str
    router: Optional[CreateAliasRouter]
    tz: Optional[str]
```

That way, `help(CreateAlias)` can provide more insight when interacting on the REPL.

Comments and other opinions welcome :3

remorses:
When generating python types, optional values gets type of `Optional[T]` but you still needs to give these values during initialization:
```python
@dataclass
class Object:
    a: float
    b: Optional[str]
```
to initialize Object you need to call `Object(a=2, b=None)`
```python
@dataclass
class Object:
    a: float
    b: Optional[str] = None
```
After this is merged you can call `Object(a=2,)`
Closes #1216

### DIFF
--- a/src/quicktype-core/ConvenienceRenderer.ts
+++ b/src/quicktype-core/ConvenienceRenderer.ts
@@ -644,7 +644,7 @@ export abstract class ConvenienceRenderer extends Renderer {
         this._alphabetizeProperties = value;
     }
 
-    getAlphabetizeProperties(): boolean {
+    protected getAlphabetizeProperties(): boolean {
         return this._alphabetizeProperties;
     }
 
--- a/src/quicktype-core/ConvenienceRenderer.ts
+++ b/src/quicktype-core/ConvenienceRenderer.ts
@@ -654,11 +654,14 @@ export abstract class ConvenienceRenderer extends Renderer {
         return propertyNames.size;
     }
 
-    protected sortClassProperties(_p: ClassProperty, jsonName: string): number | string {
+    protected sortClassProperties(properties: ReadonlyMap<string, ClassProperty>): ReadonlyMap<string, ClassProperty> {
         if (this._alphabetizeProperties) {
-            return jsonName;
+            return mapSortBy(properties, (_p: ClassProperty, jsonName: string) => {
+                return jsonName;
+            });
+        } else {
+            return properties;
         }
-        return 0;
     }
 
     protected forEachClassProperty(
@@ -667,7 +670,7 @@ export abstract class ConvenienceRenderer extends Renderer {
         f: (name: Name, jsonName: string, p: ClassProperty, position: ForEachPosition) => void
     ): void {
         const propertyNames = defined(this._propertyNamesStoreView).get(o);
-        const sortedProperties = mapSortBy(o.getProperties(), this.sortClassProperties.bind(this));
+        const sortedProperties = this.sortClassProperties(o.getProperties());
         this.forEachWithBlankLines(sortedProperties, blankLocations, (p, jsonName, pos) => {
             const name = defined(propertyNames.get(jsonName));
             f(name, jsonName, p, pos);
--- a/src/quicktype-core/ConvenienceRenderer.ts
+++ b/src/quicktype-core/ConvenienceRenderer.ts
@@ -644,6 +644,10 @@ export abstract class ConvenienceRenderer extends Renderer {
         this._alphabetizeProperties = value;
     }
 
+    getAlphabetizeProperties(): boolean {
+        return this._alphabetizeProperties;
+    }
+
     // Returns the number of properties defined for the specified object type.
     protected propertyCount(o: ObjectType): number {
         const propertyNames = defined(this._propertyNamesStoreView).get(o);
--- a/src/quicktype-core/ConvenienceRenderer.ts
+++ b/src/quicktype-core/ConvenienceRenderer.ts
@@ -654,10 +654,11 @@ export abstract class ConvenienceRenderer extends Renderer {
         return propertyNames.size;
     }
 
-    protected sortClassProperties(properties: ReadonlyMap<string, ClassProperty>): ReadonlyMap<string, ClassProperty> {
+    protected sortClassProperties(properties: ReadonlyMap<string, ClassProperty>, propertyNames: ReadonlyMap<string, Name>): ReadonlyMap<string, ClassProperty> {
         if (this._alphabetizeProperties) {
             return mapSortBy(properties, (_p: ClassProperty, jsonName: string) => {
-                return jsonName;
+                const name = defined(propertyNames.get(jsonName));
+                return defined(this.names.get(name));
             });
         } else {
             return properties;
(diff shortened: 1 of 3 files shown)


src/quicktype-core/Type.ts
```diff
@@ -479,8 +479,8 @@ export class ObjectType extends Type {
         }
     }
 
-    setProperties(properties: ReadonlyMap<string, ClassProperty>, additionalPropertiesRef: TypeRef | undefined) {
-        assert(this._properties === undefined, "Tried to set object properties twice");
+    setProperties(properties: ReadonlyMap<string, ClassProperty>, additionalPropertiesRef: TypeRef | undefined, assertFirstTime = true) {
```
remorses: I did this to be able to change properties order

src/quicktype-core/Type.ts
```diff
@@ -479,8 +479,8 @@ export class ObjectType extends Type {
...
+      if (assertFirstTime) assert(this._properties === undefined, "Tried to set object properties twice");
```
remorses: I did this to be able to change properties order

remorses: In python default properties must be last

src/quicktype-core/language/Python.ts
```diff
@@ -417,6 +419,10 @@ export class PythonRenderer extends ConvenienceRenderer {
                 if (t.getProperties().size === 0) {
                     this.emitLine("pass");
                 } else {
+                    const sort = (v: ClassProperty) => (
+                        Boolean(v.type instanceof UnionType && Boolean(nullableFromUnion(v.type))) ? 1 : 0
```
dvdsgl: Who do you need this outer `Boolean`?

src/quicktype-core/Type.ts
dvdsgl: @schani will have to approve

src/quicktype-core/language/Python.ts
remorses: tslint warnings for `null is not boolean` or something like that

remorses:
The diff test fails because the properties order has changed, but this is required to make default props last

src/quicktype-core/language/Python.ts
```diff
@@ -417,6 +419,10 @@ export class PythonRenderer extends ConvenienceRenderer {
...
+                    );
+                    t.setProperties(mapSortBy(t.getProperties(), sort), undefined, false);
```
schani: Again, not the way to do this.  A much nicer way and more general way to do it would be to generalize `ConvenienceRenderer.forEachClassProperty` by adding a `protected` method `sortClassProperties`, which, in `ConvenienceRenderer` would either just return the properties in their natural order or, if `alphabetizeProperties` is set, in alphabetical order.  Then `PythonRenderer` could overwrite it to use its own ordering.

remorses: implemented in 762d2fc

src/quicktype-core/ConvenienceRenderer.ts
```diff
@@ -644,30 +644,34 @@ export abstract class ConvenienceRenderer extends Renderer {
...
+        return this._alphabetizeProperties;
+    }
+
     // Returns the number of properties defined for the specified object type.
     protected propertyCount(o: ObjectType): number {
         const propertyNames = defined(this._propertyNamesStoreView).get(o);
         return propertyNames.size;
     }
 
+    protected sortClassProperties(_p: ClassProperty, jsonName: string): number | string {
```
schani: Please make this function actually sort the properties, and not just return a key.  First its name implies that it sorts, second `Array.sort` is not guaranteed to be stable in JS, so returning `0` doesn't necessarily mean the properties will be left in their original order.

src/quicktype-core/language/Python.ts
```diff
@@ -408,6 +410,16 @@ export class PythonRenderer extends ConvenienceRenderer {
         return this.typeHint(" -> ", this.withTyping(type));
     }
 
+    protected sortClassProperties(p: ClassProperty, jsonName: string): number | string {
+        if (this.getAlphabetizeProperties()) {
+            return jsonName;
+        }
+        if (this.pyOptions.features.dataClasses) {
+            return p.type instanceof UnionType && Boolean(nullableFromUnion(p.type)) ? 1 : 0;
```
schani: `nullableFromUnion(p.type) !== null` instead of `Boolean(nullableFromUnion(p.type))`, please.

src/quicktype-core/ConvenienceRenderer.ts
remorses: done

src/quicktype-core/language/Python.ts
```diff
@@ -408,6 +410,20 @@ export class PythonRenderer extends ConvenienceRenderer {
         return this.typeHint(" -> ", this.withTyping(type));
     }
 
+    protected sortClassProperties(properties: ReadonlyMap<string, ClassProperty>): ReadonlyMap<string, ClassProperty> {
+        if (this.getAlphabetizeProperties()) {
+            return mapSortBy(properties, (_p: ClassProperty, jsonName: string) => {
+                return jsonName;
+            });
```
schani: Sorry I'm being super pedantic here, but this method should only handle the new cast.  In this case, and the third (`return properties`) it should just do `return super.sortClassProperties(properties)`

schani:
[Kotlin has a diff](https://s3.amazonaws.com/quicktype-outputs/diffs/1b2137a568c5c25d1bd0a563b5a9fe9a4a4b46b9-6b530a5b3b96b60fed182f66f8127ff7934d7f0c.diff)

remorses:
I think a peer dependency for kotlin changed from last release, so outputs are different.
At the top of the diff there is this:
```diff
-// To parse the JSON, install Klaxon and do:
+// To parse the JSON, install jackson-module-kotlin and do:
```
My code isn’t the cause.

schani:
I see.  Sorry about that.  Thank you so much for your work!

PR #1297, opened 2019-08-21
WIP: feature/go-methods

matiasanaya:
I'm building a server that interacts with shopify through their GraphQL API. I unmarshal shopify's response with `quicktype` generated code.

I have at least five different queries that return an `Order` and would like to be able to reuse my code, so the best I could come up with was to put an `interface`.

`quicktype` by default generates structs with fields but no methods, so I went about hacking the methods onto the generation step.

This implementation is no more than a POC but I'd love to see this as an optional feature eventually.

Would you consider merging a proper implementation of this behaviour?

Is there a better way to solve for this?

cheers!

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@matiasanaya/quicktype-go-methods",
-    "version": "17.0.1",
+    "version": "17.0.2",
     "license": "Apache-2.0",
     "main": "dist/cli/index.js",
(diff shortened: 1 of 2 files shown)


schani:
Could you paste some of the generated code, please, and explain what problem it solves?

dvdsgl:
Please reopen with a reply to @schani’s comment if you want to move this forward.

PR #1303, opened 2019-09-03
FAQ: Removed extraneous period that broke a link

adriandole:


dvdsgl:
This PR is pointless! ;)

Issue #1306, opened 2019-09-05
[feature request] final fields for Dart


brooth:
Make class fields final or add an option to do so.
Thanks!

MichaelPriebe:
Yes please!

TarekkMA:
#1324  should solve this issue

brooth:
I can see that it's already in production. Closing this

Issue #1304, opened 2019-09-03
Dart double and int troubles


MichaelPriebe:
If the json has numbers that can appear as both, for example, `1` or `1.5` the code will error trying to assign the wrong type to the int or double.

brooth:
Use this template to treat with doubles in dart:
```dart
(json['value'] as num).toDouble()
```

dvdsgl:
I'm sorry, can you provide an example?

MichaelPriebe:
If you give an sample that has `1.5`, but then actually try to convert data that has `1` it will error because int is not a double.
If you give an sample that has `1`, but then actually try to convert data that has `1.5` it will error because double is not an int.
You can use the above solution by @brooth to prevent these errors.

MichaelPriebe:
@dvdsgl Please add the option for num. I want to use quicktype over json_serializable. But I get way too many errors when I don't know if the server will always return a int or a double. And It takes way too much time to go back and fix the data.

MichaelPriebe:
@dvdsgl Super needing to use `num` instead of int or double. https://github.com/flutterchina/json_model uses num.

PR #1324, opened 2019-10-17
[Dart] add required/final props

TarekkMA:
Added 2 new options for Dart language.
* Required constractor parameters (adds @required annotation in the constructor) and importing ```package:meta/meta.dart```
*  Final properties. (Solving #1306)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "quicktype",
-  "version": "15.0.0",
+  "version": "15.0.203",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "quicktype",
-    "version": "15.0.0",
+    "version": "15.0.203",
     "license": "Apache-2.0",
     "main": "dist/cli/index.js",
     "types": "dist/cli/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "quicktype",
-    "version": "15.0.203",
+    "version": "15.0.0",
     "license": "Apache-2.0",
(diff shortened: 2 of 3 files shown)


dvdsgl:
Great! Can you please drop the changes to package*.json? Thank you.

TarekkMA:
@dvdsgl Oh sorry about that. I have now removed the version change.

dvdsgl:
Thanks!

MichaelPriebe:
Thank you! May you look at the funky int/double situation. An option to use num to make sure you won't have any errors would be helpful.

TarekkMA:
@MichaelPriebe do you mean there is an issue about this PR? Can you give me more info?

MichaelPriebe:
@TarekkMA No, not about this PR, but I get problems if my samples show numbers as int's but when I'm parsing real data I get errors when a double appears in its place. My issue (https://github.com/quicktype/quicktype/issues/1304) was closed because it is a "feature" to be strict, But it takes too much time to go back to my sample and adjust it to show a double. It would be nice to cover all bases with num instead. I figured I would ask you here since you were in the process of making the Dart output better!

TarekkMA:
You should input data with decimal points if you want to have them represented as double. Quicktype can't know if ```{ "d": 1 }```  d should be double or int.  so it assigns it with int type since it has nothing that points to it being double.

Maybe there should be an option to assume all numeric values as a floating-point number, but it's outside the scope of the dart. it should be in the ```quicktype core```

MichaelPriebe:
If there was an option for everything to be [num](https://api.dartlang.org/stable/2.5.2/dart-core/num-class.html) it would cover all bases, but still retain the ability to turn it into a int, or a double.

dvdsgl:
@MichaelPriebe this is off-topic and there is already an issue for  it. Please stop.

PR #1335, opened 2019-11-09
Create .gitattributes

zippy1981:
Trying to run this from my windows laptop and this is a first step. Obviously its going to need WSL or docker to happen, but at the very least I need the bash files changes with the proper line endings.


dvdsgl:
Thanks!

PR #1339, opened 2019-11-13
"Any" types get converted to JsonObject and JsonArray in KotlinX

asarazan:
Summary:
Kotlinx is able to handle some pretty open-ended types by just dumping them into the JsonObject and JsonArray classes. You don't get fully static types, but it at least allows you to operate on those types without having to hand-fix the code.

Details:
I had to break out various parts of KotlinRenderer's `kotlinType` function, so that subclasses can override how to handle any-like types. I've updated the renderer to map the following:
`Any` -> `JsonObject`
`List<Any>` -> `JsonArray`
`Map<*, Any>` -> `JsonObject`

Tests:
Have verified it works for reasonable inputs on my machine. Have reached out on Slack for help in navigating the fixture system to write proper tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ test/fixtures/objective-c/quicktype
 /.node-persist
 /.vs
 .idea
+.DS_Store
--- a/src/quicktype-core/language/Kotlin.ts
+++ b/src/quicktype-core/language/Kotlin.ts
@@ -239,7 +239,7 @@ export class KotlinRenderer extends ConvenienceRenderer {
     }
 
     // (asarazan): I've broken out the following three functions
-    // because some renderers, such as kotlinx, can cope with `any`, while most probably can't.
+    // because some renderers, such as kotlinx, can cope with `any`, while some get mad.
     protected anyType(withIssues: boolean = false, noOptional: boolean = false): Sourcelike {
         const optional = noOptional ? "" : "?";
         return maybeAnnotated(withIssues, anyTypeIssueAnnotation, ["Any", optional]);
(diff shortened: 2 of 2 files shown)


asarazan:
Can we get this merged sometime soon? I mostly use the web portal for converting my types, and the lack of this feature is resulting uncompilable output

dvdsgl:
Hello! Can you please rebase?

dvdsgl:
Also, will this break existing code?

asarazan:
I was as careful as possible to leave existing behavior unchanged, but I had some trouble figuring out the test harness. That being said, it does seem like existing tests are green, and I've been using the changes from this PR in production with no problems.

Rebase performed!

dvdsgl:
Can you show me an example where this is useful? It seems that Kotlin supports unions—where are you getting Any where these Json types are better?

asarazan:
A pretty easy repro is to run it on the following typescript:
```typescript
export interface Foo {
    bar: any;
}
```

asarazan:
Actually, that last line seems to cause `Error: Maximum call stack size exceeded`, which I don't recall happening back when I first created the PR, but I may be misremembering. I'll see if I can get it resolved.

schani:
@asarazan I'm not sure what you mean by your last comment

> Scratch that, I was using bad input. Everything seems to be working as expected. Let me know if there's anything else I can address!

Are you saying quicktype does everything you expect without this PR, or that this PR works as expected?

asarazan:
PR works as expected. Would still very much appreciate a merge. For what
it's worth, I was responsible for contributing the original KotlinX
integration so I'm reasonably familiar with its workings 😀

schani:
Thank you so much!

Issue #336, opened 2017-12-12
Preserve property order to and from JSON Schema


schani:
We're using Javascript objects to represent JSON Schema, so we end up with essentially random order.  I think I'd like to keep using objects, but represent the keys as an array of name-type pairs, instead of as an object that maps from names to types.  That means for JSON Schema input we'll have to manually parse the data, and for output we might have to just print the JSON ourselves.

schani:
Depends on #646

amorscher:
Hi schani, are there any plans to fix this in the near future? If not I would try to solve the issue. 
Cheers

rambabusaravanan:
@schani 
Usually, I am much more concerned about the order of variables I declare. I believe the order matters a lot for clarity.
I am exploring this quicktype tool to convert JSONSchema to Typescript and may I expect this near future?

schani:
We do not take feature requests anymore. Please consider making a pull request instead. We encourage the use of coding agents, but please use at least an Opus-class model.

PR #1340, opened 2019-11-14
Added possibility to specify order in JSON-Schema #336

amorscher:
It is an alternative approach trying to resolve the ordering issue for object properties of JSON-Schemas.  Solution is to provide a propertyOrder string array on the object type. This array will be used to determine the order of properties for resulting language bindings. Example JSON:

```JSON
{
  "description": "Example for ordering properties",
  "type": "object",
  "properties": {
    "z_prop": {
      "type": "string"
    },
    "b_prop": {
      "type": "string"
    },
    "a_prop": {
      "type": "string"
    }
  },
  "propertyOrder": ["z_prop", "b_prop", "c_prop"]
}
``` 
Resulting typescript:
```typescript

export interface ExampleTest {
    z_prop?: string;
    b_prop?: string;
    a_prop?: string;
}

``` 

I would also like to add some tests but I need some support on that topic.
Perhaps we could first discuss if this is a valid approach for fixing the issue without re-writing the json parsing.

### DIFF
--- a/src/quicktype-core/input/JSONSchemaInput.ts
+++ b/src/quicktype-core/input/JSONSchemaInput.ts
@@ -634,10 +634,11 @@ async function addTypesInSchema(
         attributes: TypeAttributes,
         properties: StringMap,
         requiredArray: string[],
-        additionalProperties: any
+        additionalProperties: any,
+        sortKey: (k: string) => number | string = (k: string) => k.toLowerCase()
     ): Promise<TypeRef> {
         const required = new Set(requiredArray);
-        const propertiesMap = mapSortBy(mapFromObject(properties), (_, k) => k.toLowerCase());
+        const propertiesMap = mapSortBy(mapFromObject(properties), (_, k) => sortKey(k));
         const props = await mapMapSync(propertiesMap, async (propSchema, propName) => {
             const propLoc = loc.push("properties", propName);
             const t = await toType(
@@ -846,8 +847,15 @@ async function addTypesInSchema(
                 inferredAttributes,
                 combineProducedAttributes(({ forObject }) => forObject)
             );
+            const order = schema.propertyOrder ? schema.propertyOrder : [];
+            const orderKey = (propertyName: string) => {
+                // use the index of the order array
+                const index = order.indexOf(propertyName);
+                // if no index then use the property name
+                return index !== -1 ? index : propertyName.toLowerCase();
+            };
 
-            return await makeObject(loc, objectAttributes, properties, required, additionalProperties);
+            return await makeObject(loc, objectAttributes, properties, required, additionalProperties, orderKey);
         }
 
         async function makeTypesFromCases(cases: any, kind: string): Promise<TypeRef[]> {
--- a/src/quicktype-core/input/JSONSchemaInput.ts
+++ b/src/quicktype-core/input/JSONSchemaInput.ts
@@ -847,7 +847,7 @@ async function addTypesInSchema(
                 inferredAttributes,
                 combineProducedAttributes(({ forObject }) => forObject)
             );
-            const order = schema.propertyOrder ? schema.propertyOrder : [];
+            const order = schema.quicktypePropertyOrder ? schema.quicktypePropertyOrder : [];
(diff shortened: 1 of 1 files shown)


untitaker:
Is it a common pattern in quicktype to extend the schema of jsonschema that way? It seems there are also proposals floating around for adding a `propertyOrder` to the real jsonschema spec, and I would not like to see quicktype's definition conflict with a future one.

thegoldenmule:
@untitaker FWIW, I would posit that quicktype gets to interpret JSON any way it wants to. It seems to me, from poking around the json-schema repo, that the standards body is unlikely to make any sort of decision about this (seeing open issues from 2018). This `propertyOrder` approach seems to have precedent, so I would vote this be merged in with a proper update to the README.

untitaker:
@thegoldenmule assuming we care a lot about backwards-compat, it seems like an unnecessary risk to me to start supporting standards that may still change in a way that forces us to break backwards-compat

thegoldenmule:
@untitaker As I said, it appears the JSON schema body will never adopt an ordering standard because, according to the JSON spec, an object is explicitly unordered (see [spec](https://www.json.org/json-en.html)).

Quicktype, as I understand it, is much more concerned with target languages and already allows users to change the source input in non-standard ways to generate desired target output. A good example of this is the `/** @TJS-type integer */` you can put on fields to generate integer types. Order _does_ matter in many target languages, so we should adopt some way to specify order. Since this method has precedent (and the PR exists and passes tests), I suggest we accept.

@schani What's your take on this?

untitaker:
>As I said, it appears the JSON schema body will never adopt an ordering standard because, according to the JSON spec, an object is explicitly unordered (see spec).

My previous comments were referring to this [github thread suggesting exact syntax as proposed in op](https://github.com/json-schema/json-schema/issues/119), so if we merge this right now, we have three possible futures:

* if this exact proposal wins we're fine because we just implemented that
* if nothing is implemented at all we're also fine
* if some bigger project like [this one](https://github.com/json-schema/json-schema/issues/87) is started we might end up supporting a non-standard attribute even though there is a standard way. So we might end up supporting two ways of defining ordering, until we decide to deprecate one of them again

There is also the fourth, very unlikely, and worst-case outcome that the jsonschema ppl, sometime in the future, decides to define a field `propertyOrder` that has nothing to do with what we're doing here. In that case we end up being in direct violation of the spec, and we may not be able to fix that violation without breaking the semantics of *our version* of `propertyOrder` that people have come to rely upon.

>Quicktype, as I understand it, is much more concerned with target languages and already allows users to change the source input in non-standard ways to generate desired target output.

I don't object to that on principle and don't disagree that defining property order is useful, but I would rather like this to see this implemented in a way where it's less likely to conflict with future official extensions of the standard. A magic comment in typescript is less likely to conflict as typescript would probably rather just add new syntax.

thegoldenmule:
@untitaker Yah, I was looking at that discussion. In that, they appear to finish the conversation on "order will never be part of the JSON schema spec as it violates the JSON spec." From that thread, see https://github.com/json-schema/json-schema/issues/119#issuecomment-181330904

Having said that, prefixing with `quicktype` sounds like a good compromise to me. Very little additional work, and it's clearly non-standard.

untitaker:
I see, you're saying that it will not be part of the same spec and therefore not part of the same json blob input. I thought maybe they're gonna make a separate spec *document* that somehow extends the existing jsonschema-schema.

schani:
I'm not opposed to any of the suggested proposals.

thegoldenmule:
@schani thanks

@untitaker I don't want to steam roll you, I'm good with `quicktype` prefix if you feel strongly. Are you going to make that change?

untitaker:
Right, done!

untitaker:
hmm, why are all the diffs missing?

amorscher:
I also have no clue why the diffs are missing. What can we do to fix this!

schani:
I'll see if I have time to fix the diff bug this weekend.

Thank you for your work on this!

amorscher:
Nice that it finally made it to master!!! Thx for the support.

PR #1342, opened 2019-11-15
Allow for more lenient usage of golang optional properties. 

landonpoch:
Also allows package declaration with just-types option.

### DIFF
--- a/src/quicktype-core/language/Golang.ts
+++ b/src/quicktype-core/language/Golang.ts
@@ -84,9 +84,7 @@ function singleDescriptionComment(description: string[] | undefined): string {
 }
 
 function canOmitEmpty(cp: ClassProperty): boolean {
-    if (!cp.isOptional) return false;
-    const t = cp.type;
-    return ["union", "null", "any", "array"].indexOf(t.kind) < 0;
+    return cp.isOptional;
 }
 
 export class GoRenderer extends ConvenienceRenderer {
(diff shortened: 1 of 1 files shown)


dvdsgl:
This breaks a lot of existing Go code. Can you explain more about the rationale and make this optional?

landonpoch:
The rationale behind this is based off of a use case we have today.  We have an API that returns JSON.  This API has been implemented using golang.  However, we'd like to be able to prototype new JSON responses for new endpoints using typescript (essentially create our schema using ts interfaces), generate documentation for those responses using typedoc, and generate golang structs for those responses as well.  There's a lot of benefit in doing something like this: tight control over api changes via a separate repository, good documentation for consuming clients, json schemas for automated integration tests, etc...

The only problem is quicktype does not allow me to omitempty on arrays, anys and unions.  Take for example the following element returned as part of a response:

```typescript
interface Action {
    url: string;
    verb: "GET" | "PUT" | "POST" | "DELETE";
    body?: any;
}
```
This renders as follows:
```golang
type Action struct {
	Body interface{} `json:"body"`
	URL  string      `json:"url"` 
	Verb Verb        `json:"verb"`
}

type Verb string
const (
	Delete Verb = "DELETE"
	Get Verb = "GET"
	Post Verb = "POST"
	Put Verb = "PUT"
)
```
Every action I return now has to have a ```payload: null``` in it.  Also, empty arrays must be included as well.  This causes issues with our already existing json schemas which do recognize them as being optional properties and even omitted in certain scenarios.

I realize I need to get the tests in this PR to pass, but I honestly haven't been able to look to see how much would have to change to support this yet.  It just seems like it's an artificial limitation that should't need to exist.  I feel there are valid cases for optional anys, arrays and unions.

PR #1350, opened 2019-12-05
Allow for more lenient usage of golang optional properties.

landonpoch:


schani:
@landonpoch This break existing test cases where we parse JSON files and then serialize them again, and compare that output against the original JSON file.

PR #1377, opened 2020-01-20
Added support for Range with numbers in C#

Fylax:
JSON Schema supports [Range](https://json-schema.org/understanding-json-schema/reference/numeric.html#range) for numeric types.

This PR handles the range validation, properly adding a JSON converter that checks whether the (de)serialized value is in range.

Since quicktype does not seem to understand the `exclusiveMinimum` and `exclusiveMaximum` flags, they are assumed as false with this change.


schani:
@Fylax Thanks for this awesome PR! You're the first contributor to have discovered transformers.

Would you mind fixing these lint errors?

```
ERROR: /app/src/quicktype-core/Transformers.ts:984:1 - Consecutive blank lines are forbidden
  | ERROR: /app/src/quicktype-core/language/CSharp.ts:85:16 - This type is not allowed in the condition because it could be undefined. Only booleans are allowed.
  | ERROR: /app/src/quicktype-core/language/CSharp.ts:973:17 - Forbidden 'var' keyword, use 'let' or 'const' instead
  | ERROR: /app/src/quicktype-core/language/CSharp.ts:973:46 - == should be ===
```

Fylax:
@schani I fixed the lint errors and aligned the fork with the master.
Nevertheless, I am failing to understand why the CI diff task is failing.

schani:
@Fylax We have existing tests with those range attributes, which so far haven't done anything in C#, but not they do, so the generated code differs.  Thanks again!

Issue #1379, opened 2020-01-24
Ruby integration only works with outdated dry-rb gems


Morozzzko:
It's been a while since the integration was introduced, so the gems we use to list types are outdated

We need to update the integration and make sure it produces correct output for the latest gem versions. 

Would be cool if it works with `dry-types >= 1.0` and `dry-struct >= 1.0`

Semi-related issue: #1125
Source code: https://github.com/quicktype/quicktype/blob/master/src/quicktype-core/language/ruby/index.ts

Morozzzko:
As a side-note, the integration should probably give us a choice. Some people haven't upgraded their gems to 1.x, so they still need support for latest 0.X gems

Would be nice if someone could implement that too

dvdsgl:
Hello! I wrote the Ruby backend but I haven't used Ruby in over eight years. Would you consider contributing?

Morozzzko:
I'm afraid I might not get the chance to do it within the next few month, but I hope someone from Ruby community could pick this up

Seems like a decent task for beginners and enthusiasts, so I hope someone would be interested.

I'm sharing the link to the issue, and I'd be thankful if someone decided to take a shot at it

PR #1126, opened 2018-11-30
Update Ruby integer to Types::Integer (#1125)

dugsmith:
Fixes #1125

PR #1385, opened 2020-01-30
Add support of latest dry-types gem

rokkit:
Add configuration option to choose legacy or newest code generation for dry-types.

Related:
#1126 
#1379


Dockerfile
```diff
@@ -61,7 +61,7 @@ RUN /bin/bash -c "source /root/.sdkman/bin/sdkman-init.sh && sdk install kotlin"
 ENV PATH="/root/.sdkman/candidates/kotlin/current/bin:${PATH}"
 
 # Python
-RUN add-apt-repository ppa:jonathonf/python-3.6
+RUN add-apt-repository ppa:deadsnakes/ppa
```
Morozzzko: Could you please elaborate on the change? What's the difference between two repos?

src/quicktype-core/language/ruby/index.ts
```diff
@@ -159,12 +160,13 @@ export class RubyRenderer extends ConvenienceRenderer {
 
     private dryType(t: Type, isOptional: boolean = false): Sourcelike {
         const optional = isOptional ? ".optional" : "";
+        const intTypeDeclaration = this._options.legacyGemVesion ? "Int" : "Integer";
```
Morozzzko: I'm actually thinking if we should even use `Types::Int` instead of `Types::Integer` with legacy versions. It was deprecated since May 2018, the 0.13 release (see [changelog](https://github.com/dry-rb/dry-types/blob/master/CHANGELOG.md#0130-2018-05-03))

If we decide to support `Types::Int`, we'll have to add `require 'dry/types/compat/int'` though – it wouldn't work otherwise.

What do you think?

src/quicktype-core/language/ruby/index.ts
```diff
@@ -553,7 +555,9 @@ export class RubyRenderer extends ConvenienceRenderer {
 
     private emitTypesModule() {
         this.emitBlock(["module Types"], () => {
-            this.emitLine("include Dry::Types.module");
+            const includeType = this._options.legacyGemVesion ? "Dry::Types.module" : "Dry.Types()";
```
Morozzzko: Please keep in mind that those two modules are not exactly identical.

Back when we used `Dry::Types.module`, we could access _nominal_ types using `Types::String` and strict types using `Types::Strict::String`.

When we're using `Dry.Types()`, `Types::String` is now the same as `Types::Strict::String` – it's a possibly breaking behavior. See [changelog](https://github.com/dry-rb/dry-types/blob/master/CHANGELOG.md#changed-3)

If we want to keep the old behavior, we should replace it with `Dry.Types(default: :nominal)`

Dockerfile
rokkit: Because this repo was closed by owner.

Here is the [log](https://buildkite.com/typeguard/quicktype-pr/builds/1285#fe414035-af96-4c13-9dd2-6815272ebba9) with error

src/quicktype-core/language/ruby/index.ts
rokkit:
```suggestion
const includeType = this._options.legacyGemVesion ? "Dry::Types.module" : "Dry.Types(default: :nominal)";
```

I'm not good in Dry things.
You mean something like this, without any additional configuration?

src/quicktype-core/language/ruby/index.ts
rokkit: If only need to add one require to keep support, I think it can be done 🤔

src/quicktype-core/language/ruby/index.ts
Morozzzko: Yes, that'll do the trick

rokkit:
@Morozzzko When I try to pass ruby tests with new output, I get this exception.

Is this something wrong with code?
Here is complete output from test run.
[ruby-9e311e.zip](https://github.com/quicktype/quicktype/files/4206941/ruby-9e311e.zip)



> /Users/indmaksim/.rvm/gems/ruby-2.4.1/gems/dry-types-1.2.2/lib/dry/types/constructor/function.rb:67:in `Hash': can't convert Symbol into Hash (Dry::Types::CoercionError)
> 	from /Users/indmaksim/.rvm/gems/ruby-2.4.1/gems/dry-types-1.2.2/lib/dry/types/constructor/function.rb:67:in `call'
> 	from /Users/indmaksim/.rvm/gems/ruby-2.4.1/gems/dry-types-1.2.2/lib/dry/types/constructor.rb:59:in `call_unsafe'
> 	from /Users/indmaksim/.rvm/gems/ruby-2.4.1/gems/dry-types-1.2.2/lib/dry/types/type.rb:49:in `call'
> 	from /Users/indmaksim/workspace/quicktype/test/runs/ruby-9e311e/TopLevel.rb:123:in `from_dynamic!'
> 	from /Users/indmaksim/workspace/quicktype/test/runs/ruby-9e311e/TopLevel.rb:2339:in `block in from_dynamic!'
> 	from /Users/indmaksim/workspace/quicktype/test/runs/ruby-9e311e/TopLevel.rb:2339:in `map'
> 	from /Users/indmaksim/workspace/quicktype/test/runs/ruby-9e311e/TopLevel.rb:2339:in `from_dynamic!'
> 	from /Users/indmaksim/workspace/quicktype/test/runs/ruby-9e311e/TopLevel.rb:2385:in `from_json!'
> 	from main.rb:7:in `<main>'

dvdsgl:
Wow! Thanks for fixing the broken Python PPA — our CI has been broken for a while because of this. I will merge that into master before evaluating this.

dvdsgl:
Looks great except for the test failure—probably some small detail. We'll merge this if you can fix it.

We will also soon merge your PPA fix if it fixes CI, so you will have a small conflict to fix.

Morozzzko:
@rokkit sorry it took me so long to answer. I'll take a look today and come back if I get any ideas

Morozzzko:
This thing is fairly tricky. Here's what happens:

1. We have a [test script](https://github.com/quicktype/quicktype/blob/7470b3139aebce15ec7bbbe00b2a91e16cac99c1/test/fixtures/ruby/main.rb) that runs the script and prints the output
2. We run it with the [Gemfile](https://github.com/quicktype/quicktype/blob/7470b3139aebce15ec7bbbe00b2a91e16cac99c1/test/fixtures/ruby/Gemfile) which locks dry-types to verison 1.2
3. However, when we test the pre-1.0 dry-rb code, it fails because the version is different and the `compat` doesn't work anymore. 

Here's what might help us:

1. Update test defaults and not test the `pre-1.0` version. It's a bad solution, but it works
2. Use `bundler/inline` in the generated script. It will help us, but it's going to make generated code verbose. I like this solution as it clearly communicates the gem versions
3. Figure out a way to pass arguments to the `main.rb` script using argv or env. It's good for tests, but may be tricky. 

I'll try the option 3 and come back to you. You could try to make it work. See [bundler docs](https://bundler.io/guides/bundler_in_a_single_file_ruby_script.html) for reference.

Morozzzko:
However, it might be easier to, make it a bit different.

See what I mean:

We're getting problems because of `compat`. Let's get rid of it, then. `Types::Strict::Integer` works with dry-types over 0.13.1. I think we can drop it.

So the only real difference would be the way we include types. It's good enough for now, I think.

src/quicktype-core/language/ruby/index.ts
```diff
@@ -553,7 +555,9 @@ export class RubyRenderer extends ConvenienceRenderer {
 
     private emitTypesModule() {
         this.emitBlock(["module Types"], () => {
-            this.emitLine("include Dry::Types.module");
+            const includeType = this._options.legacyGemVesion ? "Dry::Types.module" : "Dry.Types(default: :nominal)";
+            const intTypeDeclaration = this._options.legacyGemVesion ? "Integer" : "Int";
```
Morozzzko: Also this might be a bug. Should probably be the other way around:

```suggestion
            const intTypeDeclaration = this._options.legacyGemVesion ? "Int" : "Integer";
```

Like there:

https://github.com/quicktype/quicktype/pull/1385/files#diff-94c47b5848b3bd8b852453de72cac7f5R163

schani:
@rokkit This is the run failure:

```
/usr/local/lib/ruby/gems/2.7.0/gems/bundler-1.17.3/lib/bundler/rubygems_integration.rb:200: warning: constant Gem::ConfigMap is deprecated

/app/test/runs/ruby-bef364/vendor/bundle/ruby/2.7.0/gems/dry-types-1.2.0/lib/dry/types/any.rb:18: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call

/app/test/runs/ruby-bef364/vendor/bundle/ruby/2.7.0/gems/dry-types-1.2.0/lib/dry/types/nominal.rb:48: warning: The called method `initialize' is defined here

/app/test/runs/ruby-bef364/vendor/bundle/ruby/2.7.0/gems/dry-types-1.2.0/lib/dry/types/decorator.rb:18: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call

/app/test/runs/ruby-bef364/vendor/bundle/ruby/2.7.0/gems/dry-types-1.2.0/lib/dry/types/options.rb:15: warning: The called method `initialize' is defined here

/app/test/runs/ruby-bef364/vendor/bundle/ruby/2.7.0/gems/dry-types-1.2.0/lib/dry/types/sum.rb:44: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call

/app/test/runs/ruby-bef364/vendor/bundle/ruby/2.7.0/gems/dry-types-1.2.0/lib/dry/types/meta.rb:9: warning: The called method `initialize' is defined here

LoadError: cannot load such file -- dry/types/compat/int

  /app/test/runs/ruby-bef364/TopLevel.rb:14:in `require'

  /app/test/runs/ruby-bef364/TopLevel.rb:14:in `<top (required)>'

  main.rb:4:in `require'

  main.rb:4:in `<top (required)>'
```

jedschneider:
Hi Curious what is left to get this over the line and merged?

I wanted to point out that the built in types include JSON types geared specifically towards coercion of JSON values: https://dry-rb.org/gems/dry-types/1.2/built-in-types/

maybe quicktype is already leveraging this but at a quick glance, it didn't seem to be part of the upgrade.

dvdsgl:
@jedschneider there are merge conflicts and some test failures. Once those are fixed, we can reevaluate for merge.

dvdsgl:
@rokkit our CI is fixed and Ruby tests are enabled here. I believe there are errors in case you can take a look!

dvdsgl:
Moved to #2080

Issue #1331, opened 2019-10-31
Infection? ChiangMaiGoose, BarcelonaHawaiianMonkSeal and BasseTerreElephant, TrapaniSnowLeopard


oxsd:
I love this extension, but...  It created a type as "ChiangMaiGoose."  Researching online, i found another schmea definition that contained this along with "BarcelonaHawaiianMonkSeal" and "BasseTerreElephant."  Now, I love animals more than most folks ,but... :) 

See also:
https://docs.rs/vega/0.0.1/x86_64-apple-darwin/vega/schema/enum.ChiangMaiGoose.html

And, this gist which mentions this extension by name with "BarcelonaHawaiianMonkSeal":
https://gist.github.com/schani/681fd46e7c67871e49143b7dffd1717c

oxsd:
Ooh!  Now it's a TrapaniSnowLeopard.  I love critters!  :)

oxsd:
TrapaniSnowLeopard showed up here, btw:  
https://www.reddit.com/r/typescript/comments/d3mko0/typescript_still_complains_after_hasownproperty/

oxsd:
Not a virus but a bug.  quicktype-core/TypeNames.js is calling Chance to in initTypeNames() which is generating the critter names.  This is because i have a definition that has over 20 names, which exceeds the value of `tooManyNamesThreshold`.

VolatileCoder:
This is idiotic. Why a limit of 20? 

For those of you facing this problem, bump the threshold to something more reasonable. Like 1000. Because, you know, this is 2019.

VolatileCoder:
@schani 

This arbitrary threshold of 20 came from here; this makes this product unusable off the shelf in any real business environment. 

https://github.com/quicktype/quicktype/commit/8bc4127a64eb660d9cc59a1e292410510252540c

inferrinizzard:
Resolved in #1406

PR #1406, opened 2020-02-24
Increasing tooManyNamesThreshold to 1000 from 20.

VolatileCoder:
As referenced in #1331, the tooManyNamesThreshold is set too low for real business cases and needs to be increased. 

This looks like it was set to 20 by @schani during some testing and was never set back to a reasonable value. 

This has the impact of people misreporting this behavior as a defect and is an embarrassment of this otherwise fantastic project.


dvdsgl:
This looks like a great change, thank you. Can you please bump the major version in package.json? This will break a lot of code.

PR #1428, opened 2020-03-26
kotlinx/fix-deprecation

asarazan:
### Replacement of deprecated class
This PR fixes the deprecated use of KotlinX's `StringDescriptor`. We now use `PrimitiveDescriptor`, which is Jetbrains' recommended approach.

### Top Level Declarations
We now properly override `emitTopLevelMap` and `emitTopLevelArray` to use `JsonObject` and `JsonArray`. I'm not entirely sure the old declarations would have even been useable.

### Quality of Life
I have also taken this opportunity to introduce some improvements based on my own usage of the renderer.
* Invalid enum strings will now give you more information in the exception message
* The `serialize` method for the generated instance of `KSerializer` now properly uses its superclass' parameter names, which had previously been generating a compiler warning.

### DIFF
--- a/src/quicktype-core/language/Kotlin.ts
+++ b/src/quicktype-core/language/Kotlin.ts
@@ -1021,6 +1021,20 @@ export class KotlinXRenderer extends KotlinRenderer {
         return super.mapType(mapType, withIssues, noOptional);
     }
 
+    protected emitTopLevelMap(t: MapType, name: Name): void {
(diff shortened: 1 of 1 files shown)


asarazan:
Also to put everyone's mind at ease: I am the original author of the KotlinX Renderer, and have personally validated that this produces the desired output for nontrivial input <3

schani:
Thank you so much!

PR #1435, opened 2020-04-07
still ensure python valid names even without style

brunorzn:
when `pyOptions.nicePropertyNames` was set to false, Python.ts would just use
original provided names for properties, which can be invalid.

Instead, we now ensure that the names are valid, but don't force
snake_case when `pyOptions.nicePropertyNames` is set to false.

We also do the same for Enums, based on the same option.


schani:
Thank you so much!

Issue #1353, opened 2019-12-09
"Just Types" for C# omits namespace


atrauzzi:
I'm looking to generate plain C# types from TypeScript with no dependencies on any external libraries. Unfortunately, when I generate using "just types", I end up losing the namespace (as well as the json.net dependencies).

Is there any way I can have namespaces without json.net?

thegoldenmule:
I made a PR here: https://github.com/quicktype/quicktype/pull/1436

thegoldenmule:
@atrauzzi After some discussion, it was deemed that `just-types` serves some specific purposes that we can't break backward compat for. Instead, I started a new PR with a new option for the C# renderer: `just-types-and-namespace`. See https://github.com/quicktype/quicktype/pull/1461.

atrauzzi:
Slick!

thegoldenmule:
@atrauzzi FYI this has been merged into master.

thegoldenmule:
This has been resolved.

PR #1436, opened 2020-04-09
Fix issue where just-types feature does not emit namespace/package

thegoldenmule:
This issue affects C# and Go, though I have not checked other targets.

C# issue: https://github.com/quicktype/quicktype/issues/1353

Proposed C# solution: Rather than inferring from the `helpers` and `attributes` options, explicitly define namespace as an option. The fix here is logically equivalent for `complete` and `attributes-only` features.

Proposed Go solution: Always output the package.

### DIFF
--- a/src/quicktype-core/language/CSharp.ts
+++ b/src/quicktype-core/language/CSharp.ts
@@ -578,9 +578,9 @@ export class CSharpRenderer extends ConvenienceRenderer {
 
 export const newtonsoftCSharpOptions = Object.assign({}, cSharpOptions, {
     features: new EnumOption("features", "Output features", [
-        ["complete", { helpers: true, attributes: true }],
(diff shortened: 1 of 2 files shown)


thegoldenmule:
@untitaker That would be handy for some targets, though in C# + Java's case (and probably others) would not necessarily lead to idomatic code: the entire file needs indented and usually `using` statements that the renderer needs to write would need to live outside of the namespace.

For these two renderers, at the suggestion of @schani, I will make a new attribute option (perhaps `just-types-and-namespace`) since `just-types` is meant more for copy/pasting.

thegoldenmule:
@schani Nevermind! I was expecting a specific step but was able to find the issues by digging through the logs. Thanks

schani:
@thegoldenmule Sounds good!

thegoldenmule:
I will create a new branch for this.

Issue #1099, opened 2018-10-24
Feature - Using Lombok


g4rcez:
My feature suggestion is using [lombok](https://projectlombok.org) instead getters and setters of tradicional Java Style.

schani:
@vandalvnl Would you consider contributing this?  We'd be happy to help you along.

schani:
That's awesome!

A few pointers:

* [Here's the code](https://github.com/quicktype/quicktype/blob/master/src/quicktype-core/language/Java.ts) that produces the Java output.

* Right now it's just one class, `JavaRenderer`, that produces Java classes with all the Jackson stuff.  [C# is better separated](https://github.com/quicktype/quicktype/blob/master/src/quicktype-core/language/CSharp.ts).  There, we have a `CSharpRenderer` which produces "naked" C# classes, and a subclass `NewtonsoftCSharpRenderer` which also produces the additional stuff for Json.NET.

* The proper way to do it is to first separate `JavaRenderer` like with C# into a `JavaRenderer` and a `JacksonJavaRenderer`.  That would be a great first pull request.

* Then you can subclass `JavaRenderer` with a `LombokJavaRenderer` which generates the Lombok code you need.

Again, if you need any help at all, please don't hesitate to ask!  You might also want to [join us on Slack](http://slack.quicktype.io) where we can discuss more efficiently.

Cheers!

PR #1443, opened 2020-04-25
Add support for Lombok in Java

skornel02:
This pull request extends the already existing JavaTargetLanguage with Lombok capabilities. 
Originally it was suggested that a separate Lombok renderer should be created, but I think that it is not necessary because Lombok is not replacing the JSON part, only removing the getter and setter boilerplate.

related:
#1099


skornel02:
'Pseudo language' test implementation seems to break it, because the normal java and java-lombok's pom.xml gets mixed up (therefore errors about package lombok does not exist). Running either one locally passes, but running both reproduces the issue. 

I don't see a clear way to have an integration test for both at the same time. Any suggestions?

schani:
This is amazing!  Thank you so much!

PR #1451, opened 2020-05-02
Fix build issues and missing exports around quicktype-core

untitaker:
Need this type for programmatically using quicktype in https://github.com/untitaker/quicktype-markdown


untitaker:
I hope you don't mind my second commit, I had to do this because tsc would complain about missing imports/unknown namespaces otherwise.

PR #1458, opened 2020-05-06
Move FetchingJSONSchemaStore to quicktype-core to allow it to be used  in applications using quicktype as a library

jpsav:
quicktype-core includes a JSONSchemaStore class that can be used to load JSON schema files however this does not support resolving JSON schema references (`$ref`) to different files or URI's. 

The quicktype CLI supports resolving references to other files or URI's using a FetchingJSONSchemaStore that extends the JSONSchemaStore exposed by quicktype-core.

When using quicktype programmatically via quicktype-core it is useful to be able to resolve JSON schema's that reference other files and URI's. 

This change moves the FetchingJSONSchemaStore from the cli to quicktype-core to enable programmatic use. This avoids users of quicktype-core  needing to re-implement this functionality. 

All this change really involves is moving the FetchingJSONSchemaStore  and its dependencies to quicktype-core, then exporting the class and its dependencies and updating imports.

### Testing

```
script/dev
npm run build 
npm run test
```


schani:
Doesn't this make `quicktype-core` require Node?  `quicktype-core` must be able to run in the browser, too.

jpsav:
> Doesn't this make `quicktype-core` require Node? `quicktype-core` must be able to run in the browser, too.

Apologies I did not realize it was a requirement for quicktype-core to run in the browser.  With this change that would be achievable using webpack or Browserify but it would be wrong to force consumers to use a particular build system.

That said I still feel that the functionality in `FetchingJSONSchemaStore` is valuable to expose as it is quite common for JSON schemas to reference other files or URI's. This does not really apply to browsers though as it references to other files depend on a file system.

Would it be acceptable to simply export `FetchingJSONSchemaStore` from quciktype ? so that any consumers who want to use the `FetchingJSONSchemaStore` could depend on both quicktype and quicktype-core.

dvdsgl:
I believe we can use `isomorphic-unfetch` for `fetch` that works in browser and node, then move this to `quicktype-core`. Please try it.

jpsav:
Thanks for suggestion Im happy to try that but the `FetchingJSONSchemaStore also uses Node's FS module to read files as a stream. Im not sure if that is even possible in browser as browsers have limited access to the file system

jpsav:
@dvdsgl @schani, I have done some research and I do not think it is possible for code running in the browser to open arbitrary files without user input. If this was possible it would obviously be a massive security vulnerability.

Would it be acceptable to change the ```FetchingJSONSchemaStore``` to use `isomorphic-fetch` so that it will be able to resolve schemas that are available at a URL both in Node and the browser and then throw some kind of invalid usage exception if the the `$ref` is a relative file path when not running in Node ?

schani:
I'm ok with that.

jpsav:
I've added a new revision that makes the `FetchingJSONSchemaStore` isomorphic. The change
* Switches to using isomorphic-fetch instead of node fetch
* Switches to readable-stream (isomorphic replacement for node strem)
* User browser or node to guard calls to fs so we do not try to call methods on fs when in the browser
* Added the following to the `package.json` for quicktype-core to make bundlers for the browser handle imports of `fs`
```
"browser": {
        "fs": false
    }
```

jpsav:
I saw that the build kite diff-against-last-master step failed. I have struggled to figure out how to reproduce this locally in order to debug. Im not even sure what it is that is being tested by that  test step.

Any pointers would be much appreciated. 

Thanks !

Issue #1392, opened 2020-02-13
Option to keep union of classes as such


jbussemaker:
When defining a type as the union of two referenced object types using `anyOf`, quicktype merges the classes and references this merged class as the type. Sometimes it might be needed to keep the defined objects as-is, and reference the union of the two classes.

Currently, this JSON schema:
```json
{
  "$ref": "#/definitions/Problem",
  "definitions": {
    "Problem": {
      "type": "object",
      "properties": {
        "variables": { "type": "array", "items":
        { "anyOf": [
            { "$ref": "#/definitions/ContinuousVariable" },
            { "$ref": "#/definitions/DiscreteVariable" }
            ] }
        }
      },
      "required": ["variables"]
    },
    "ContinuousVariable": {
        "type": "object",
        "properties": {
            "lowerBound": {"type": "number"},
            "upperBound": {"type": "number"}
        },
        "required": ["lowerBound", "upperBound"]
    },
    "DiscreteVariable": {
        "type": "object",
        "properties": {
            "options": {"type": "array", "items": {"type": "string"}}
        },
        "required": ["options"]
    }
  }
}
```

Produces this code (TypeScript):
```typescript
export interface Problem {
    variables: Variable[];
}

export interface Variable {
    lowerBound?: number;
    upperBound?: number;
    options?:    string[];
}
```

However, the structure would be more understandable, if the following code would be generated:

```typescript
export interface Problem {
    variables: Array<ContinuousVariable | DiscreteVariable>;
}

export interface ContinuousVariable {
    lowerBound: number;
    upperBound: number;
}

export interface DiscreteVariable {
    options: string[];
}
```

This behavior could be achieved by including an additional parser option, and then if this option is set, modifying `include` callback passed to `makeGroupsToFlatten` in `FlattenUnions.flattenUnions` to something like:
```typescript
import { iterableEvery } from 'collection-utils';
import { ObjectType } from "../Type";

let cliOptionPreserveUnionClasses = true;

const groups = makeGroupsToFlatten(nonCanonicalUnions, members => {
    messageAssert(members.size > 0, "IRNoEmptyUnions", {});
    if (!iterableSome(members, m => m instanceof IntersectionType)) {

        // If all the union members are objects, we do not flatten them to preserve the
        // individual classes
        if (cliOptionPreserveUnionClasses) {
            const allObjects = iterableEvery(members, m => m instanceof ObjectType);
            if (members.size > 1 && allObjects) return false;
        }
        return true;
    }

[...]

PR #1460, opened 2020-05-06
Avoid flattening class unions for typescript

untitaker:
TypeScript supports unions of arbitrary types so we do not have to merge union variants, generally speaking. Add a new language prop that completely skips union flattening.

This likely won't work since it bypasses number variant flattening which is required for typescript. I am thinking that maybe number flattening should be done inside the renderer somehow anyway, as right now the typescript renderer has to handle `double` and `integer` cases explicitly anyway. What if we deduplicated the emitted tokens instead, post-rendering, instead of rewriting the type graph?

The alternative outlined in #1392 to only stop union flattening if the union consists of objects and maps entirely seems to be less transparent since it randomly introduces the flattening behavior for cases like `type MyUnion = number | ClassA | ClassB | null;`

I think we will have to refactor UnionBuilder *somehow* but I am not sure yet how to do that.


dvdsgl:
@untitaker we’ve fixed our CI and have some test errors for these changes. I’ve changed this to a Draft until you have a chance to take a look. Thanks!

dvdsgl:
Please reopen if you plan to take another crack at this. Thanks!

PR #1461, opened 2020-05-06
Just types and package/namespace

thegoldenmule:
- Added `just-types-and-namespace` to C# renderer.
- Added `just-types-and-package` to Go renderer.

### DIFF
--- a/src/quicktype-core/language/CSharp.ts
+++ b/src/quicktype-core/language/CSharp.ts
@@ -9,7 +9,7 @@ import {
     ArrayType,
     TransformedStringTypeKind,
     PrimitiveStringTypeKind,
-    PrimitiveType
+    PrimitiveType,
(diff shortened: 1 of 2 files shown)


thegoldenmule:
@schani The diff test fails because the prop-types emitter now emits a different comment on the file (see [diff](https://s3.amazonaws.com/quicktype-outputs/diffs/55877b44fd22501eb5f706640259a2761d5db450-9f00b88229e45f7502702bb304c96b1e21d3092a.diff) from S3). Could we either force merge this or somehow get the new prop-types output up? I'm not sure why it doesn't have it.

schani:
Oh, maybe it builds the commit PR, vs the commit PR merged with `master`?  That would explain that.

In any case, it's good to go.  Thanks!

PR #1465, opened 2020-05-12
Option to ignore unknown properties during runtime typecheck

dvdsgl:
By default, JavaScript, Flow, and TypeScript will throw an error if an unexpected property is found during runtime typecheck.


dvdsgl:
We're getting a failing diff against master due to the new PropTypes language, but it doesn't seem to have anything to do with this or other PRs.

Issue #316, opened 2017-12-05
Support date/time types in all target languages


schani:
* [ ] Swift
* [ ] Go
* [ ] TypeScript
* [ ] C++
* [ ] Java
* [ ] Elm

## Swift

https://useyourloaf.com/blog/swift-codable-with-custom-dates/
https://stackoverflow.com/questions/46537790/iso8601-date-json-decoding-using-swift4

mattbudish:
I would like to second that with the addition of C#.

schani:
@mattbudish The reason C# is not on the list is because it's [already implemented](https://github.com/quicktype/quicktype/pull/312) ;-)

buguibu:
I've made this for Swift but it uses the same _dateDecodingStrategy_ for the whole JSON data and need to be passed at top of structure init even that date properties are in deeper structures inside the original JSON, hope it helps:

```
func newJSONDecoder(dateFormatter: DateFormatter? = nil) -> JSONDecoder {
    let decoder = JSONDecoder()
    if #available(iOS 10.0, OSX 10.12, tvOS 10.0, watchOS 3.0, *) {
        if let formatter = dateFormatter {
            decoder.dateDecodingStrategy = .formatted(formatter)
        } else {
            decoder.dateDecodingStrategy = .iso8601
        }
    }
    return decoder
}
```

```
struct LeafElement: Codable {
    let startDate: Date?
    
    static var dateFormatter: DateFormatter {
        return DateFormatter(withFormat: Constants.defaultDateFormat,
                             locale: Locale.current.languageCode ?? Constants.defaultLocale
        )
    }
```

```
extension Array where Element == RootJsonElement.Element {
    init(data: Data) throws {
        self = try newJSONDecoder(dateFormatter: LeafElement.dateFormatter).decode(LeafElement.self, from: data)
    }
```

schani:
@buguibu We already have date-time in Swift.  I didn't update this issue - sorry about that.

https://app.quicktype.io?share=1O1VFdIVOTzAjHHk4YkA

schani:
@buguibu Sorry for getting back to you so late.

The problem is that Swift doesn't parse that date format - try it out in a playground:

```swift
// To parse the JSON, add this file to your project and do:
//
//   let pokedex = try Pokedex(json)

import Foundation

struct Pokedex: Codable {
    let foo: Date
}

// MARK: Convenience initializers and mutators

extension Pokedex {
    init(data: Data) throws {
[...]

PR #1470, opened 2020-05-12
Java renderer separation, datetime and UUID support.

### LINKED ISSUE #1262: use java.time for JSON schema string with format date-time
Given the following JSON schema:
```json
{
  "type": "object",
  "$schema": "http://json-schema.org/draft-06/schema",
  "properties": {
    "timestamp": {
      "type": "string",
      "format": "date-time"
    }
  }
}
```

The generated Java POJO class uses `String` for the property type:
```java
public class Test {
    private String timestamp;
    @JsonProperty("timestamp")
    public String getTimestamp() { return timestamp; }
    @JsonProperty("timestamp")
    public void setTimestamp(String value) { this.timestamp = value; }
}
```

I expect the generated Java POJO to use `java.time.Instant` or `java.time.ZonedDateTime` for the property's type instead of `String`:

```java
public class Test {
    private ZonedDateTime timestamp;

    @JsonProperty("timestamp")
    public ZonedDateTime getTimestamp() { return timestamp; }
    @JsonProperty("timestamp")
    public void setTimestamp(ZonedDateTime value) { this.timestamp = value; }
}
```

Seems like it should be as simple as adding some conditional branching here:

https://github.com/quicktype/quicktype/blob/2ac4aa372772418a5b6c649fecaa99381ebcc8b8/src/quicktype-core/language/Java.ts#L339-L346

Should we use `Instant`, `LocalDateTime` or `ZonedDateTime`? We also have to ensure that Jackson reads and writes an ISO8601 string.

skornel02:
This pullrequest contains the following changes:
- Separated JavaRenderer from Jackson Renderer
- Only types uses JavaRenderer otherwise JacksonRenderer is used
- Added UUID support
- Added DateTime support with modular extension with two implementation:
    1. Java 8 date time: Supports all unit test cases and should be used.
    2. Legacy date time with java.util.Date: Does not support every edge case and requirement of all unit tests. 
- Fixed Lombok keyword reservation
- Added setting for Lombok to generate accessors
- Java imports are now determined based on requirement (for types)

Existing issues:
fixes #1262 (bit differently than suggested)
helps with #316

I'm open to debate everything about the changes, please feel free to ask!


dvdsgl:
Whoa! I haven’t written Java in 14 years but whoa!

PR #1475, opened 2020-05-16
Fixes undefined Key error when passing unexpected values to non-irregular objects

davidhodge:
Problem:

I experienced a crash when I had unexpected JSON data sent to "-fromJSONDictionary:" for an object. I learned that there was a fix in place to address this problem already, but the fix only applied to objects that were "irregular". From reading the code, it sounds like "irregular" objects are defined as objects where the JSON property names and the objc property names would differ.

Solution:

Emit the check for a valid property, regardless of other or not the affected object is flagged as having irregular properties.

Also:

Thank you!


davidhodge:
Hold the phone. I committed too much stuff here. Will retract my project file change and my build setting change

davidhodge:
Alrighty, amended my commit, but now I have not a clue what's up with these tests. Will investigate, but would also love pointers. Thank you

schani:
Thank you so much for your fix!

PR #1491, opened 2020-06-03
Change slack link in readme to invite based

derberg:
The current link takes you to slack sign-in page, but if it is your first time (like mine) you are doomed. I had to look for a proper invite-like link on a website, so I just want to avoid others follow my struggle.

Also fixed broken link to example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ The argument to `quicktype` is a complex object with many optional properties. [
 
 ## Contributing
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -206,11 +206,11 @@ async function main() {
 main();
 ```
 
-The argument to `quicktype` is a complex object with many optional properties. [Explore its definition](https://github.com/quicktype/quicktype/blob/javascript-use-examples/src/quicktype-core/Run.ts#L171) to understand what options are allowed.
+The argument to `quicktype` is a complex object with many optional properties. [Explore its definition](https://github.com/quicktype/quicktype-vscode/blob/master/src/extension.ts#L141) to understand what options are allowed.
```
skornel02, line 209: Isn't the definition in quicktype-vscode only a subset of total functionality?
[The main entry point](https://github.com/quicktype/quicktype/blob/master/src/quicktype-core/Run.ts#L119) has more options to offer.

derberg: @skornel02 you are absolutely right, I've just learned it today after digging into the project. The learning curve here is pretty steep 😅

derberg: changed

schani:
Thank you! :-)

Issue #1210, opened 2019-04-24
Adapt Rust output for compatibility with Rust 2018 edition


yonkeltron:
Hello,

Back at the end of last year, [Rust 2018 was announced](https://blog.rust-lang.org/2018/12/06/Rust-1.31-and-rust-2018.html). It brought several changes which make the current Rust output of the `quicktype` command incompatible with the 2018 edition. Specifically, the new macro and module system changes [described in the edition guide](https://doc.rust-lang.org/edition-guide/rust-2018/index.html) mean that using Serde to implement the compiler-derived traits of `Serialize` and `Deserialize` requires a different `use` directive than was previously required. The change is described in detail [here](https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html).

The old way:

```rust
#[macro_use]
extern crate serde_derive;
```

The new way:

```
use serde::{Serialize, Deserialize};
```

As a result of this improvement, we need to post-process all Rust code generated by the `quicktype` command in order to replace the old line with this new line.

While I advocate fixing this problem and making the new output the default behavior, as the adoption of Rust 2018 is still underway, I suggest adding a CLI flag which can revert to the legacy behavior for use with Rust versions prior to 1.31.

Some information about my environment (where Quicktype was installed via `npm`):

```bash
$ quicktype --version
quicktype version 15.0.182
Visit quicktype.io for more info.

$ uname -a
Darwin <HOSTNAME> 16.7.0 Darwin Kernel Version 16.7.0: Wed Feb 27 00:29:57 PST 2019; root:xnu-3789.73.43~1/RELEASE_X86_64 x86_64

$ node --version
v11.14.0

$ npm --version
6.9.0
```

Please let me know if I can provide any additional information to aid in the resolution of this problem. I love Quicktype and hope to see its wider adoption in the Rust world.

Thanks,
+Jonathan

ajventurini:
Would love to see this updated with the legacy version option.

briansmith-agilebits:
I noticed that the generated code doesn't even depend on `serde_json` as far as I can tell; it only depends on `serde_derive`, at least in my case. I had to add `serde_json` dependencies into my crates' Cargo.toml just to cope with the `extern crate serde_json;`. If I remove `extern crate serde_json;` then my crates compile just fine. (I'm using the 2018 Edition.)

PR #1496, opened 2020-06-10
Add edition 2018 style use sedre rather than extern crate

### LINKED ISSUE #1493: Use of serde instead of serde_json
https://github.com/quicktype/quicktype/blob/7a6a0fbdaa817ac36d0acf8701fe92e36ae261d6/src/quicktype-core/language/Rust.ts#L378
Think this line might be outdated. I changed it to `use serde::{Deserialize, Serialize};` in my own code.

mocsy:
Closes #1493 and #1210

### DIFF
--- a/src/quicktype-core/language/Rust.ts
+++ b/src/quicktype-core/language/Rust.ts
@@ -37,13 +37,18 @@ export enum Visibility {
 }
 
 export const rustOptions = {
-    density: new EnumOption("density", "Density", [["normal", Density.Normal], ["dense", Density.Dense]]),
+    density: new EnumOption("density", "Density", [
+        ["normal", Density.Normal],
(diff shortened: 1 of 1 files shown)


schani:
Thank you so much!

Snazzie:
How do we specifiy this in cli?

mocsy:
```
echo '{ "name": "David" }' | quicktype -l rust --derive-debug --edition-2018 --leading-comments
```
Would be the right answer.

Doing `npm i -g quicktype` doesn't work for me either and master also doesn't seem to build on my machine atm.

Snazzie:
yeah, i tried that at first but i get `Error: Option parsing failed: Unknown option: --edition-2018` :P

Issue #1288, opened 2019-08-13
Go: missing `omitempty` tag for optional arrays


Nutzlastbohne:
If an optional field is of type `array`, quicktype doesn't add the `omitempty` tag

example schema:
### schema.json
```json
{
  "type": "object",
  "properties": {
    "id": {
      "type": "string"
    },
    "name": {
      "type": "string"
    },
    "users": {
      "type": "array",
      "items": {
        "type": "string",
        "format": "uuid"
      }
    }
  },
  "required": ["id"]
}
```

generate code via `quicktype -s schema -o types.go schema.json`
### types.go
```go
// This file was generated from JSON Schema using quicktype, do not modify it directly.
// To parse and unparse this JSON data, add this code to your project and do:
//
//    types, err := UnmarshalTypes(bytes)
//    bytes, err = types.Marshal()

package main

import "encoding/json"

func UnmarshalTypes(data []byte) (Types, error) {
	var r Types
	err := json.Unmarshal(data, &r)
	return r, err
}

func (r *Types) Marshal() ([]byte, error) {
	return json.Marshal(r)
}

type Types struct {
	ID    string   `json:"id"`            
	Name  *string  `json:"name,omitempty"`
	Users []string `json:"users"`         
}
```

The only required field is `ID`, so `Name` correctly gets the `omitempty` tag. But `Users` is missing it. Because of that, marshaling an instance of `Types` with a nil `Users` array will always generate an invalid json (i.e. won't pass validation with the above schema):
```json
{
    "id":"",
    "users":null
}
```

The `omitempty` tag must be added for optional arrays also, so marshaling leaves them out.

haya14busa:
Any update on this?

schani:
I just retested this on current master with the exact schema and command from this issue (`quicktype -s schema --just-types -o types.go schema.json`). The generated Go now correctly includes the tag for the optional array field:

```go
type Types struct {
	ID    string   `json:"id"`
	Name  *string  `json:"name,omitempty"`
	Users []string `json:"users,omitempty"`
}
```

`Users` gets `omitempty` just like the optional `Name` field, so a nil slice is now omitted from the marshaled JSON instead of producing `"users":null`, which matches what was requested here. This will be included in the next release. If you can still reproduce the missing tag on a current build, please reopen with the schema you're using and I'll take another look.

PR #1513, opened 2020-06-30
Go: Support `omitempty` tag for optional arrays

haya14busa:
Optional array can be (or should be) marked as omitempty.

ref: #1288


schani:
Thank you so much!

Issue #1391, opened 2020-02-13
Correct support for optional class properties in Python


jbussemaker:
Defining a class property as optional (i.e. by not including it in the "required" list) in a JSON schema does not work for Python currently, since Python does not distinguish between `null` and `undefined` as JavaScript/TypeScript: for Python these are both represented as `None` values. This leads to a problem when serializing from Python to JSON to TypeScript, since the Python side would have added the class property in the JSON serialization with a `null` value, while the TypeScript parser expects either a value or undefined, and thus raises an error because of the `null` value.

The behavior defined above stems from the `optionalToNullable` conversion done when encountering a `PythonRenderer` (because `supportsOptionalClassProperties` is `false`).

However, in my opinion it looks much nicer to define a class property as optional instead of as a union between the type and `null` (i.e. `prop?: string` looks better than `prop: string | null`).

The following could be a solution for adding proper support for optional class properties to Python:
1. Set `supportsOptionalClassProperties` to `true`
2. Modify `PythonRenderer.pythonType` to also accept the `isOptional` flag from the class property (or just the ClassProperty as its only argument and derive the type from that), and wrap the produces type descriptor in `Optional[<type>]` if `isOptional` is `true`
3. Modify `JSONPythonRenderer.emitClassMembers` to include some extra code in the `to_dict` function for removing optional fields from the generated dict for optional properties if their value is `None`:
```python
class MyObject:
    opt_prop: Optional[str]

    ....

    def to_dict(self) -> dict:
        result: dict = {}
        if self.opt_prop is not None:
            result["optProp"] = from_str(self.opt_prop)
        return result
```

PR #1516, opened 2020-07-07
Corrects optional class properties in Python

thatsokay:
Stops serialising missing optional types from Python as `None`.

- Preserves `isOptional` property in type graph for Python
- Emits `None` check for optional types in Python `to_dict` method

Implements #1391 except no changes to Python's `supportsOptionalClassProperties`, `PythonRenderer.pythonType`, or `from_dict`.


dvdsgl:
It looks like this created a test failure in a non-Python language:https://buildkite.com/typeguard/quicktype-pr/builds/1403#ae6f2d7d-7fa0-4f61-b017-8f73402f7c5c/206-5260

thatsokay:
Fixed the JS prop types tests. Still looking at the Python tests.

thatsokay:
The Python tests are failing on `json/priority/combinations{1,2,3,4}.json` and `schema/direct-union.1.json` when there is a type of `{optional?: null}`. Without this PR, the Python serializer would always produce `{optional: null}` but with the PR it produces `{}`. Neither is really correct but this PR could break for anyone with such a type.

thatsokay:
Fixed tests

thatsokay:
@dvdsgl @schani would you be able to look at the diff test to check if it looks ok?

dvdsgl:
Diff looks good to me but @schani will have to review the changes.

maximeCony:
@schani I know this is an old PR but any chance you give it a look? :pray: 

Is there anything I can do to push this to the finish line?

maximeCony:
Thank you! 🙏

PR #1518, opened 2020-07-08
Added convert JSON option for JavaScript

thatsokay:
Added option to not `JSON.parse` and `JSON.stringify` when converting to/from JSON.

## Motivation

Using [PouchDB](https://pouchdb.com) which provides document updates as a JS object but I still want to use quicktype for type generation and (de)serializing dates.


dvdsgl:
I don't think the way the option is phrased makes much sense.

If quicktype is a bidirectional transformation `string <-> T` between JSON strings and your type `T`, this change allows you to instead create `any <-> T` where `any` is the dynamic data represented by the JSON string?

thatsokay:
A lot of failures with

```
The command '/bin/sh -c curl -sL "https://get.haskellstack.org/" | sh' returned a non-zero code: 1
```

schani:
Thank you!

PR #1521, opened 2020-07-09
C++: Implement support for large enums (fixes #1517)

### LINKED ISSUE #1517: C++ generator: large enums result in compiler error
For [enums](https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values), the C++ generator generates one `else if` branch for each enum field.

I.e. the following schema
```
{
  "type": "string",
  "enum": ["red", "amber", "green"]
}
```
compiles to
```
inline void from_json(const json & j, color & x) {
        if (j == "red") x = color::red;
        else if (j == "green") x = color::green;
        else if (j == "blue") x = color::blue;
        // ...
}
```

This behaviour emits tons of branches for large enums and in case of MSVC it leads to a compiler error with more than 128 fields. GCC and clang probably have higher limits, but I assume they break at some point too.
```
fatal error C1061: compiler limit: blocks nested too deeply
```
Also, checking it this way is pretty slow.

As an alternative, I'd suggest using a lookup table (e.g. hashmap) at least for larger enums.

Jusonex:
Currently, the C++ generator generates one else-if branch for each enum
field. This causes troubles with compilers. MSVC for example limits the
number of branches to 128.

Also, conditionally checking enum fields step-by-step is at some point
slower than looking it up in a hashmap.

This PR fixes issue #1517.

### DIFF
--- a/src/quicktype-core/language/CPlusPlus.ts
+++ b/src/quicktype-core/language/CPlusPlus.ts
@@ -1,4 +1,4 @@
-﻿import {
+import {
     setUnion,
     arrayIntercalate,
     toReadonlyArray,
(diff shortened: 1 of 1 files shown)


schani:
Thank you!

PR #1523, opened 2020-07-14
setting nil value for scalar keys should not raise an exception in obj-c

davidhodge:
Quicktype object initialization dictionaries may eventually receive null values in places where the user intended to send a scalar value. This change makes that non-fatal in objective-c.

At parse-time, unexpected null values for scalar types will raise exceptions and effectively throw out entire object graphs due to the default behavior of NSObject. Specifically, passing in null (NSNull) values for a scalar key in setValue:forKey: in objective-c triggers the default implementation of setNilValueForKey:, which raises an exception. This lives on a primary initialization path for Quicktype objects. Meanwhile, if the key/value were simply not included, no exception would be thrown even though that is semantically pretty close to passing in null.

The change does the following:

1. Overrides setNilValueForKey: to explicitly handle this case.
2. An exception will not be thrown.
3. The value is then explicitly set to 0.

Setting the value to 0 here matches general objective-c behavior of having un-set scalars be 0. Technically we could no-op this and the value would still be zero at init, but we want to correctly handle a case where the parent is mutated after creation. Another option here would be to simply not alter the value, but putting in 0 *feels* more right to me. Would be happy to change the behavior to a no-op as the outcome for most use cases is the same.

### DIFF
--- a/src/quicktype-core/language/Objective-C.ts
+++ b/src/quicktype-core/language/Objective-C.ts
@@ -744,9 +744,9 @@ export class ObjectiveCRenderer extends ConvenienceRenderer {
                 this.emitLine("if (resolved) [super setValue:value forKey:resolved];");
             });
 
-            // setNilForKey: is automatically invoked by the NSObject setValue:forKey: when it is passed nil for a scalar (a.k.a. non-nullable) object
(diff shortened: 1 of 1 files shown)


schani:
Thank you!

davidhodge:
You're welcome, sir!

Issue #1528, opened 2020-07-20
Fields of Enum type not respecting optionality when generating swift from graphql


pagr:
Members of type Enum are never generated as optionals even when they are supposed to.

A simple example would be the following types
```
Movie {
  id: ID!
  name: String!
  accessibility: Accessibility
}
Enum Accessibility {
  Default
  AudioDescribed
  SignInterpreted
}
```
When generating code for the movie object, accessibility will not be nullable as it should.

schani:
Retested this on current master (662ad5b6) using a GraphQL introspection schema matching your Movie/Accessibility example, with `node dist/index.js --lang swift --graphql-schema schema.json query.graphql`. The nullable `accessibility` field is now correctly emitted as `let accessibility: Accessibility?`, while the non-null `id` and `name` fields stay non-optional. So enum optionality is respected as expected. This fix will be in the next release; if you can still reproduce the problem on a current build, please reopen or comment with a repro and we'll take another look.

PR #1529, opened 2020-07-20
Generate correct optionality of Enums  in swift from graphql

pagr:
Nullable enums are always made NON_NULLABLE in generated code. This PR fixes that.

Solves:
#1528


schani:
Thank you!

PR #1531, opened 2020-07-28
Improve ts and js type validation error message

Rigellute:
When using the generated `Convert` with Typescript (and Javascript) it is currently very hard to track down the point at which validation fails. 

For example, here is an error message thrown during a Convert

```
Error: Invalid value undefined for type 0
```

This error doesn't help find the error (especially in larger objects).

This PR attempts to improve validation errors by adding the failing object key to the error message. Which would produce an error like this

```
Error: Invalid value for key "block_index". Expected type 0 but got undefined
```


schani:
Superseded by #1534

PR #1546, opened 2020-08-20
C++: Improve compiler support

Jusonex:
It seems that some compilers can't implicitly cast the json value. Explicitly doing that by calling get<T> fixes the problem.

I reproduced the issue with the latest IAR C++ compiler for ARM.


schani:
Thank you!

PR #1553, opened 2020-09-02
Go: prepend struct field comment (support GoDoc)

cv:
This makes the Go generator prepend struct field comments, so instead of:

```go
type Bar struct {
  Foo string `json:"foo"` // this bar's foo
}
```

...we get:

```go
type Bar struct {
  // this bar's foo
  Foo string `json:"foo"`
}

```
...which [GoDoc](https://godoc.org/) supports and documents accordingly.


bruth:
Bump for this to get merged..

dvdsgl:
@cv thank you! There is a small build error.

dvdsgl:
Including the comments in the column formatting produces code that `gofmt` does not like:

CleanShot 2023-01-03 at 11 29 00@2x

cv:
@dvdsgl is the expectation that the code generated by the tool passes `gofmt` without change?

In some of our projects that involve Go code generation we've been successful so far with letting the generator make a bit of a mess, then let `gofmt` clean everything up later.

PR #1560, opened 2020-09-21
Fix for import reference module path

nareshjs:
Fixed the correct relative module path reference for import from quicktype-core


schani:
Thank you!

PR #1565, opened 2020-10-03
feat: export MultiFileRenderResult

grant:
Exports `MultiFileRenderResult` in top level. Otherwise I have to do this to get the TS interface:

```ts
import {MultiFileRenderResult} from '../node_modules/quicktype-core/dist/TargetLanguage'
```

Related: https://github.com/googleapis/google-cloudevents/issues/83


schani:
Thank you!

Issue #1563, opened 2020-09-29
optional field C++ invalid schema (null)


Ender-events:
## Issue
The function to_json for optional field generate invalid json from json schema

```json
{
  "id": "http://json-schema.org/geo",
  "$schema": "http://json-schema.org/draft-06/schema#",
  "description": "bug example",
  "type": "object",
  "properties": {
    "requiredField": {
      "type": "number"
    },
    "optionalField": {
      "type": "number"
    }
  },
  "required": ["requiredField"]
}
```
https://app.quicktype.io?share=LCT9IHwHrkYjvL2ChlA9

The json output is
```json
{"optionalField":null,"requiredField":42.0}
```
instead of
```json
{"requiredField":42.0}
```

https://godbolt.org/z/5d7Msh

That doesn't respect the schema
>    Invalid type. Expected Number but got Null.

### Actual
```cpp
    inline void to_json(json & j, const quicktype::Optional & x) {
        j = json::object();
        j["optionalField"] = x.optional_field;
        j["requiredField"] = x.required_field;
    }
```
### Expected
```cpp
    inline void to_json(json & j, const quicktype::Optional & x) {
        j = json::object();
        if (x.optional_field)
            j["optionalField"] = x.optional_field;
        j["requiredField"] = x.required_field;
    }
```

PR #1566, opened 2020-10-03
optional field unset not in the json with null

Ender-events:
Close: #1563 

## Previous output

```cpp
    inline void to_json(json & j, const quicktype::Test & x) {
        j = json::object();
        j["optionalField"] = x.optional_field;
        j["requiredField"] = x.required_field;
    }
```

## New output

```cpp
    inline void to_json(json & j, const quicktype::Test & x) {
        j = json::object();
        if (x.get_optional_field()) {
            j["optionalField"] = x.get_optional_field();
        }
        j["requiredField"] = x.get_required_field();
    }
```

### DIFF
--- a/src/quicktype-core/language/CPlusPlus.ts
+++ b/src/quicktype-core/language/CPlusPlus.ts
@@ -837,6 +837,14 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
 
     protected cppType(t: Type, ctx: TypeContext, withIssues: boolean, forceNarrowString: boolean, isOptional: boolean): Sourcelike {
         const inJsonNamespace = ctx.inJsonNamespace;
+        if (isOptional && t instanceof UnionType) { // avoid have optionalType<optionalType<Type>>
+            for (const tChild of t.getChildren()) {
+                if (tChild.isNullable) {
+                    isOptional = false;
+                    break;
+                }
+            }
+        }
         let typeSource = matchType<Sourcelike>(
             t,
             _anyType =>
--- a/src/quicktype-core/language/CPlusPlus.ts
+++ b/src/quicktype-core/language/CPlusPlus.ts
@@ -106,7 +106,8 @@ export const cPlusPlusOptions = {
         pascalUpperAcronymsValue,
         camelUpperAcronymsValue
     ]),
-    boost: new BooleanOption("boost", "Require a dependency on boost. Without boost, C++17 is required", true)
+    boost: new BooleanOption("boost", "Require a dependency on boost. Without boost, C++17 is required", true),
+    hideNullOptional: new BooleanOption("hide-null-optional", "Hide null value for optional field", false),
 };
 
 export class CPlusPlusTargetLanguage extends TargetLanguage {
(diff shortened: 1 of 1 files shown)


Ender-events:
Found `p.isOptional` (`p: ClassProperty`) but in `CPlusPlusRenderer` is always false
and in `SwiftRenderer` look like to have correct value (only true for optionalField)

schani:
@Ender-events Thank you for your contribution!

This fails a test.  Could you please take a look at it?

thatsokay:
@Ender-events the `supportsOptionalClassProperties` method on the `CPlusPlusTargetLanguage` returns `false` (inherited from [`TargetLanguage`](https://github.com/quicktype/quicktype/blob/aedfdd2f4976fc69c47fb52a5fe81c250ac890e3/src/quicktype-core/TargetLanguage.ts#L68)), which causes [`optionalToNullable`](https://github.com/quicktype/quicktype/blob/aedfdd2f4976fc69c47fb52a5fe81c250ac890e3/src/quicktype-core/Run.ts#L468) to be called on the type graph, which [sets all `p.isOptional` to `false`](https://github.com/quicktype/quicktype/blob/aedfdd2f4976fc69c47fb52a5fe81c250ac890e3/src/quicktype-core/TypeGraph.ts#L503).

`SwiftTargetLanguage` [overrides the `supportsOptionalClassProperties` method](https://github.com/quicktype/quicktype/blob/aedfdd2f4976fc69c47fb52a5fe81c250ac890e3/src/quicktype-core/language/Swift.ts#L149) to return `true` so preserves `p.isOptional` flags.

I got around this in [my PR](https://github.com/quicktype/quicktype/pull/1516/files) by modifying `optionalToNullable` to preserve `p.isOptional` flags (Doing this breaks the `JavaScriptPropTypes` test but the fix is included in the PR).

pvaibhav:
Any estimate on when this can be merged?
This is *crucial* if we want to be able to save back the loaded json. At the moment this causes schema validation failure as soon as the json is regenerated.

Thanks.

Ender-events:
I had a new rendered option `hide-null-optional`

Is it possible to run a test with an additional rendered option?

Something like
```json
{
    "stuff": {
        "only-required": {
            "required": true
        },
        "optional": {
            "required": true,
            "optional": 1
        }
    }
}
```
> direct-union.1.hide-null-optional.json

```json
{
    "stuff": {
        "fail": {
            "required": true,
            "optional": null
        }
    }
}
```
> direct-union.1.fail.hide-null-optional.json

schani:
I don't think we have allowances for tests specific to certain options.  If you do want to test that you'd have to add that capability to our test suite.  I'm fine merging it without.

schani:
@Ender-events This introduces some changes in order and naming, but that's nothing to worry about.  [Here's the output diff](https://s3.amazonaws.com/quicktype-outputs/diffs/02b1e479ce37f28e782cc88d2f2144bea66e924c-6c1f589e8b67309af539093200ad72d7b56d290c.diff) in case you're curious.

Do you want us to merge this, or is there still stuff you'd like to add/change?

Ender-events:
You can merge

schani:
Thank you so much!

Issue #1026, opened 2018-08-10
C++: multi-source files + generators


harsszegi:
Hello Mark,
just realized a few things:
- if single-source is generated and multiple (more than one) generated JSON structures are used in the application, compilation will fail, as - because of the single source generation logic - the same common (""helper"") functions will be defined multiple times in the same namespace.

- if multi-source is generated and multiple (more than one) generated JSON structures are used, the application will not compile. the background reason is that the testing framework expects one single file (Toplevel.hpp + Generators.hpp) so currently the C++ lang generates Generators.hpp for every requests (e.g. instead of Foo_generator.hpp, Pokedex_generator.hpp, etc.) Obviously I can fix that, just that have zero idea how to fix that in the testing framework.

schani:
I see.  To me the first point is much more problematic, since that's quicktype's default mode of operation.

As to the testing: unfortunately our test suite doesn't currently test multiple top-levels, or it would have caught those issues.  I guess that's my responsibility.

PS: If you do fix those, please make a PR for one of them first, and only when we've merged that, for the other one ;-)

flounder5:
I believe that I'm running into this same issue.  I only have one single-source generated file in my project, but I'm getting "mulitple definition" linking errors on the two `void CheckConstraint(...)` methods.  Making them inline seems to fix the problem (similar to the `inline json get_untyped(...)` methods).  

In this file, would changing "void" to "inline void" on lines 1771 and 1828 fix the issue?  If so, I'm happy to attempt a pull request.

schani:
I retested this on current master (662ad5b6). The single-source multiple-definition problem flounderpinto ran into is fixed: all the shared C++ free functions (get_untyped, get_heap/stack_optional, from_json/to_json, and the six CheckConstraint overloads) are now emitted as inline, so a single-source file with multiple top-levels no longer produces multiple-definition link errors. Confirmed with `node dist/index.js -l c++ --src srcdir -o Both.hpp` and `node dist/index.js -l c++ --src-lang schema a.schema -o WithConstraints.hpp` against a schema with constraints.

This will be included in the next release. If you're still seeing multiple-definition or link failures on a current build, please reopen or comment with a repro and I'll take another look.

PR #1612, opened 2020-12-18
C++: inline for CheckConstraint functions to allow mutliply header in…

satinowl:
C++: inline for CheckConstraint functions to allow mutliply header includes
partially relates to #1026


schani:
Thank you!

PR #1617, opened 2020-12-28
Fix programmatic example in README ("Calling `quicktype` from JavaScript")

### LINKED ISSUE #1585: Programmatic example in README instanciates an abstract class
In the readme, the programmatic example shows the following line:

```typescript
  const schemaInput = new JSONSchemaInput(new JSONSchemaStore());
```

This doesn't work in TypeScript because `JSONSchemaStore` is [marked as abstract](https://github.com/quicktype/quicktype/blob/d4a8e68abdf09b403ba02ed9b2de2325ceac5fe1/src/quicktype-core/input/JSONSchemaStore.ts#L5). I guess the simplest fix is to unmark `JSONSchemaStore` as abstract.

ineffyble:
Current example is broken because JSONSchemaStore is an abstract class and cannot be instantiated.

Fixes #1585.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ const {
   InputData,
   jsonInputForTargetLanguage,
   JSONSchemaInput,
-  JSONSchemaStore,
+  FetchingJSONSchemaStore,
 } = require("quicktype-core");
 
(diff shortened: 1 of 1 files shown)


schani:
Thank you!

PR #1636, opened 2021-01-20
Make Kotlin base emit functions protected so it can be overwritten

tiaanl:
It is very restricting when creating a new variant of the Kotlin target language when the main emit functions are private and can not be overwritten any more.


schani:
Thank you! :-)

PR #1648, opened 2021-02-09
added dart related export at index.ts

ondbyte:
dart support is much needed for quick type as flutter is on the rise


schani:
Thank you!

PR #1720, opened 2021-05-13
Update Kotlin (kotlinx) usage API sample code

pedromassango:
The old code is now deprecated and it is an error to use it.

Screen Shot 2021-05-13 at 09 25 20


pedromassango:
Cc @schani  for review and merge this.

Issue #1673, opened 2021-03-07
Add Dart non-nullable support


jarekb123:
As Dart non-nullable is stable now, please add option to generate non-nullable classes with the tool (especially freezed classes)

xamantra:
**+1** we really need this feature.

chrislaurie-bs:
It looks to me like that all that is needed for null safety is to make it work exactly like make all fields required, but to use just 'required' as the filed prefix, instead of '@required'. It should probably be under its own switch so that you keep the old functionality.

jarekb123:
Also the tool should detect if some field may be nullable (eg. If some field has null value in provided json).

xamantra:
Think about `copyWith(...)`. The dart migration tool treats the copyWith parameters as non-nullable. I had to manually put question marks to each of the parameters _(sorry I'm bad at regex)_. All parameters in the copyWith method should be nullable to properly implement the `prop: prop ?? this.prop` expression.

It shouldn't even be an option. Quicktype must generate `copyWith` with all parameters as nullable. With `?`

![image](https://user-images.githubusercontent.com/37391380/110780674-e945aa80-829f-11eb-8a2e-80c01ecd6e11.png)

Imagine a `copyWith` method with all parameters required (non-nullable). That would be a disaster.

chrislaurie-bs:
This is what freezed null-safely generates:

class _$ShopCopyWithImpl<$Res> implements $ShopCopyWith<$Res> {
  _$ShopCopyWithImpl(this._value, this._then);

  final Shop _value;
  // ignore: unused_field
  final $Res Function(Shop) _then;

  @override
  $Res call({
    Object? shopCode = freezed,
    Object? shopName = freezed,
  }) {
    return _then(_value.copyWith(
      shopCode: shopCode == freezed
          ? _value.shopCode
          : shopCode // ignore: cast_nullable_to_non_nullable
              as String,
      shopName: shopName == freezed
          ? _value.shopName
          : shopName // ignore: cast_nullable_to_non_nullable
              as String,	
    ));
  }
}

xamantra:
**_code generation_**.. stopped using since 2019. especially with localizations. I just personally hate it. the explanation would be too long. It's based on experience.

TheWCKD:
+1 We really need to generate **dart non-nullable classes** with quicktype!

xang555:
+1

nimi0112:
+1

RLZW:
+1

[...]

Issue #1739, opened 2021-06-15
Dart Json Model(About Null Safety) 


ulvinasirli99:
Dart language has switched to null safety in version 2.0 and the site no longer supports it. If possible, switch to null safety to model the dart.

NatoBoram:
Duplicate of #1673

Issue #1737, opened 2021-06-12
Dart null safety


tmtong:
There is this new feature in dart that requires dart code to be null safety?
Any plans to support that?

NatoBoram:
Duplicate of #1673

Issue #1729, opened 2021-06-01
Dart Flutter Null Safety


HabibRasyid:
Dear app.quicktype,

can app.quicktype support null safety on dart-flutter?

regards,
Okid

sharpred:
duplicate of #1673

PR #1740, opened 2021-06-15
Dart null safty and fix for freeze

stha-ums:
Following are the things I worked on:

- added detect enum option 
- added null-safety
- make all properties nullable 
- keep properties name, enum, and union name same as Key in JSON (this is useful while generated for Freezed with JsonSerializer) .

I have done manual testing. So it should be working. For automatic testing. I do not have enough knowledge. Is anyone can test or guide me. That would be great.


NatoBoram:
Fixes #1673 #1711 #1723 #1729 #1737 #1739 #1741 #1744

schani:
@stha-ums Thank you for this contribution!  There's a lot of reformatting of unrelated files in this PR, as well as huge changes to `package-lock.json`.  It also breaks the Elm tests - why does it touch Elm at all?

NatoBoram:
@schani Looks like someone on your team misused the `package.json` file by manually entering/ordering dependencies instead of using `npm`. The dependencies in `package.json` are generally in the same order to avoid this kind of formatting issue.

As for `elm-package.json`, is it an auto-generated file? This project includes `prettier`. If someone uses this project's settings, it is reasonable that prettier would be applied.

stha-ums:
@schani  there was an error while building that project with elm versioning so I had to change the version to run on my machine. I have relatively new to Node, what do you suggest for the resolution?

NatoBoram:
@stha-ums Can you edit this PR to keep their original config, even if it doesn't build on your machine? It's clear they don't have much time to give to this project, so if it works on their machine I think it's good enough.

stha-ums:
sent another PR #1764 .

Issue #1755, opened 2021-07-14
Make All properties required ( @ is no longer required in Dart latest version)


Adil-Rangila:
in Updated dart version symbol  @  is no longer required.

NatoBoram:
Duplicate of #1673

RSandhir:
#1888

PR #1761, opened 2021-07-18
Got rid of @ which is not used in current version

jdebecka:
```
class Dart {

const Dart({required this.name});

final String name;
}
```


NatoBoram:
Closes #1755, #1790, #1818, #1826, #1857

NatoBoram (approved):
Closes #1755, #1790, #1818, #1826, #1857

dvdsgl:
Fixed in v16

PR #1789, opened 2021-09-01
TypeScript renderer: Option to render uninons insted of enums

zewish:
This PR adds an extra option to the render options of the TypeScript language (called `prefer-unions`).
When the option is enabled some easier to use union types are generated instead of enums.

When the option is disabled we get still get the old result:
```typescript
enum Language {
    En: 'en',
    De: 'de'
}
```

But when the option is enabled we get this instead:
```typescript
type Language = 'en' | 'de';
```

P. S.: I'm open to any suggestions about code improvements. 😉

### DIFF
--- a/src/quicktype-core/language/TypeScriptFlow.ts
+++ b/src/quicktype-core/language/TypeScriptFlow.ts
@@ -20,7 +20,8 @@ import { isES3IdentifierStart } from "./JavaScriptUnicodeMaps";
 export const tsFlowOptions = Object.assign({}, javaScriptOptions, {
     justTypes: new BooleanOption("just-types", "Interfaces only", false),
     nicePropertyNames: new BooleanOption("nice-property-names", "Transform property names to be JavaScripty", false),
-    declareUnions: new BooleanOption("explicit-unions", "Explicitly name unions", false)
+    declareUnions: new BooleanOption("explicit-unions", "Explicitly name unions", false),
+    preferUnions: new BooleanOption("prefer-unions", "Use union type instead of enum", false)
 });
 
 const tsFlowTypeAnnotations = {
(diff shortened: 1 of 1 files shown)


zewish:
Hey @schani, I know you guys are probably quite busy with projects that you get paid for and you most probably don't have any free time for quicktype right now, but please spare a minute and read this:

1. The JavaScript/Flow/TypeScript tests are passing locally without an issue, but they seem to be broken on your CI environment.
2. We're in need for this feature to be merged for an open source project we've been using quicktype as dependency for quite some time (several years now). In case quicktype is dead and will no longer be supported, please inform us so we can start searching for alternatives.

zewish:
Unfortunately we couldn't find a way to get this moved ahead, @daniel-ro-github - to me quicktype seems abandoned at this point. We're currently planing a way to rid of quicktype in our project and replace it with another type-generating tool which doesn't have this issue.

daniel-ro-github:
> Unfortunately we couldn't find a way to get this moved ahead, @daniel-ro-github - to me quicktype seems abandoned at this point. We're currently planing a way to rid of quicktype in our project and replace it with another type-generating tool which doesn't have this issue.

@zewish - thanks for your prompt reply! did you consider branching the quicktype master and publishing a new npm package with your change?

zewish:
@daniel-ro-github, We wouldn't want to support a huge beast as big as quicktype (i.e. too many languages being generated here), since it does way more than what we need. I'm planning to release a TypeScript-only solution which does exactly what's need for our project some time soon. Will keep you updated once it's out, and hopefully it helps your use-case also.

schani:
@zewish  My apologies.

Could you please rebase this on the current `master` and remove the change to the constructor and I'll merge it.

dvdsgl:
Merged in #1908

Issue #1021, opened 2018-08-09
Option to use the JSON property names as the member names


schani:
A lot of people are requesting this, with a variety of different slightly weird reasons.

hi2u:
@schani Is there still no way to disable the automatic renaming of properties at all?  Or am I missing something?

I've spent many hours looking into this, and there's heaps of other issues asking for it... is there at least some dodgy hack I can do to disable the renaming of properties?  

Right now I'm:

* Using quicktype-core in my Node app
* passing in sample JSON files
* outputting Rust structs

Here's my code so far:

```typescript
const jsonInput = jsonInputForTargetLanguage('rust');
await jsonInput.addSource({
    name: 'MainStructName',
    samples: json_samples,
});

const inputData = new InputData();
inputData.addInput(jsonInput);

const result = await quicktype({
    inputData,
    lang: 'rust',
    inferDateTimes: false,
    alphabetizeProperties: true,
    inferEnums: false,
    rendererOptions: {},
});
```

Although I'm keen to figure this out for any output language, not just Rust.

Looking through some of the source code, I can see stuff such as:

    protected namerForObjectProperty(): Namer | null;

...the `| null` implies it can be disabled?  And the fact that it's a property implies that these are pluggable and maybe I can supply my own naming function?

But I can't figure out how to actually do it.  

Is it possible at all?

ulquiomaru:
Most other similar apps have a toggle for PascalCasing. Being able to disable the renaming completely or toggle it would be awesome.

schani:
We do not take feature requests anymore. Please consider making a pull request instead. We encourage the use of coding agents, but please use at least an Opus-class model.

PR #1820, opened 2021-11-03
Add keep-naming csharp option

FreshlyBrewedCode:
A quick workaround for #1021 in C#. This PR adds the `--keep-naming` C# option.

I do not have the time to properly test or verify this solution, however, it works for my use case. 

Here is an example of using the option:
```typescript 
// test.ts
interface Test {
    someProp: string;
    anotherProp: number;
}
```
Output of `quicktype test.ts --out test.cs`:
```c#
// test.cs
public partial class Test
{
    [JsonProperty("anotherProp")]
    public double AnotherProp { get; set; }

    [JsonProperty("someProp")]
    public string SomeProp { get; set; }
}
```
Output of `quicktype test.ts --out test.cs --keep-naming`:
```c#
// test.cs
// [...]
public partial class Test
{
    [JsonProperty("anotherProp")]
    public double anotherProp { get; set; }

    [JsonProperty("someProp")]
    public string someProp { get; set; }
}
// [...]
```


dvdsgl (commented):
Test new --keep-naming option

dvdsgl:
@FreshlyBrewedCode we finally fixed our CI! It looks like adding this option to the test suite produces some invalid C# files. Since it's been quite a while since you authored this PR, I will close it, but please reopen if you'd like to fix the errors.

PR #1824, opened 2021-11-09
Option to emit string union instead of enum

daniel-ro-github:
In emitEnum on line 310, we'd like this to output string union instead of enum type. Is there an option which supports this?
Could we create a pr to update this behavior?

Thanks in advance!

cc @schani, @mamantoha , @jnpdx , @uhop , @proppy


talks2much:
Nobody.

clystian:
@daniel-ro-github 

I think it isn't not recommended to create a commit with a suggested feature

> In emitEnum on line 310, we'd like this to output string union instead of enum type. Is there an option which supports this?

No, string union type for typescript enum structure isn't supported

> Could we create a pr to update this behavior?

Yeah, you can create code and PR with implemented behavior you want, that will be tested by CI inicially

zardoy:
@clystian

> No, string union type for typescript enum structure isn't supported

But why? It is possible, I'm using a workaround to get them even inline. Inline enums look much more better. Always wanted to have this!: https://github.com/zardoy/vscode-framework/blob/3ac2146c53831fccfe20f9f6f7a85c2e17153d96/integration-tests/fixture-snapshots/vscode-gitlens/src-generated-bare.ts#L464-L468

btw I'm surprised how well my workaround works.

Beat-YT:
#1824 1789

dvdsgl:
Great idea! Please try to implement it but this PR only contains comments.

PR #1832, opened 2021-11-21
Fixing the required keyword syntax

romme86:
Just removed the `@required` annotation for the correct `required` keyword


schani:
@romme86 Which Dart language version is required for this change to work?

Could you please rebase on/merge the current `master`?

PR #1838, opened 2021-11-28
golang: enable toDict/fromDict like python

mabels:
I enhance the golang implemtation to enable an abstract type handling.
Abstract means the json serializer is not the central part of binding between the serialized value and the internal golang type representation.
This enables to use quicktype to use for example with a protobuf implemtation. Or to use quicktype nested types without multiple serialization and deserialization.


schani:
@mabels Apologies!  Could you please rebase/merge the latest `master` so that CI can run, and remote the changes to `package.json` and `package-lock.json`?  I don't think those are necessary.

mabels:
I just rebaseed it again.
I added the vercel ncc tool to build a one jsfile quicktype cli. Which i used during development.
I won't like to remove it makes the usage and debugging much easier.

thx

schani:
If you'd like to add ncc and change the Dockerfile, please make separate PRs for those.  As it is, your changes to both make CI fail - `elm-make` isn't available anymore because you updated elm, and I don't know what the Swift change will do.

Also, your changes make Go tests fail: https://buildkite.com/typeguard/quicktype-pr/builds/1617#484a736e-1075-4fe3-ae9b-8390615af370

mabels:
I will have a reorganize my changes to enable a better intergration.

meno

Issue #1852, opened 2021-12-15
[ruby] global namespace collisions


jamesarosen:
### Goal
Ruby gems, when loaded, put all declared objects into the global namespace. In order to prevent collisions with things like `::Date` and `::Time`, I'd like to put the generated classes into a namespace.

### Namespace Flag
One way to solve this would be to pass a `--namespace` to the Quicktype CLI:
```sh
quicktype -o src/my/namespace/index.rb --namespace "My::Namespace"
```
which would generate
```
require 'json'
require 'dry-types'
require 'dry-struct'

module Types
  include Dry::Types.module

  Hash   = Strict::Hash
  Double = Strict::Float | Strict::Int
end

module My
  module Namespace
    class Date < Dry::Struct
      attribute :year,  Types::Double.optional
      # …
    end
  end
end
```

### Templates
Another way to solve this would be to use the now-removed [template support](https://github.com/quicktype/quicktype/blob/29f459dc95fe887d9e03cc19d65bc9908364b11a/doc/Templates.md). This would let users define their own wrapping modules as necessary.

schani:
Thanks for the clear request. Ruby output now supports this through `--namespace`; for example, `quicktype --lang ruby --namespace 'My::Namespace' ...` wraps the generated `Types` module and model classes in nested `My` and `Namespace` modules. Chained `::` namespaces are supported, and the resulting output is valid Ruby. This issue can be closed.

PR #1853, opened 2021-12-15
Allow namespaces to be specified by `namespace` flag.

renatodex:
# Allow namespaces to be specified by `namespace` flag.
Well, this is my first time working in this repo, so I've probably have missed a lot of stuff.
The goal of this PR is to implement Namespaces in Ruby, so we can't prevent **global class name collision** in projects. (see #1852)

For context, by default, when you require classes in Ruby, it automatically throw everything in the Global scope, so if you are trying to export Quicktype objects for a Gem like we are doing, you will eventually find problems with name collisions.

So this PR introduces an `emitModule` method, which adds a simple Ruby block and accept a function scope, so all the rest can be appended.

Also, the `rubyOptions.namespace` option was added, to allow the Users to specify a string namespace, as in:

```npm start -- "--lang ruby pokedex.json --namespace MyCustomNamespace"```

Please give me a feedback if those changes makes sense for you all!

Note: I'm still trying to figure out why my editor is not letting lines 435 and 437 to be indented in the original way, it's automatically removing 2 spaces.

### CODE REVIEW

File: src/quicktype-core/language/ruby/index.ts [hunk 1]
```diff
@@ -629,13 +637,28 @@ export class RubyRenderer extends ConvenienceRenderer {
...
+                }
             }
         });
 
         this.forEachNamedType(
             "leading-and-interposing",
-            (c: ClassType, n: Name) => this.emitClass(c, n),
+            (c: ClassType, n: Name) => {
+                if (this._options.namespace) {
+                    this.emitModule(this._options.namespace, () => {
```
jamesarosen, line 655: Given `_options.namespace = 'Foo::Bar'`, this will emit
```rb
module Foo::Bar
  class Baz
  end
end
```
That will fail if `Foo` doesn't exist beforehand.

renatodex: Oh yes, we could use some validations here. What you think it's the best way to validate this?
Stripping submodules would be the most easy path here, after all, I'm not sure how relevant is to use Submodules with quickup output files.
What do you think?

jamesarosen: I wouldn't silently strip them. I'd either (1) split on `::` and nest the modules or (2) raise an exception if `namespace` has `::` anywhere other than the beginning of the string.

renatodex: Hmmmm splitting in nested modules is a great idea!

renatodex:
Hey @jamesarosen ,

I've added support for multiple submodules!
So running something like this in your local env:

```bash
npm start -- "--namespace MyNamespace::WithAChainOf::Multiple::Submodules --lang ruby pokedex.json"
```

Should generate a code like this:

```ruby
module MyNamespace
  module WithAChainOf
    module Multiple
      module Submodules
        class Pokedex
          def self.from_json!(json)
            pokedex = JSON.parse(json, quirks_mode: true).map { |x| PokedexElement.from_dynamic!(x) }
            pokedex.define_singleton_method(:to_json) do
              JSON.generate(self.map { |x| x.to_dynamic })
            end
            pokedex
          end
        end
      end
    end
  end
end
```

renatodex:
Hey @schani  !
What do we need to get this approved and released?

Do **quicktype** has some kind of process for incorporating new changes into new releases?

Thank you for all the hard work!

sarkstephan (approved):

schani:
My apologies!

@renatodex Could you please rebase/merge the latest `master` so that CI can run?

dvdsgl:
@renatodex I have this new option running in our test suite now.

For `pokedex.json`, here is the output. Is it correct for the `Types` and `Egg` modules not to be namespaced?

```ruby
# This code may look unusually verbose for Ruby (and it is), but
# it performs some subtle and complex validation of JSON data.
#
# To parse this JSON, add 'dry-struct' and 'dry-types' gems, then do:
#
#   top_level = TopLevel.from_json! "{…}"
#   puts top_level.pokemon.first.weaknesses.first == Type::Bug
#
# If from_json! succeeds, the value returned matches the schema.

require 'json'
require 'dry-types'
require 'dry-struct'

module Types
  include Dry.Types(default: :nominal)

  Integer = Strict::Integer
  Nil     = Strict::Nil
  Hash    = Strict::Hash
  String  = Strict::String
  Double  = Strict::Float | Strict::Integer
  Egg     = Strict::String.enum("Not in Eggs", "Omanyte Candy", "10 km", "2 km", "5 km")
  Type    = Strict::String.enum("Bug", "Dark", "Dragon", "Electric", "Fairy", "Fighting", "Fire", "Flying", "Ghost", "Grass", "Ground", "Ice", "Normal", "Poison", "Psychic", "Rock", "Steel", "Water")
end

module Egg
  NotInEggs    = "Not in Eggs"
  OmanyteCandy = "Omanyte Candy"
  The10KM      = "10 km"
  The2KM       = "2 km"
  The5KM       = "5 km"
end

module QuickType
  class Evolution < Dry::Struct
    attribute :evolution_name, Types::String
    attribute :num,            Types::String

    def self.from_dynamic!(d)
      d = Types::Hash[d]
      new(
        evolution_name: d.fetch("name"),
        num:            d.fetch("num"),
      )
    end

    def self.from_json!(json)
      from_dynamic!(JSON.parse(json))
    end

    def to_dynamic
      {
        "name" => evolution_name,
        "num"  => num,
      }
    end

    def to_json(options = nil)
      JSON.generate(to_dynamic, options)
    end
  end
end

module Type
  Bug      = "Bug"
  Dark     = "Dark"
  Dragon   = "Dragon"
  Electric = "Electric"
  Fairy    = "Fairy"
  Fighting = "Fighting"
  Fire     = "Fire"
  Flying   = "Flying"
  Ghost    = "Ghost"
  Grass    = "Grass"
  Ground   = "Ground"
  Ice      = "Ice"
  Normal   = "Normal"
  Poison   = "Poison"
  Psychic  = "Psychic"
  Rock     = "Rock"
  Steel    = "Steel"
  Water    = "Water"
end

module QuickType
  class Pokemon < Dry::Struct
    attribute :avg_spawns,     Types::Double
    attribute :candy,          Types::String
    attribute :candy_count,    Types::Integer.optional
    attribute :egg,            Types::Egg
    attribute :height,         Types::String
    attribute :id,             Types::Integer
    attribute :img,            Types::String
    attribute :multipliers,    Types.Array(Types::Double).optional
    attribute :next_evolution, Types.Array(Evolution).optional
    attribute :num,            Types::String
    attribute :pokemon_name,   Types::String
    attribute :pokemon_type,   Types.Array(Types::Type)
    attribute :prev_evolution, Types.Array(Evolution).optional
    attribute :spawn_chance,   Types::Double
    attribute :spawn_time,     Types::String
    attribute :weaknesses,     Types.Array(Types::Type)
    attribute :weight,         Types::String

    def self.from_dynamic!(d)
      d = Types::Hash[d]
      new(
        avg_spawns:     d.fetch("avg_spawns"),
        candy:          d.fetch("candy"),
        candy_count:    d["candy_count"],
        egg:            d.fetch("egg"),
        height:         d.fetch("height"),
        id:             d.fetch("id"),
        img:            d.fetch("img"),
        multipliers:    d.fetch("multipliers"),
        next_evolution: d["next_evolution"]&.map { |x| Evolution.from_dynamic!(x) },
        num:            d.fetch("num"),
        pokemon_name:   d.fetch("name"),
        pokemon_type:   d.fetch("type"),
        prev_evolution: d["prev_evolution"]&.map { |x| Evolution.from_dynamic!(x) },
        spawn_chance:   d.fetch("spawn_chance"),
        spawn_time:     d.fetch("spawn_time"),
        weaknesses:     d.fetch("weaknesses"),
        weight:         d.fetch("weight"),
      )
    end

    def self.from_json!(json)
      from_dynamic!(JSON.parse(json))
    end

    def to_dynamic
      {
        "avg_spawns"     => avg_spawns,
        "candy"          => candy,
        "candy_count"    => candy_count,
        "egg"            => egg,
        "height"         => height,
        "id"             => id,
        "img"            => img,
        "multipliers"    => multipliers,
        "next_evolution" => next_evolution&.map { |x| x.to_dynamic },
        "num"            => num,
        "name"           => pokemon_name,
        "type"           => pokemon_type,
        "prev_evolution" => prev_evolution&.map { |x| x.to_dynamic },
        "spawn_chance"   => spawn_chance,
        "spawn_time"     => spawn_time,
        "weaknesses"     => weaknesses,
        "weight"         => weight,
      }
    end

    def to_json(options = nil)
      JSON.generate(to_dynamic, options)
    end
  end
end

module QuickType
  class TopLevel < Dry::Struct
    attribute :pokemon, Types.Array(Pokemon)

    def self.from_dynamic!(d)
      d = Types::Hash[d]
      new(
        pokemon: d.fetch("pokemon").map { |x| Pokemon.from_dynamic!(x) },
      )
    end

    def self.from_json!(json)
      from_dynamic!(JSON.parse(json))
    end

    def to_dynamic
      {
        "pokemon" => pokemon.map { |x| x.to_dynamic },
      }
    end

    def to_json(options = nil)
      JSON.generate(to_dynamic, options)
    end
  end
end
```

PR #1893, opened 2022-02-23
Fix an issue that some Rust options were not available.

nazo6:
#1496 added a new option to support Rust 2018 Edition, but it doesn't seem to be published successfully. This PR fixes it.


john-michaelburke:
+1 Any chance an owner could look into this? @schani

PR #1917, opened 2022-03-29
Add gson serialization for kotlin

matsjoe:
Adds gson as framework for kotlin.


matsjoe:
Nvm build keeps failing on existing code. I hoped to add some functionality but that is quite annoying this way

PR #1904, opened 2022-03-11
Dart null-safety and other improvements

RmondJone:
1、适配Dart空安全代码生成不对的问题
2、删除Dart生成冗余代码
3、修改Dart字段注释不对的问题
4、适配required-props属性生成代码@required不对的问题

1、The problem of incorrect generation of Dart empty security code
2、 Delete Dart to generate redundant code
3、Modify the problem of incorrect Dart field annotations
4、The problem of adapting the required-props property to generate code @required is incorrect

```
// To parse this JSON data, do
//
//     final test = testFromJson(jsonString);

import 'dart:convert';

Test? testFromJson(String str) => Test.fromJson(json.decode(str));

String testToJson(Test? data) => json.encode(data!.toJson());

class Test {
    Test({
        this.code,
        this.info,
        this.msg,
        this.traceId,
    });

    
    ///响应码
    double? code;
    
    ///结果数据
    Info? info;
    String? msg;
    String? traceId;

    factory Test.fromJson(Map<String, dynamic> json) => Test(
        code: json["code"].toDouble(),
        info: json["info"],
        msg: json["msg"],
        traceId: json["traceId"],
    );

    Map<String, dynamic> toJson() => {
        "code": code,
        "info": info,
        "msg": msg,
        "traceId": traceId,
    };
}


///结果数据
class Info {
    Info({
        this.baseDetail,
        this.contractActivityInfos,
        this.productActivityInfos,
        this.productParam,
        this.serviceInfo,
        this.shoppingActivityInfos,
    });

    
    ///商品详情页基础详情（图片、视频、展示价格、普通价、名称、描述、是否收藏、是否在活动）
    BaseDetail? baseDetail;
    
    ///合约通详情
    List<ContractActivityInfo?>? contractActivityInfos;
    
    ///商品参与的活动的信息，仅商品参与了活动时有值
    List<ProductActivityInfo?>? productActivityInfos;
    
    ///产品参数
    ProductParam? productParam;
    
    ///商品服务信息
    ServiceInfo? serviceInfo;
    
    ///加购浮层、购物车切换活动的活动信息列表
    List<ShoppingActivityInfo?>? shoppingActivityInfos;

    factory Info.fromJson(Map<String, dynamic> json) => Info(
        baseDetail: json["baseDetail"],
        contractActivityInfos: json["contractActivityInfos"] == null ? [] : List<ContractActivityInfo?>.from(json["contractActivityInfos"]!.map((x) => ContractActivityInfo.fromJson(x))),
[...]

PR #1920, opened 2022-04-03
Removed "@" notation for required fields in dart

### LINKED ISSUE #1905: no need to use the "@" notation in modern dart
just use the keyword "required" instead of "@required"

marvelefe:
# Whats inside
**Removed "@" notation for required fields in dart**

`@required` is just an annotation that allows analysers let you know that you're missing a named parameter and that's it. so you can still compile the application and possibly get an exception if this named param was not passed.

However sound null-safety was added to dart, and required is now a keyword that needs to be passed to a named parameter so that it doesn't let the compiler run if this parameter has not been passed. It makes your code more strict and safe.

https://dart.dev/null-safety/faq#how-does-required-compare-to-the-new-required-keyword

Also closes #1905

### DIFF
--- a/src/quicktype-core/language/Dart.ts
+++ b/src/quicktype-core/language/Dart.ts
@@ -496,7 +496,7 @@ export class DartRenderer extends ConvenienceRenderer {
                 this.emitLine(className, "({");
                 this.indent(() => {
                     this.forEachClassProperty(c, "none", (name, _, _p) => {
-                        this.emitLine(this._options.requiredProperties ? "@required " : "", "this.", name, ",");
+                        this.emitLine(this._options.requiredProperties ? "required " : "", "this.", name, ",");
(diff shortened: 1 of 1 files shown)


RmondJone:
关于Dart空安全问题，不仅仅是要解决这个@的问题，还有其他一堆问题需要解决。我早先已经提交了一个PR给官方了。具体参考：[#1904](https://github.com/quicktype/quicktype/pull/1904)

Regarding the Dart null safety issue, it's not just this @ issue that needs to be addressed, there are a bunch of other issues that need to be addressed. I have submitted a PR to the official earlier. specific reference: [#1904](https://github.com/quicktype/quicktype/pull/1904)

dvdsgl:
I've merged an equivalent fix. Thank you!

PR #1941, opened 2022-05-07
fix: #1592 c++ type re-definition errors

### LINKED ISSUE #1592: Multiple json schemas cause type re-definition errors for the boilerplate types like
Hello,

I am trying to use quicktype to auto-generate C++ code for several JSON Schema documents.  I have tried a few different combinations of parameters (e.g. multi-source, output path) but cannot get headers that contain all the to/from_json() definitions (Generators.hpp gets clobbered with whatever the last JSON Schema document is) AND don't have name collisions for redefinitions of helper functions like get_untyped() (helper.hpp gets replicated if separate output directories are used for each JSON schema doc).

Is quicktype designed with the expectation that only a single JSON Schema document would be used for a compilation unit?

If required, I can provide a canonical example to enable a local repro.  But it's a pretty basic setup:
A.schema.json
B.schema.json

quicktype .... A.schema.json
quicktype .... B.schema.json

main.cpp:
#include "A.hpp"
#include "B.hpp"

Thanks!

foobra:
add ifndef macro to supress re-definition errors

fix #1592


moodboom:
Hey all.  I need this fix, is there any reason it can't be merged?  C++ code generation is broken without it when you combine multiple generated files.  Adding a simple ifndef check seems like a very good idea to get this resolved.  Thanks!

moodboom:
Thanks so much!

Issue #1880, opened 2022-02-08
Support .Net 6 


eaydemir:
Support .Net 6 System.Text.Json library.

its only work with newtonsoft.json

bzuidgeest:
This is basically a duplicate of #1347

dvdsgl:
Fixed in #1997

PR #1944, opened 2022-05-08
Add "C# (System.Text.Json)" support

### LINKED ISSUE #1347: Support for .Net Core 3 Json library
.Net Core 3 comes with its own set of classes to serialise and deserialise JSON, located in the `System.Text.Json` and `System.Text.Json.Serialization`.
Previous versions used to rely on `Newtonsoft.Json` and `Newtonsoft.Json.Converters` instead.

This also means some attribute names need to be adapted:
```
        [JsonProperty("id")]
        public Guid Id { get; set; }
```
becomes 
```
        [JsonPropertyName("id")]
        public Guid Id { get; set; }
```

Serialisation of enums to string can be done this way
```
    [JsonConverter(typeof(JsonStringEnumConverter))]
    public enum Color { Red, Green, Blue };
```
but cannot be done by involving `JsonReader` and `JsonWriter` like it was in previous versions.

It would be nice to be able to choose between Newtonsoft.Json and System.Text.Json when using quicktype to create classes from Json.

Here is a good starting place to understand the changes I am refering to:
https://docs.microsoft.com/en-us/dotnet/core/whats-new/dotnet-core-3-0#fast-built-in-json-support

Cheers!

doggy8088:
This PR fixes #1347 and #1880

- Target Language Name: `C# (System.Text.Json)`
- Support `DateOnly` and `TimeOnly` (.NET 6)
- Pass all test samples (`test\inputs\json\samples\*.json`)


dvdsgl:
@doggy8088 thank you! Your branch is now up-to-date and running in CI, but I see test failures for your implementation–please take a look.

dvdsgl:
Moved to #1997

PR #1963, opened 2022-06-22
Swift: Add flag for optional enums

afajl:
Added swift option `optionalEnums`

Use case:
When an API where Enums are used and might be expanded later versions. Decoding would fail in those cases, with this option it's now decoded as nil.


johanrydenstam:
⭐ ⭐ ⭐ 
Very useful, been thinking about this addition!

dvdsgl:
@afajl thank you! Sorry for delay. We've just updated our CI to work again.

Can you add a test for this option? Add it to `quickTestRendererOptions` in `languages.ts` under the Swift definition.

Issue #1494, opened 2020-06-07
Placing to_json / from_json methods in nlohmann namespace leads to multiply compile errors


llilakoblock:
Take a look at this issue in nlohmann's repo: https://github.com/nlohmann/json/issues/780

Jusonex:
This is a duplicate of #1185. Unfortunately, this problem makes the codegen unusable with MSVC.

Have you considered integrating the solution suggested in https://github.com/quicktype/quicktype/issues/1185#issuecomment-506614282? It seems to fix the problem for me.

Issue #1333, opened 2019-11-01
C++ Compile Time Error with Googletest


Alph0r:
When trying to build a test project with Googletest in VS 2017 I receive several compile time errors:

binary '=': no operator found which takes a right-hand operand of type 'const std::vector<testing::internal::string,std::allocator<_Ty>>' (or there is no acceptable conversion)

nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>::get': no matching overloaded function found

This error seems to be related to ADL. Moving the block with the from_json/to_json functions from namespace nlohmann to namespace quicktype fixes the error.

zoink47:
I have the same problem with Microsoft Visual Studio 2019 16.3.7.

Issue #1185, opened 2019-03-09
Incorrect namespace in C++ to/from json (nlohmann)


hemantvallabh:
The to_json/from_json functions should be in the same namespace as the structure. I'm using Visual Studio 2017 (VC++)

Additional information: https://github.com/nlohmann/json/issues/780#issuecomment-376696058

**QuickType Code Gen**
```
#pragma once

#include "json.hpp"

namespace quicktype {
    using nlohmann::json;

    inline json get_untyped(const json & j, const char * property) {
        if (j.find(property) != j.end()) {
            return j.at(property).get<json>();
        }
        return json();
    }

    inline json get_untyped(const json & j, std::string property) {
        return get_untyped(j, property.data());
    }

    struct Welcome {
        int64_t response_code;
        std::string response_message;
    };
}

namespace nlohmann {
    void from_json(const json & j, quicktype::Welcome & x);
    void to_json(json & j, const quicktype::Welcome & x);

    inline void from_json(const json & j, quicktype::Welcome& x) {
        x.response_code = j.at("response_code").get<int64_t>();
        x.response_message = j.at("response_message").get<std::string>();
    }

    inline void to_json(json & j, const quicktype::Welcome & x) {
        j = json::object();
        j["response_code"] = x.response_code;
        j["response_message"] = x.response_message;
    }
}
```

**Working Code**
```
#pragma once
namespace quicktype {
	using nlohmann::json;

	inline json get_untyped(const json & j, const char * property) {
		if (j.find(property) != j.end()) {
			return j.at(property).get<json>();
		}
		return json();
	}

	inline json get_untyped(const json & j, std::string property) {
		return get_untyped(j, property.data());
	}

	struct Welcome {
		int64_t response_code;
		std::string response_message;
	};

	void from_json(const json & j, quicktype::Welcome & x);
	void to_json(json & j, const quicktype::Welcome & x);

	inline void from_json(const json & j, quicktype::Welcome& x) {
		x.response_code = j.at("response_code").get<int64_t>();
		x.response_message = j.at("response_message").get<std::string>();
	}

	inline void to_json(json & j, const quicktype::Welcome & x) {
		j = json::object();
		j["response_code"] = x.response_code;
		j["response_message"] = x.response_message;
	}
}
```

smalhotra-spirent:
[...]

PR #1109, opened 2018-11-07
User/lukec/windowscpp

cartwrightluke:
The main change in this PR is that I've added support for wstrings to the cpp language typescript. It's a new option that can be turned on or off. When this option is enabled, all std::string object properties are replaced with std::wstring properties. The Nholmann json library supports only Utf-8 std::strings, but when this option is enabled the header outputs conversion routines between the UTF-8 std::string json properties and the Utf-16 std::wstring header properties, allowing them to transparently convert. Both Keys and Values are supported.
Additional changes: 
1. Moved the to_json and from_json classes into nholmann::detail (I appeared to be hitting https://github.com/quicktype/quicktype/issues/1024, and this change resolved the issue)
1. Updated the readme with instructions for how to build the project on windows
1. Added forward declarations for the from_json and to_json functions. I noticed my success with these depended on the ordering, which was resolved with the forward declarations
1. Changed the type for get_min/max_length to size_t

Tested with the following schema:
```
/** #TopLevel */
class AddressBook {
  /**
   * A dictionary of Contacts, indexed by unique ID
   */
  contacts: { [id: string]: Contact };
}

class Contact {
  firstName: string;
  lastName?: string;
  
  birthday?: Date;
  title?: "Mr." | "Mrs." | "Ms." | "Prof.";
  
  emails: string[];
  phones: PhoneNumber[];
  
  /** @TJS-type integer */
  highScore: number ;
  
  timeout: string | PhoneNumber | number;
}

/**
 * A Contact's phone number.
 */
class PhoneNumber {
    number: string;
    
    /** An optional label (e.g. "mobile") */
    label?: string;
}
```

And the following data:
```
    quicktype::AddressBook data = nlohmann::json::parse(LR"(
    {
        "contacts" : {
            "te)"s L"\U0001F60E \U00000AE8 \U0001B18D"s  LR"(rry" : {
                "emails" : [ "abc@h)"s L"\U0001F60E \U00000AE8 \U0001B18D"s  LR"(otmail.com", "test2@microsoft.com"],
                "firstName" : "A)"s L"\U0001F60E \U00000AE8 \U0001B18D"s  LR"(J",
                "highScore" : 10,
                "phones" : [ { "number" : "2)"s L"\U0001F60E \U00000AE8 \U0001B18D"s  LR"(5" } ],
                "timeout" : "0.2"
            },
            "jesse" : {
                "emails" : [ "abc@hotmail.com", "test2@microsoft.com"],
                "firstName" : "AJ",
[...]

PR #1996, opened 2022-09-27
C++: Emit to_json/from_json helpers in the correct namespace

### LINKED ISSUE #1024: C++ code generated by quicktype does not compile in Visual Studio 2017 when "conformance mode" is off
I have generated C++ code from JSON Schema "Coordinate" and tried compiling it with VS 15.7.6. It turned out that if you set "Conformance mode" to "No", it doesn't compile. I have attached the sample project and a screenshot of the said option.
The error is as follows:

```
1>------ Build started: Project: quicktype_compilation, Configuration: Debug Win32 ------
1>stdafx.cpp
1>quicktype_compilation.cpp
1>s:\quicktype_compilation\quicktype_compilation\vd_json_serializer.h(18): error C2679: binary '=': no operator found which takes a right-hand operand of type 'double' (or there is no acceptable conversion)
...
```
You can see it yourself when compiling my project.

[quicktype_compilation.zip](https://github.com/quicktype/quicktype/files/2273880/quicktype_compilation.zip)
image

louiswins:
Right now quicktype emits `to_json` and `from_json` customizations in the `nlohmann` namespace. This is incorrect (although it happens to work with some compilers in some configurations depending on which methods you call). The correct approach is to emit the helpers in the _user code's_ namespace, e.g. the `quicktype` namespace in the default configuration.

#1109 attempted to fix this, but that PR took the approach of putting these functions in the `nlohmann::detail` namespace, which is even worse! (See https://github.com/nlohmann/json/discussions/3280#discussioncomment-2937051)

Here are some choices I made during this PR that I'd like feedback on since I'm not very familiar with this project's conventions:
- I unilaterally changed the namespace instead of adding an option. I thought this was appropriate because the old generated code is simply incorrect, but I am willing to hide this behind a new option.
- I removed the `msbuildPermissive` option entirely for similar reasons. I can put it back if this is a back-compat problem.

Additionally, the code now emits
```c++
namespace quicktype {
    /* type definitions */
}

namespace quicktype {
    /* to_json and from_json declarations & definitions */
}
```
with redundant close/open of the namespace. Fixing this would be a bigger change than I am comfortable making with my limited understanding of the quicktype code.

Fixes #1024, #1185, #1333, #1494 (and maybe others).


dvdsgl:
@louiswins thank you. Our test suite is now running again, apologies for the delay.

Can you check out the build errors? It looks like the `ourQualifier` type signature is mismatched.

dvdsgl:
@louiswins build now passes, but there are errors in tests.

To run locally, do:

```shell
$ FIXTURE=cplusplus,schema-cplusplus npm test
```

dvdsgl:
@louiswins would love to get this passing tests, but I don't know C++. Please reopen if you have some time!

louiswins:
Thanks for taking a look! And for getting the test suite running, this will make things a lot easier.

It looks like the problem is that now `boost::variant` deserializers are being emitted in the `quicktype` namespace so they aren't being found by [ADL](https://en.cppreference.com/w/cpp/language/adl). They need to be a specialization of `nlohmann::adl_serializer` instead (like is currently done for `std::shared_ptr<T>`). I hadn't tested with any variants so I didn't come across this earlier.

I'll look into it but it may take a while, so I'll leave this PR closed for now.

louiswins:
Actually I did get a chance to look at this today. Thanks again for fixing the build, I'm not sure what happened there. But I've got all the the tests passing now.

I don't think I can reopen this PR myself. @dvdsgl can you reopen it? Thanks for all your help!

PR #2013, opened 2022-11-05
removed @require and updated with require

HeyAvi:
@require deprecated so we need to update after making the model


Crayon2000:
It looks like a duplicate of PR #2000

dvdsgl:
Merged duplicate PR – thanks!

PR #2062, opened 2023-01-04
Remove YAML.parse from JSON parsing

dvdsgl:
For some reason, `YAML.parse` was being used instead of `JSON.parse`. Changing it seems to have no effect and we have security warnings about `YAML`.

@schani any idea why we were using `YAML`?


dvdsgl:
https://github.com/quicktype/quicktype/pull/1214

PR #2083, opened 2023-01-08
Don't use combinations samples for quickTestRendererOptions

dvdsgl:
This are skipped by many fixtures


dvdsgl:
These options are basically untested

PR #2087, opened 2023-01-08
Strict nullable support for Dart

dvdsgl:
Non-nullable types should not have ? annotation.

- [ ] Mark non-nullable properties `required`


dvdsgl:
@RmondJone want to help with this? Types that QuickType thinks are non-nullable should have non-nullable semantics in Dart.

RmondJone:
> @RmondJone want to help with this? Types that QuickType thinks are non-nullable should have non-nullable semantics in Dart.

Okay, I'm thinking about how to make the design more reasonable recently. My time zone is different from yours.  I am in China, and you are in the United States just the opposite.  Sorry for not seeing the information in time.

dvdsgl:
You can compare outputs of TypeScript and Dart with a simple sample where the `duration` property is nullable.

TypeScript knows the difference between `name` (non-nullable) and `duration` (nullable):

CleanShot 2023-01-08 at 16 19 12@2x

In Dart, there is no difference when there should be:

CleanShot 2023-01-08 at 16 19 28@2x

RmondJone:
@dvdsgl Hi，I have submitted a new PR, in which I will modify Dart's question。[#2135](https://github.com/quicktype/quicktype/pull/2135)

PR #2098, opened 2023-01-10
Fix type error in TypeScript output

blixt:
TypeScript with `noImplicitAny` enabled will error and fail to compile on the `let ref` statement because its type cannot be inferred. By explicitly declaring `let ref: any` the error goes away.

The error TypeScript outputs, for the curious:

> Variable 'ref' implicitly has type 'any' in some locations where its type cannot be determined. ts(7034)

Probably for another PR, but the `ref` variable is used before it is declared, which is technically not valid because referencing a `let` binding before it has been initialized throws in JavaScript. A `var` binding would be okay as it becomes hoisted to the top of the scope with an `undefined` value. But perhaps the better fix would be to move the statement up.

Screenshot 2023-01-10 at 21 20 36

### DIFF
--- a/src/quicktype-core/language/JavaScript.ts
+++ b/src/quicktype-core/language/JavaScript.ts
@@ -447,7 +447,7 @@ function transform(val${anyAnnotation}, typ${anyAnnotation}, getProps${anyAnnota
         return invalidValue(typ, val, key, parent);
     }
     if (typ === false) return invalidValue(typ, val, key, parent);
-    let ref = undefined;
+    let ref${anyAnnotation} = undefined;
     while (typeof typ === "object" && typ.ref !== undefined) {
         ref = typ.ref;
         typ = typeMap[typ.ref];


dvdsgl:
Thank you very much @blixt. I'd love to hear more about how you're using quicktype!

blixt:
@dvdsgl No worries! I am currently using `quicktype` in a tool UI that runs a BigQuery query to get thousands of items, augments them with extra data from APIs, puts everything in a JSON file, then I use `quicktype` to generate the types from that list and consume them in a regular Vite+TS+React app.

This is the part in the `Makefile` that creates the types in a way that worked best for this project:

```
data/SiteItem.ts: data/sites.json
	yarn run quicktype \
		--acronym-style camel \
		--nice-property-names \
		--prefer-unions \
		$< -o $@
```

Thanks for the quick merge! How often do you do NPM releases of changes? I can use `yarn patch-commit` for now to bring this change into the repo, but just wondering when I should check back here for the new version. :)

dvdsgl:
We publish to npm on merge to master, so it should be released.

PR #2103, opened 2023-01-12
Add missing export declarations for Python

hernando:


src/quicktype-core/index.ts
```diff
@@ -84,6 +84,7 @@ export { uriTypeAttributeKind } from "./attributes/URIAttributes";
 
 export { CPlusPlusTargetLanguage, CPlusPlusRenderer, cPlusPlusOptions } from "./language/CPlusPlus";
 export { CSharpTargetLanguage, cSharpOptions, CSharpRenderer } from "./language/CSharp";
+export { PythonTargetLanguage, PythonRenderer, pythonOptions } from "./language/Python";
```
hernando, line 87: @dvdsgl, I've just realized I missed JSONPythonRenderer in this list. Maybe it's meant to be internal, but I would like to extend this class because PythonRenderer does not contain the logic for `class` to `dict` and `dict` to `class`.

Issue #2097, opened 2023-01-10
just-type option for csharp


heavenwing:
quicktype version 18.0.0
There is no just-type option for csharp, earlier version have `--features just-types`

Xzelsius:
Related to #2091?
Might be caused by #1997

bzuidgeest:
I have checked and the options are still there. lines 1410 en 665 in the charp.ts  The only thing I can think of for this possible regression is that @dvdsgl  made some small changes to the PR before accepting it. One of them modified some export from the way they seem to have been. In this commit https://github.com/quicktype/quicktype/pull/1997/commits/9e8fd131e71ccdd59f0a4038d10fa3821e5a88f4

Maybe that causes the problem, making the option "undetected". The option that's missing is the same for both frameworks, so maybe it needs to be moved to a more common base option set,.

bzuidgeest:
@heavenwing I am not really done yet, so if you don't need the system.text.json framework option then I would suggest sticking with an older quicktype version for now. Seems the safe option
However, if you are in dire need I made a PR #2104  that I think fixes the problem, but it's not properly tested. So use at your own risk. If you could see if it works that would be usefull to me.

heavenwing:
thanks.
I have already used old version now.

获取 Outlook for iOS<https://aka.ms/o0ukef>
________________________________
发件人: bzuidgeest ***@***.***>
发送时间: Friday, January 13, 2023 6:02:59 AM
收件人: quicktype/quicktype ***@***.***>
抄送: Yongguang Zhu ***@***.***>; Mention ***@***.***>
主题: Re: [quicktype/quicktype] just-type option for csharp (Issue #2097)


@heavenwing<https://github.com/heavenwing> I am not really done yet, so if you don't need the system.text.json framework option then I would suggest sticking with an older quicktype version for now. Seems the safe option
However, if you are in dire need I made a PR #2104<https://github.com/quicktype/quicktype/pull/2104> that I think fixes the problem, but it's not properly tested. So use at your own risk. If you could see if it works that would be usefull to me.

―

Issue #2091, opened 2023-01-09
Error: Option parsing failed: Unknown option: --check-required. (Json Schema -> C#)


ChrisRichner:
Hi there

Since Version 16.0.11 and newer we face the following issue 

> Error: Option parsing failed: Unknown option: --check-required.

when generating C# Code with the following arguments

> quicktype -s schema schema.json -l csharp -o c:\temp\ --array-type list --check-required

bzuidgeest:
There are now two rendering frameworks, the old newtonsoft and system.text.json. from .net 6. Does the problem go away if you specify which framework you want to use?

ChrisRichner:
Thank you for this suggestion, 

I've tried both frameworks values with latest version 18.0.1 and the error remains eventough the error message is slightly different.

>  Error: Option parsing failed: UNKNOWN_OPTION: Unknown option: --check-required.

dvdsgl:
I bet this is the same as root cause as #2097

bzuidgeest:
@ChrisRichner I am not really done yet, so if you don't need the system.text.json framework option then I would suggest sticking with an older quicktype version for now. Seems the safe option
However, if you are in dire need I made a PR #2104  that I think mostly fixes the problem, but it's not properly tested. 
The check-required option is back, but it's only used by the old newtonsoft renderer. I did not have time to (re)-enable it for the new systemtextjspon renderer.
So use at your own risk. If you could see if it works that would be usefull to me. I would require you being able to build your own version of quicktype though.

ChrisRichner:
@bzuidgeest  Thank you for your effort 💪 I've already pinned the version so that we're up and running until a mitigation is found.

ChrisRichner:
At least the Web Version now shows the options again 👍

![image](https://user-images.githubusercontent.com/7531283/212361707-9db11619-bcfc-4545-9cd1-04f33d69f000.png)

bzuidgeest:
> At least the Web Version now shows the options again 👍
> 
> ![image](https://user-images.githubusercontent.com/7531283/212361707-9db11619-bcfc-4545-9cd1-04f33d69f000.png)

Yes the options are back I pushed a quick fix. The functionality itself was never gone, it was just that the option was not recognized. But as it is the QuickType framework does not support showing options dependent on framework. Therefore the option Check-required at the moment only has a real effect for newtonsoft serialization, but shows for both. 
[...]

PR #2104, opened 2023-01-12
QuickFix for #2091 and #2097

bzuidgeest:
The missing options where relegated to the classes newtonsoftCSharpOptions and systemTextJsonCSharpOptions. Yet for both they are exactly the same at the moment. Therefore the quickest way to restore the missing options for me was to merge them back to the base options class cSharpOptions.

Two points to consider.

* check required is accepted for system.text.json, but the code for it is commented out, thus not functional. This was the case in the code I based this on. Probably there was some problem with it. I would like to look into that, but I am pushing this anyway for those who need it to use. It should still work as before for newtonsoft json. That renderer was not changed.
* I need to look into how to advertise renderer specific options. The option where still there, but quicktype couldn't "see" them

Fixing above requires more time. As the PR name suggest these are quickfixes, and certainly not properly tested. I consider this PR an announcement and something that the "desperate" might use.
In a little more time I hope to remove both bullet points. 

edit 1: I have done a little more reading and functionality of Check-required does not exist in NET 5 and 6 and even 7 though it has a required property doesn't offer similar functionality. I will have to think on this, but I think I should find a way to declare that option a newtonsoft framework only. Which means me understanding how to do bullet point 2.

### DIFF
--- a/src/quicktype-core/language/CSharp.ts
+++ b/src/quicktype-core/language/CSharp.ts
@@ -173,7 +173,24 @@ export const cSharpOptions = {
         ],
         "double",
         "secondary"
-    )
+    ),
+    features: new EnumOption("features", "Output features", [
+        ["complete", { namespaces: true, helpers: true, attributes: true }],
+        ["attributes-only", { namespaces: true, helpers: false, attributes: true }],
+        ["just-types-and-namespace", { namespaces: true, helpers: false, attributes: false }],
+        ["just-types", { namespaces: true, helpers: false, attributes: false }]
+    ]),
+    baseclass: new EnumOption(
+        "base-class",
+        "Base class",
+        [
+            ["EntityData", "EntityData"],
+            ["Object", undefined]
+        ],
+        "Object",
+        "secondary"
+    ),
+    checkRequired: new BooleanOption("check-required", "Fail if required properties are missing", false)
 };
 
 export class CSharpTargetLanguage extends TargetLanguage {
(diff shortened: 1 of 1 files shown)


dvdsgl:
@bzuidgeest thank you very much for the quick fix!

bzuidgeest:
@dvdsgl I appreciated the enthusiasm and I think the fix ok. But as I wrote it wasn't heavily tested....

If you don't mind I have a question related to this you might be able to answer. I have been looking at the other languages. And so far i see no example of options appearing and disappearing based on how other options are set. Are context aware or interdependent options even possible in the current way QuickType is build? Can I have an option that only appears for one framework and not the other?

dvdsgl:
That could be added as a capability.

bzuidgeest:
> That could be added as a capability.

Pretend I am not completely familier with quicktype code, typescript, nodejs and npm. Somthing that should not be that hard :)

Do you mean adding the possiblity for this functionality to quicktype as it doesn't exist as of this moment or is "capability" a keyword for a class or other functionality type that already exists somewhere.
You no doubt have limited time, but your answer is to short for me to get a start from

If interdependent or framework specific features is something that needs adding to the base framework as a whole, then I think there are more important things. As an alternative I could just name an option "check-required (newtonsoft only)" for a description.

dvdsgl:
Sorry, I haven't worked on that code for nearly four years. Your guess is as good as mine! I would just start by reading the code paths that get options.

PR #2129, opened 2023-01-16
Document `quicktypePropertyOrder` custom JSON schema property

danielsharvey:
Referring to #1340:

> It is an alternative approach trying to resolve the ordering issue for object properties of JSON-Schemas. Solution is to provide a propertyOrder string array on the object type. This array will be used to determine the order of properties for resulting language bindings. 

This was renamed `quicktypePropertyOrder` before merging in #1340.

This PR adds documentation to the FAQ.


dvdsgl:
Thank you @danielsharvey.

Issue #2166, opened 2023-02-09
C++ Codegen - throw std::exception& instead of const char*


SimonCahill:
Reviewing an internal PR, I noticed that Quicktype generates C++ code that throws `const char*` as exception types.
While this is technically allowed by the language, code **shouldn't** throw built-ins, but **should** throw objects derived from `std::exception`!

Some more information can be found here:
[isocpp.org](https://isocpp.org/wiki/faq/exceptions)
[StackOverflow](https://stackoverflow.com/a/6249350/2921426)

The reasoning behind throwing "complex" types are as follows:
 - as code evolves, it may become necessary to expand on the error
 - programmers generally except well-known types to be thrown (`catch (const std::exception& ex)` is not a rare sight!)

The best option would be to implement a custom `quicktype_error` or `parse_error` for the schemas and then have it contain the error information.

An example of generated code would be:

```cpp
#include <exception>
#include <string>

namespace foo {
    class quicktype_error: public std::exception {
        public:
            quicktype_error(const std::string& msg, /*other related info*/): exception(), m_error(msg) {
                // ???
            }

            const char* what() override { return someFuncToGatherAllErrorInfo().c_str(); }

        private:
            std::string m_error;
            /* other related info */
    };
}
```

dvdsgl:
Sounds good! Please consider a PR that changes this.

SimonCahill:
I'll try. I'll admit I'm having a hard time finding my way around the sources and (appropriately) adding things.
So far I've created a branch that'll simply throw `std::runtime_error`, but no custom exceptions. Haven't quite figured that one out yet

PR #2171, opened 2023-02-12
Quicktype C++ output now throws instances of std::runtime_error instead of const char*

SimonCahill:
As mentioned in my issue #2166, the current implementation of Quicktype's C++ code generator throws `const char*` as an exception type, instead of standard exception classes.

While technically legal, this is generally considered bad practice.

Instead, it is better to use well-known exception types instead of `const char*`.

# Things to note

 - This **will** break existing `catch` expressions in production code!
 - This uses the standard std::runtime_error class for throwing the exception

### DIFF
--- a/packages/quicktype-core/src/language/CPlusPlus.ts
+++ b/packages/quicktype-core/src/language/CPlusPlus.ts
@@ -1845,7 +1845,7 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
                         onFirst = false;
                     });
                     // this.emitLine('else throw "Input JSON does not conform to schema";');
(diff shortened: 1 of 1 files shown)


dvdsgl:
@SimonCahill no sure why the fixture does not pass – please take a look, thank you!

SimonCahill:
I'll have a look once I'm home. Fixing this stuff while at work is likely to be frowned upon - even though we use Quicktype.

SimonCahill:
Fixed the error. It was a copy-paste error. I accidentally replaced an `else` clause with a `default` clause

PR #2165, opened 2023-02-08
Add support for Java Records (JEP 395)

simon04:
Fixes #2019.

cc: @dvdsgl

PR #2179, opened 2023-02-16
Add support for Java Records (JEP 395)

### LINKED ISSUE #2019: Add support for Java Records (JEP 395)
Please support Java Records as an alternative output mode (similar to "Use lombok").

Java 16 introduced Java Records:

- JEP 395: Records – https://openjdk.org/jeps/395
- https://www.baeldung.com/java-record-keyword

The welcome example from https://app.quicktype.io/ should output

```java
package io.quicktype;

public record Welcome (
    String greeting,
    String[] instructions) {
}
```

The multiple JSON example should output

```java
// Album.java

package io.quicktype;

public record Album (
    String name,
    Artist artist,
    Track[] tracks) {
}

// Artist.java

package io.quicktype;

public record Artist (
    String name,
    long founded,
    String[] members) {
}

// Track.java

package io.quicktype;

public record Track (
    String name,
    long duration) {
}
```

Thank you for maintaining this great tool!

simon04:
Fixes #2019.

Followup for #2165.

cc: @dvdsgl

### DIFF
--- a/packages/quicktype-core/src/language/Java.ts
+++ b/packages/quicktype-core/src/language/Java.ts
@@ -102,6 +102,7 @@ export class JavaTargetLanguage extends TargetLanguage {
 }
 
 const javaKeywords = [
+    "_", // as of release 9, '_' is a keyword, and may not be used as an identifier
     "Object",
     "Class",
     "System",


dvdsgl:
@simon04 looks like this does not actually contain the code for records! That was reverted from master due to tests failing.

PR #2189, opened 2023-02-24
Scala3

Quafadas:
Follow up to https://github.com/quicktype/quicktype/pull/1932


Quafadas:
https://github.com/quicktype/quicktype/actions/runs/4260485439

Booyakashah!

Although I think I've realised it only tests JSON. Is there a way to include schema tests in a follow up?

dvdsgl:
@Quafadas You'll need to add a JSONSchema fixture too.

PR #2192, opened 2023-02-27
Require only node 18.12 (not 18.13)

mdornseif:
It seems that in https://github.com/quicktype/quicktype/pull/2128 node version 18.13 was required (by accident). That makes quicktipe unusable on Google App Engine which is still on node 18.12. Since Node 16.x is still supported until April 2024 seweshould at least support old Node 18.x releases.


dvdsgl:
@mdornseif can you please also update .nvmrc to the minimum version supported? This will cause CI to use this version of node.

mdornseif:
@dvdsgl Thanks for the hint; done.

PercevalSA:
Why don't we set the minimal version to 16 as it is still supported? I would like to use quicktype in a environment stuck with nodejs16 but I can't because of the minimal requirement set.

Is it possible to change from 18 to 16 ?

vnd:
Quicktype still seems to work if node version is set to 16.20.0 in package.json and package-lock.json (it was just a quick test though).

PR #2205, opened 2023-03-10
add missing optional header

beku-epitome:
The generated code doesn't compile without this. Please review my changes because this is my first 5 minutes in this codebase :sweat_smile:


dvdsgl:
@beku-epitome are you sure? The tests were passing without this.

beku-epitome:
Well it's not that hard to fix this problem in the tests, as they can just include the necessary `optional` header before the generated header is included. But testing the problem is pretty simple, compile a C++ source file with only the generated header included, and you get the compilation failure.

PR #2209, opened 2023-03-12
feat: Make the input name optional

taorepoara:


dvdsgl:
Can you explain?

taorepoara:
I'm using quicktype to generate classes from JSON Schema files:

```js
await schemaInput.addSource({ uris: ["component.schema.json"] });
```

My [root schema](https://github.com/lenra-io/components-api/blob/beta/api/component.schema.json) has a title and refs to the components schemas.

Since the root schema has a title, the name should not be required.

Also, I have a circular reference with components (like [Actionable](https://github.com/lenra-io/components-api/blob/beta/api/components/actionable.schema.json)) that has refs to the root schema. When the name is given, quicktype see the root schema called by a ref as a different element that the one given as input (and all its refs two...).

The final result is the same, but I think that the generation could be faster.

packages/quicktype-core/src/input/JSONSchemaInput.ts
```diff
@@ -1166,7 +1166,7 @@ export class JSONSchemaInput implements Input<JSONSchemaSourceData> {
             const refs = await refsInSchemaForURI(resolver, normalizedURI, givenName);
             if (Array.isArray(refs)) {
                 let name: string;
-                if (this._schemaSources.length === 1) {
+                if (this._schemaSources.length === 1 && givenName) {
```
dvdsgl, line 1169: Please add !== undefined

dvdsgl (changes requested):
Small lint change

packages/quicktype-core/src/input/JSONSchemaInput.ts
taorepoara: Done :smile:

PR #2210, opened 2023-03-12
feat(Rust): Use rename_all = "camelCase"

### LINKED ISSUE #528: Use rename_all = "camelCase" for rust instead of renaming each value individually
I have a json schema in which all attributes are in camelCase, so they are renamed to snake_case when generating rust code. Instead of adding a single [`#[serde(rename_all = "camelCase")]`](https://serde.rs/attr-rename.html) once, all attributes are renamed individually.

So while the expected output is the following,

```rust
#[derive(Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
pub struct File {
    id: String,
    #[serde(rename = "type")]
    file_type: String,
    name: Option<String>,
    file_name: Option<String>,
    mime_type: Option<String>,
}
```

I actually got the following:

```rust
#[derive(Serialize, Deserialize)]
pub struct File {
    #[serde(rename = "id")]
    id: String,
    #[serde(rename = "type")]
    file_type: String,
    #[serde(rename = "name")]
    name: Option<String>,
    #[serde(rename = "fileName")]
    file_name: Option<String>,
    #[serde(rename = "mimeType")]
    mime_type: Option<String>,
}
```

taorepoara:
Closes #528

### DIFF
--- a/packages/quicktype-core/src/language/Rust.ts
+++ b/packages/quicktype-core/src/language/Rust.ts
@@ -288,14 +288,6 @@ export class RustRenderer extends ConvenienceRenderer {
         return isCycleBreaker ? ["Box<", rustType, ">"] : rustType;
     }
 
-    private emitRenameAttribute(propName: Name, jsonName: string) {
-        const escapedName = rustStringEscape(jsonName);
-        const namesDiffer = this.sourcelikeToString(propName) !== escapedName;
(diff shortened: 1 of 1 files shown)


dvdsgl:
@taorepoara I am not sure this is sufficient–quicktype does more renaming than merely camel casing. I've enabled the tests, so we will see.

PR #2214, opened 2023-03-13
feat(Rust): Use rename_all

taorepoara:
Detects the the most used naming style in the class properties and enum cases to define it in the rename_all when needed.

Replaces https://github.com/quicktype/quicktype/pull/2210
Closes https://github.com/quicktype/quicktype/issues/528


taorepoara:
I don't understand the test error. Can someone help me with this ?

taorepoara:
Ok, I understood the problem.

The `union` field of this schema test is renamed to `top_level_union`: https://github.com/quicktype/quicktype/blob/master/test/inputs/schema/pattern.schema#L12

Does someone know why the field name is not kept ?

taorepoara:
I could fix the problem with the unit tests.

I also add one with camel case properties to test the new feature.

packages/quicktype-core/src/language/Rust.ts
```diff
@@ -314,11 +374,23 @@ export class RustRenderer extends ConvenienceRenderer {
...
+        // List the possible naming styles for every class property
+        const propertiesNamingStyles: { [key: string]: string[] } = {};
+        this.forEachClassProperty(c, "none", (_name, jsonName, _prop) => {
+            propertiesNamingStyles[jsonName] = listMatchingNamingStyles(jsonName);
+        });
+
+        // Set the default naming style on the struct
+        const defaultStyle = "snake_case";
+        const preferedNamingStyle = getPreferedNamingStyle(Object.values(propertiesNamingStyles).flat(), defaultStyle);
+        if (preferedNamingStyle !== defaultStyle)
```
dvdsgl, line 386: Please add braces around if body.

packages/quicktype-core/src/language/Rust.ts
```diff
@@ -52,6 +52,64 @@ export const rustOptions = {
...
+    },
+    "lowercase": {
+        regex: /^[a-z][a-z0-9]*$/,
+        toParts: (name: string): string[] => [name],
+        fromParts: (parts: string[]): string => parts.map(p => p.toLowerCase()).join(''),
+    },
+    "UPPERCASE": {
+        regex: /^[A-Z][A-Z0-9]*$/,
+        toParts: (name: string): string[] => [name],
+        fromParts: (parts: string[]): string => parts.map(p => p.toUpperCase()).join(''),
```
dvdsgl, line 109: You don't need to write the types for each toParts and fromParts wince you declared them for the entire struct.

packages/quicktype-core/src/language/Rust.ts
```diff
@@ -52,6 +52,64 @@ export const rustOptions = {
     leadingComments: new BooleanOption("leading-comments", "Leading Comments", true)
 };
 
+type nameToParts = (name: string) => string[];
```
dvdsgl, line 55: Type names should be like NameToParts.

packages/quicktype-core/src/language/Rust.ts
```diff
@@ -52,6 +52,64 @@ export const rustOptions = {
...
+type partsToName = (parts: string[]) => string;
+type namingStyle = {
+    regex: RegExp,
+    toParts: nameToParts,
+    fromParts: partsToName,
+};
+
+const NamingStyle: { [key: string]: namingStyle } = {
```
dvdsgl, line 63: Values should be named like namingStyle.

dvdsgl: The type should be Record<string, NamingStyle>

dvdsgl (changes requested):
Very nice work! Just a few small stylistic changes requested.

packages/quicktype-core/src/language/Rust.ts
taorepoara: Actually I use them to detect the input fields formats.
I also use them to detect if a field does not match the global type.

dvdsgl:
I’ve just fixed the tests, please merge master.

PR #2229, opened 2023-03-29
Make arrays optional when they should be in swift

pagr:


johanrydenstam:
This is a very good addition!

dvdsgl:
Thank you @pagr — can you please explain the change?

pagr:
This fixes the issue where optional arrays from graphql are decoded as non optional arrays in swift

PR #1771, opened 2021-08-01
TypeScript Zod target

Southclaws:
I use Zod daily and wanted a converter for it, not quite perfect but hoping to get some guidance to get this finished!

The tests don't pass yet, not sure why, I get:

```
TSError: ⨯ Unable to compile TypeScript:
error TS5083: Cannot read file '/quicktype/test/runs/typescript-zod-c65096/test/tsconfig.json'.
```

but some runs actually do produce some output so I'm probably just missing some minor config in the test setup.

PR #2238, opened 2023-04-09
Typescript zod

ryoid:
Made some fixes from #1771

There are 2 main issues I noticed so far
- When comparing diff via schema, the key that is used for the name of types is different. Seems to be first occurrence for schema, last for JSON (`0cffa.json`)
- Generated schemas sometimes references types that are defined after (`spotify-album.json`)
- Also, properly handling top-level arrays

Not sure how to address these problems, skipped these tests for now. Might look into it when I have time, but would appreciate any help.


dvdsgl:
@ryoid thank you! Please see build error.

ryoid:
> @ryoid thank you! Please see build error.

Updated!

dvdsgl:
Thank you @ryoid and @Southclaws !

Southclaws:
Oh this is awesome to see, thanks for finishing off what I started!

kriswest:
> Generated schemas sometimes references types that are defined after (spotify-album.json)

They do indeed! https://github.com/glideapps/quicktype/issues/2414

PR #2253, opened 2023-04-24
(python) Fix error related to having an ending double quote in the docstring

### LINKED ISSUE #2247: Python docstring should escape last character if it is a double quote
I'm using quicktype to create python code from my jsonschema. 
One of the descriptions ends in a double quote . 
The generated python code ends up with a syntax error, due to this quote.

Here is a quick reproducer, notice that the `description` of this ends up in a double quote (`"`).

```json
{
  "$id": "my-stuff",
  "description": "Cool \"stuff\"",
  "type": "object",
  "properties": {
    "foo": {
      "type": "boolean"
    }
  }
}
```

Which generates this part of a python file
```python
class Stuff:
    """Cool "stuff""""  # <-- This generates an error `Missing closing quote ["]`
    foo: Optional[bool]
```

The last double quote (or all of them) should be escaped to prevent the syntax error or one extra space should be added to prevent this from happening.

josejulio:
Before:
screenshot_2023-04-24T09:42:28

The error is
> Missing closing quote ["]

After:
screenshot_2023-04-24T09:40:51

Fixes #2247


packages/quicktype-core/src/language/Python.ts
```diff
@@ -275,7 +275,14 @@ export class PythonRenderer extends ConvenienceRenderer {
 
     protected emitDescriptionBlock(lines: Sourcelike[]): void {
         if (lines.length === 1) {
-            this.emitLine('"""', lines[0], '"""');
+            const docstring = modifySource(content => {
+                if (content.length > 0 && content[content.length - 1] === '"') {
```
dvdsgl, line 279: Try `endsWith`?

dvdsgl:
Thanks! One small change suggested.

josejulio:
Thanks @dvdsgl ! I will update it once the suggestion is available. I didn't see any, maybe it is still in the review queue?

dvdsgl (changes requested):
Use endsWith

packages/quicktype-core/src/language/Python.ts
josejulio:
```suggestion
                if (content.endsWith('"')) {
```

PR #2287, opened 2023-05-24
Swift: Fix optional enums

johanrydenstam:
The `optionalEnum` flag was a great extension. Unfortunately it didn't work if the actual enum was optional.

This PR fixes that, and makes use of third party code that needs to be added to the project where the generated models are used. This is an improvement, previously each generated file had an implementation of the property wrapper

### DIFF
--- a/test/languages.ts
+++ b/test/languages.ts
@@ -692,7 +692,6 @@ export const SwiftLanguage: Language = {
         { "access-level": "internal" },
         { "access-level": "public" },
         { protocol: "equatable" },
-        { "optional-enums": "true" },
         ["simple-object.json", { protocol: "hashable" }]
     ],
     sourceFiles: ["src/language/Swift.ts"]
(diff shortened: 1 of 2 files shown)


test/languages.ts
```diff
@@ -692,7 +692,6 @@ export const SwiftLanguage: Language = {
         { "access-level": "internal" },
         { "access-level": "public" },
         { protocol: "equatable" },
-        { "optional-enums": "true" },
```
pagr, line 695: We removed this because optional-enums now requires an external dependency. Since the swift tests are built using ´swiftc´ it's hard to have dependencies. One option would be to switch to using spm and ´swift´ to build and install dependencies, but that would probably slow down the testing. Another would be to change the `import OptionallyDecodable` to be within `#if canImport` declarations and inline a OptionallyDecodable implementation in main.swift, but that seems like an ugly solution. For now we disabled the tests (in the same way it is disabled for Alamofire).

PR #2291, opened 2023-05-25
Swift: Add option sendable

johanrydenstam:
Generate structs with `Sendable` conformance for upcoming concurrency in Swift.


gnrcampos:
Obviously late to the party here but as far as I understand we shouldn't require `--no-mutable-properties` in order to be able to mark the types as `Sendable` if we're using structs insetad of classes. Is there a reason why we are mandating this?
Thanks for the hard work and for this awesome addition by the way!

Issue #1205, opened 2019-04-12
Go output has spaces in const variables


andyedison:
Greetings!

First off, awesome little app! It's super cool!

I did notice when playing around with the example pokedex.json, the go output has spaces in the const variables

```go
type Type string
const (
	Bug Type = "Bug"
	Dark Type = "Dark"
	Dragon Type = "Dragon"
	Electric Type = "Electric"
	Fairy Type = "Fairy"
	Fighting Type = "Fighting"
	Fire Type = "Fire"
	Flying Type = "Flying"
	Ghost Type = "Ghost"
	Grass Type = "Grass"
	Ground Type = "Ground"
	Ice Type = "Ice"
	Normal Type = "Normal"
	Poison Type = "Poison"
	Psychic Type = "Psychic"
	Rock Type = "Rock"
	Steel Type = "Steel"
	Water Type = "Water"
)
```

inferrinizzard:
Closing this as a non issue, this works as intended for Go enums

PR #2309, opened 2023-06-16
fix(Golang): fix enums generation

gifanonim:
This PR fix golang enums generation requested in issue #1205 
Fix is only enum values type positioning (whitespaces change only).

## Before
```go
package schema

type Gender string
const (
	Female Gender = "female"
	Male Gender = "male"
)
```

## After
```go
package schema

type Gender string

const (
	Female Gender = "female"
	Male   Gender = "male"
)
```


dvdsgl:
Nice!

PR #2312, opened 2023-06-30
feat: add typescript-effect-schema language

timsuchanek:
Adding support for [`@effect/schema`](https://www.npmjs.com/package/@effect/schema), which is a modern alternative to Zod with full support for decoding / encoding (Zod only does decoding).

In order to prevent `const` statements not being defined yet when referencing them for object types, I used `@effect/schema`'s lazy feature. This can be further optimized by only doing that when it's actually needed. Trying to implement that though proved very difficult, so for now this is a fully working reference implementation.

Note, that the reformatting is just reflecting the `.prettierrc` of this repo. Let me know if you want that to be disabled.

### DIFF
--- a/.github/workflows/test-pr.yaml
+++ b/.github/workflows/test-pr.yaml
@@ -19,7 +19,7 @@ jobs:
 
             matrix:
                 fixture:
-                    - typescript,typescript-zod
+                    - typescript,typescript-zod,typescript-effect-schema
                     - javascript,schema-javascript
                     - golang,schema-golang
                     - cjson,schema-cjson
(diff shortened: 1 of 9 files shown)


timsuchanek:
I see that DeepSource is failing, however, the file I introduced is based on the zod implementation, which also has a few `any` in there. How can we make DeepSource succeed?

dvdsgl:
Thanks! Running CI now.

dvdsgl:
Bravo! Worked on first try.

schickling:
Awesome! Anything we can help to get this deployed to the website and published on NPM? :)

Issue #1277, opened 2019-07-26
const not generated as a single-value enum


rikoe:
According to https://json-schema.org/understanding-json-schema/reference/generic.html#constant-values, the `const` type has the same semantics as a single-value enum.

However, quicktype currently generates const values as either a string or an object (I have seen both depending on quicktype version), with no reference at all to the constant value specified in the schema.

I would expect it to exhibit the same behaviour during code-generation as for single-value enums (which is to say a generated enum type with a single value matching the constant string).

Example: https://app.quicktype.io?share=rogbg8vSs4VHh1E9Zuq5

magiepooh:
I want this feature too 👍

joelhoisko:
I tested this one out again, and there still isn't support for the `const` keyword, it doesn't get parsed into an single enum. Would really love to see this feature happen 👍

Also linked to https://github.com/quicktype/quicktype/issues/1225

zhenyakovalyov:
echoing @rikoe's request to introduce correct const value generation

kriswest:
Working on the same project @rikoe was when raising this issue 4 years ago and we still need this (an use cases have multiplied). Very happy to see [the PR](https://github.com/quicktype/quicktype/pull/2314) @DanielBretzigheimer raised 4 days ago to resolve! Will try and find time to test it out soon.

DanielBretzigheimer:
@dvdsgl: I think this ticket may be closed?

PR #2314, opened 2023-07-01
Support for "const" in JSON Schema

DanielBretzigheimer:
Hey everyone, I added support for `const` in JSON Schema. I created two commits, the first addresses the issue #1277 and the second improves handling of single value enums for typescript (because const is a single value enum).

If the property has the const type, I interpret it as an enum with a single value. For TypeScript, I added a new case that if an enum has only a single value, it is also created as a constant. This will also change normal enums with a single value. I don't know if this causes problems, for my use case this is okay. What do you think? If the typescript change is not accepted, I can remove it from the pull request.

Could someone please review my changes and give me feedback? Your assistance is greatly appreciated! :)

### DIFF
--- a/packages/quicktype-core/src/input/JSONSchemaInput.ts
+++ b/packages/quicktype-core/src/input/JSONSchemaInput.ts
@@ -676,6 +676,7 @@ async function addTypesInSchema(
 
     async function convertToType(schema: StringMap, loc: Location, typeAttributes: TypeAttributes): Promise<TypeRef> {
         const enumArray = Array.isArray(schema.enum) ? schema.enum : undefined;
+        const isConst = schema.const !== undefined;
         const typeSet = definedMap(schema.type, t => checkTypeList(t, loc));
 
         function isTypeIncluded(name: JSONSchemaType): boolean {
@@ -697,6 +698,9 @@ async function addTypesInSchema(
                 }
 
                 return enumArray.find(predicate) !== undefined;
+            }
+            if (isConst) {
+                return name === "string";
             }
             return true;
         }
@@ -901,8 +905,8 @@ async function addTypesInSchema(
             return unionType;
         }
 
-        const includeObject = enumArray === undefined && (typeSet === undefined || typeSet.has("object"));
-        const includeArray = enumArray === undefined && (typeSet === undefined || typeSet.has("array"));
+        const includeObject = enumArray === undefined && !isConst && (typeSet === undefined || typeSet.has("object"));
+        const includeArray = enumArray === undefined && !isConst && (typeSet === undefined || typeSet.has("array"));
         const needStringEnum =
             includedTypes.has("string") &&
             enumArray !== undefined &&
@@ -913,7 +917,8 @@ async function addTypesInSchema(
             schema.additionalProperties !== undefined ||
             schema.items !== undefined ||
             schema.required !== undefined ||
-            enumArray !== undefined;
+            enumArray !== undefined ||
+            isConst;
 
         const types: TypeRef[] = [];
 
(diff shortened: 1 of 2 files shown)


dvdsgl:
Cool!

Should we make constantification an option?

DanielBretzigheimer:
What do you have in mind? An option for the typescript part or on how to handle the const value in general?

kriswest:
Very happy to see this PR raised!
I'm agnostic as to whether an option is created as I would assume that a const would always be preferable to a single value enum. Hence, if one is created I'd suggest it is on by default ;-)
Will try and find some time to test the change in next few days.

DanielBretzigheimer:
I think the same. Enum with a single value doesn't make sense to me, but maybe someone has a use case where he needs this. This would be a breaking change for someone who already used this. I will add an option with the default set to generate constant values.

dvdsgl:
Just because it's unexpected behavior, and we have a principle that we don't make breaking changes to code generation without a strong reason. Any change to default generation will create diffs for thousands of users.

DanielBretzigheimer:
@dvdsgl so you mean we should also add an option for the recognition of const values? I am unsure about that because I think this is more like a bug because const values should be generated as single value enums. For the typescript part, I added an option `prefer-const-values` which changes the generation. If the option is set to false, the enums are generated as before.

What do you think?

dvdsgl:
Looks good as-is – thank you!

ulazdins-afs:
Hey!

Besides https://github.com/quicktype/quicktype/issues/2333, the new behavior also generates incorrect Swift models.
Consider this JSON Schema:
```
{
  "type": "object",
  "properties": {
    "valid": {
      "type": "string",
      "const": "true"
    }
  },
  "required": ["valid"]
}
```

quicktype command 
```
quicktype ConstantSchema.json  -s schema -l swift --just-types
```

will produce 
```
import Foundation

// MARK: - ConditionCheckResponseDTO
struct ConditionCheckResponseDTO {
    let valid: Valid
}

enum Valid: String {
    case validTrue
}
```
which is incorrect, the expected output would be
```
import Foundation

// MARK: - ConditionCheckResponseDTO
struct ConditionCheckResponseDTO {
    let valid: Valid
}

enum Valid: String {
    case validTrue = "true" // change here!
}
```

I wouldn't mind fixing it myself, but I tried to read source and it's a bit overwhelming 😀. I don't mind trying again but a few pointers in the right direction would be helpful.

Thank you!

DanielBretzigheimer:
@ulazdins-afs do other single value enums work? The const generation should be the same as a enum with only a single value

ulazdins-afs:
Sorry for late response.
Yes, in general they do. Take this example
```
{
  "type": "object",
  "properties": {
    "valid": {
      "type": "string",
      "const": "something"
    }
  },
  "required": ["valid"]
}
```
will produce a valid and working Swift code
```
import Foundation

// MARK: - ConditionCheckResponseDTO
struct ConditionCheckResponseDTO {
    let valid: Valid
}

enum Valid: String {
    case something
}
```

timbunce:
We have also been impacted by #2333, apparently caused by this change.

kriswest:
@DanielBretzigheimer It doesn't look like the `prefer-const-values` option is exposed to the CLI, I think you need to do so by adding a line [here](https://github.com/glideapps/quicktype/blob/9b570a73a896306778940c793c0037a38815304a/packages/quicktype-core/src/language/TypeScriptFlow.ts#L54). 

Also since this was merged schema generation we rely on fails with  `Error: Internal error: We got an empty string type.`, which is proving very hard to track down in the schemas, but start between 23.0.49 and 23.0.54, which seems to only differ by this PR's changes: https://npmdiff.dev/quicktype-core/23.0.49/23.0.54/

Still looking into it, but any thoughts appreciated (note we only have string consts)

kriswest:
...and I found the error in our schemas (disagreement between an enum restricting values for a field, composed with another schema that set a const value for that field that wasn't in the enum). Not sure why it only breaks after your PR.

I'll drop a quick PR to expose the option you added.

PR #2316, opened 2023-07-03
C# add keep property name options

CabinIkaros:


packages/quicktype-core/src/language/CSharp.ts
```diff
@@ -189,7 +190,8 @@ export const cSharpOptions = {
         "Object",
         "secondary"
     ),
-    checkRequired: new BooleanOption("check-required", "Fail if required properties are missing", false)
+    checkRequired: new BooleanOption("check-required", "Fail if required properties are missing", false),
+    keepPropertyName: new BooleanOption("keep-property-name", "Keep original field name generate", true),
```
dvdsgl, line 194: Does this change the default behavior? Please make this `false` if the default is currently equivalent to `false`.

packages/quicktype-core/src/language/CSharp.ts
```diff
@@ -301,6 +305,109 @@ function csNameStyle(original: string): string {
     );
 }
 
+function csNameStyleKeep(original: string): string {
+
+    const keywords = [
+        "abstract",
```
dvdsgl, line 311: Is this list declared somewhere else? I am sure we already have C# keywords defined somewhere, and we should not have two copies.

packages/quicktype-core/src/language/CSharp.ts
```diff
@@ -346,7 +453,10 @@ export class CSharpRenderer extends ConvenienceRenderer {
     }
 
     protected namerForObjectProperty(): Namer {
-        return namingFunction;
+        if (this._csOptions.keepPropertyName)
```
dvdsgl, line 456: If you're omitting `{` and `}`, please just do this as a ternary.

dvdsgl (changes requested):
Thank you! Some small changes requested.

packages/quicktype-core/src/language/CSharp.ts
CabinIkaros: > Is this list declared somewhere else? I am sure we already have C# keywords defined somewhere, and we should not have two copies.

I added it because I didn't find a list of keywords for C#, only for C++/Dart/Java and other languages

packages/quicktype-core/src/language/CSharp.ts
CabinIkaros: > If you're omitting `{` and `}`, please just do this as a ternary.

Thanks for your review, edited

Issue #2319, opened 2023-07-09
Error when generating from JSON Schema with overriden allOf property


DanielBretzigheimer:
Hey, I get the error `Value of type string is not valid JSON Schema at ...` when generating TypeScript from the following JSON Schema.

## JSON Schema file
```json
{
	"allOf": [
		{
			"$ref": "https://XXX.de/schemas/order/Test.json#"
		}
	],
	"properties": {
		"msgId": {
			"const": "Test456"
		}
	}
}

/* referenced Test.json */
{
	"type": "object",
	"properties": {
		"msgId": {
			"const": "Test123"
		}
	}
}
```

I will create a PR which resolves this issue.

schani:
Thanks for the report and for the PR! As part of a systematic triage sweep of the issue backlog (assisted by Claude), I re-tested this against current master (v23.3.0, commit 68ecb284).

I reconstructed the schema (allOf `$ref` to Test.json, with the root overriding `msgId`'s const) and ran it through current master — it now generates the correct TypeScript output (`enum MsgID { Test456 }`) with no error. This matches your merged fix in #2326 (commit bb5f0335), which handles the case where the enum-case intersection is empty by falling back to the overriding case.

Since this no longer reproduces on master, I'm closing it as fixed. Thanks again for tracking this down and fixing it — let us know if you still see this anywhere and we'll reopen.

PR #2326, opened 2023-07-10
fix: enum property overriding when using $ref #2319

DanielBretzigheimer:
I fixed a bug where a enum property which is overriden with allOf is not replaced correctly.

I am unsure if this causes other problems. Could you have a look on the change and provide feedback?


dvdsgl:
Thank you!

PR #2351, opened 2023-07-28
Add TypeScript Zod support for strings with date-time format

danwsong:
The TypeScript Zod generator currently converts strings with the `date-time` format to `zod.string()`. This modifies the generator to emit `zod.coerce.date()` instead, matching the TypeScript Flow generator.

### DIFF
--- a/packages/quicktype-core/src/language/TypeScriptZod.ts
+++ b/packages/quicktype-core/src/language/TypeScriptZod.ts
@@ -1,10 +1,14 @@
+import { StringTypeMapping } from "TypeBuilder";
 import { arrayIntercalate } from "collection-utils";
-import { ClassProperty, EnumType, ObjectType, Type } from "../Type";
-import { matchType } from "../TypeUtils";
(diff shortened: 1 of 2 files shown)


packages/quicktype-core/src/language/TypeScriptZod.ts
```diff
@@ -38,6 +39,13 @@ export class TypeScriptZodTargetLanguage extends TargetLanguage {
         super(displayName, names, extension);
     }
 
+    get stringTypeMapping(): StringTypeMapping {
+        const mapping: Map<TransformedStringTypeKind, PrimitiveStringTypeKind> = new Map();
+        const dateTimeType = "date-time";
```
kvnxiao, line 44: Recommend adding `"date"` format type support as well. `zod.coerce.date()` supports dates without time, and this is also a valid format specified by JSON Schema https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times

dvdsgl:
Thank you!

Issue #2347, opened 2023-07-27
Feature: Make print function selectable (cJSON)


pseudotronics:
Add an option for using the unformatted print vs the default formatted print.
For low resource systems the formatted print takes up unnecessary or unavailable memory.

User should have the option between using

`cJSON_Print`

or

`cJSON_PrintUnformatted`

schani:
We do not take feature requests anymore. Please consider making a pull request instead. We encourage the use of coding agents, but please use at least an Opus-class model.

Issue #2346, opened 2023-07-27
Feature: Add typedef structs as an option (cJSON)


pseudotronics:
Struct definitions are restrictive and a PITA. 

There should be a flag to change the code generation to use typedefs instead of structs.
I may try to tackle adding this in myself, but I am not much of a typescript programmer.

So:
```
struct DemoAppState {
    bool * green_led;
    bool * red_led;
    bool * user_button;
};
```
and
```
struct DemoAppState * cJSON_ParseDemoAppState(const char * s);
struct DemoAppState * cJSON_GetDemoAppStateValue(const cJSON * j);
cJSON * cJSON_CreateDemoAppState(const struct DemoAppState * x);
char * cJSON_PrintDemoAppState(const struct DemoAppState * x);
void cJSON_DeleteDemoAppState(struct DemoAppState * x);
```
would become:
```
typedef struct {
    bool * green_led;
    bool * red_led;
    bool * user_button;
} DemoAppState;
```
and
```
DemoAppState * cJSON_ParseDemoAppState(const char * s);
DemoAppState * cJSON_GetDemoAppStateValue(const cJSON * j);
cJSON * cJSON_CreateDemoAppState(const DemoAppState * x);
char * cJSON_PrintDemoAppState(const DemoAppState * x);
void cJSON_DeleteDemoAppState(DemoAppState * x);
```
All other references would need to be updated within the body of the functions also.

mundusnine:
This can be closed as doing `--typedef-alias add-typedef` was implemented by you ;)

https://github.com/glideapps/quicktype/blob/9b570a73a896306778940c793c0037a38815304a/packages/quicktype-core/src/language/CJSON.ts#L85

schani:
We do not take feature requests anymore. Please consider making a pull request instead. We encourage the use of coding agents, but please use at least an Opus-class model.

PR #2356, opened 2023-07-31
cJSON Additional Settings

pseudotronics:
These changes implement the feature requests in #2347 and #2346


dvdsgl:
Could you centralize the rendering of typdefed versus non-typedefed type names within a single function, so you are not testing the option in so many places?

pseudotronics:
> Could you centralize the rendering of typdefed versus non-typedefed type names within a single function, so you are not testing the option in so many places?

I will see what I can do.

pseudotronics:
I changed the implementation strategy. It is much cleaner now. 

Instead of modifying the existing type declarations I instead just add the typedef declaration after the original declaration. 

For example:
```
struct Poll {
    enum Event * event;
    int32_t * reqid;
    enum Request * request;
};

typedef struct Poll Poll;
```

PR #2365, opened 2023-08-11
Python: Ensure default properties follows non-default properties in dataclasses

### LINKED ISSUE #2226: Python - generated code for union properties can be invalid due to field ordering
When a property has a union type, the resulting Python code can be invalid as a union field having a default value can be ordered prior to a required field.

JSON example:
```{
    "$schema": "https://json-schema.org/draft-07/schema",
    "type": "object",
    "required": ["b"],
    "properties": {
        "a": {
            "oneOf": [
                {
                    "type": "string"
                },
                {
                    "type": "array",
                    "items": {"type": "string"}
                }
            ]
        },
        "b": {
            "type": "string"
        }
    }
}
```

Resulting Python code:
```@dataclass
class Test:
    a: Optional[Union[List[str], str]] = None
    b: str
```

But this results in the following error from the `dataclasses` module when I attempt to use this code:
```TypeError: non-default argument 'b' follows default argument```

Kylmakalle:
fixes #2226

### DIFF
--- a/packages/quicktype-core/src/language/Python.ts
+++ b/packages/quicktype-core/src/language/Python.ts
@@ -444,7 +444,7 @@ export class PythonRenderer extends ConvenienceRenderer {
     ): ReadonlyMap<string, ClassProperty> {
         if (this.pyOptions.features.dataClasses) {
             return mapSortBy(properties, (p: ClassProperty) => {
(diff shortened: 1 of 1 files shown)


dvdsgl:
Nice! Thank you.

Issue #2333, opened 2023-07-12
CLI type generation fails - s.codePointAt is not a function


BadMagic100:
I generate types as a pre-build step for one of my libraries. On the latest release, type generation has begun to fail unexpectedly with no change to input. See the error message occuring in my build workflow here: https://github.com/hpackage/HPackage.Net/actions/runs/5503071564/jobs/10081743010#step:7:15

Last known successful run was July 2nd on release 23.0.49

ulazdins-afs:
Here is a JSON Schema file that will produce the error:
```
{
  "anyOf": [
    {
      "$ref": "#/definitions/ConditionCheckResponseValid"
    },
    {
      "$ref": "#/definitions/ConditionCheckResponseInvalid"
    }
  ],
  "definitions": {
    "ConditionCheckResponseValid": {
      "type": "object",
      "properties": {
        "valid": {
          "type": "boolean",
          "const": true
        }
      },
      "additionalProperties": false,
      "required": ["valid"]
    },
    "ConditionCheckResponseInvalid": {
      "type": "object",
      "properties": {
        "valid": {
          "type": "boolean",
          "const": false
        },
        "reason": {
          "type": "string"
        }
      },
      "additionalProperties": false,
      "required": ["reason", "valid"]
    }
  },
  "$schema": "http://json-schema.org/draft-07/schema#"
}
```

And you get this error:
```
quicktype ConditionCheckResponseDTO.json  -s schema -l swift --just-types
Error: s.codePointAt is not a function.
```

A slightly different version of JSON Schema with basically the same information work as expected:
```
{
  "anyOf": [
    {
      "$ref": "#/definitions/ConditionCheckResponseValid"
    },
    {
      "$ref": "#/definitions/ConditionCheckResponseInvalid"
    }
  ],
  "definitions": {
    "ConditionCheckResponseValid": {
      "type": "object",
      "properties": {
        "valid": {
          "type": "boolean",
          "enum": [true]
        }
      },
      "additionalProperties": false,
      "required": ["valid"]
    },
    "ConditionCheckResponseInvalid": {
      "type": "object",
      "properties": {
        "valid": {
          "type": "boolean",
          "enum": [false]
        },
        "reason": {
          "type": "string"
        }
      },
      "additionalProperties": false,
      "required": ["reason", "valid"]
    }
  },
  "$schema": "http://json-schema.org/draft-07/schema#"
}
```

```
[...]

PR #2371, opened 2023-08-17
fix: const generation for not string types (#2333)

DanielBretzigheimer:
This should fix the bug in #2333, but I had to disable the functionality to generate const for not string types. The problem is that the cases in line 957 in `JSONSchemaInput.ts` currently expect the cases to be strings. But for const generation they could be every type which causes an error when the names are generated.

I may look into it when I have time, for now, I have added a FIXME comment. If someone would like to address this issue, feel free to fix this.


a3ng7n:
I'm running in to the issue that this pr resolves - any chance we could get movement on this? it's been open for a year now.

schani:
Thanks for this fix — the bug is real and was still crashing every target language on master until today (`Error: s.codePointAt is not a function` for any non-string `const`), and your approach — keeping non-string consts out of the string-enum path so they degrade to plain primitives instead of crashing — was the right call, honestly documented FIXME and all.

The branch is 77+ commits behind and `JSONSchemaInput.ts` has been refactored since, so rather than asking for a rebase we re-ported the change onto current master in #2875, with a test-first fixture (integer/number/boolean consts, which crashed generation outright before the fix, plus a string const pinning the unchanged single-case-enum behavior). One improvement over the original: the string-enum path is now gated on `typeof schema.const === "string"` rather than the declared type, which also fixes the degenerate `{"type": "string", "const": 5}` case. Your FIXME about proper literal-type support for non-string consts is preserved as future work. You're credited as co-author on both commits.

Closing in favor of #2875. This also resolves the crash reported in #2333 and #2516 once it merges.

PR #2407, opened 2023-09-18
Add support for UUIDs in PHP

pengpixel:
For PHP the UUID is already mapped as a string, but had no implementation yet and threw an error.
If you have a schema which includes UUIDs and want to generate code for e.g. C# and PHP, the latter will fail. Giving you the choice to either maintain 2 schema files or just use strings for UUID values.

So this PR will:
* keep the string mapping and resolve the todo by validating the value with a regex, as PHP does not have built-in support for UUIDs


schani:
Thanks for this fix — while reviewing all open PRs we confirmed the problem is still live on current master (any input containing UUID-format strings crashes PHP generation with `Error: transformedStringType.kind === "uuid"`), and your approach — treating UUID as a validated string with an anchored regex, covering all seven matcher branches — is exactly right.

The branch can't be merged anymore because `language/Php.ts` was since split into `Php/language.ts` + `Php/PhpRenderer.ts`, so we re-ported it onto the new layout in #2872, keeping your renderer changes and test input essentially as-is. You're credited as co-author on both commits. One deliberate deviation: instead of flipping `skipMiscJSON` to false (which would have expanded PHP's test scope to the whole misc corpus), the UUID sample lives in the priority inputs and is whitelisted for PHP explicitly.

Closing this in favor of #2872.

PR #2418, opened 2023-10-02
Fixing code generation for zod

### LINKED ISSUE #2414: Typescript zod (and effect) language uses block scoped variables before they've been declared
While attempting to generate zod schemas from a fairly large set of JSON Schemas we've noted multiple places where a block-scoped variable representing one of the schemas is used before its been declared, meaning the source file generated is unusable

```ts
export const RaiseIntentResultBridgeResponsePayloadSchema = z.object({
    "intentResult": IntentResultSchema,
});

...

export const IntentResultSchema = z.object({
    "context": z.union([ContextElementSchema, z.null()]).optional(),
    "channel": z.union([ChannelSchema, z.null()]).optional(),
});
```

The same happens in the Typescript generator, but it doesn't matter there as the interfaces aren't variables

```ts
 * The message payload typically contains return values for FDC3 API functions.
 */
export interface RaiseIntentResultAgentResponsePayload {
  intentResult: IntentResult;
}

...

export interface IntentResult {
  context?: ContextElement;
  channel?: Channel;
}
```

Presumably, this occurs due to a logical error in this code: https://github.com/glideapps/quicktype/blob/9b570a73a896306778940c793c0037a38815304a/packages/quicktype-core/src/language/TypeScriptZod.ts#L201-L228

I assume the issue also affects the TypeScript effect schema language as it uses the same code.

@ryoid @Southclaws @dvdsgl

kriswest:
By ensuring that referenced types are defined before they are referenced. Tested on a fairly large set of interconnected schemas for the [bridging API](https://fdc3.finos.org/docs/agent-bridging/spec) in the [FDC3 standard](https://fdc3.finos.org/docs/fdc3-standard).

Please let me know if there's a more idiomatic way to get to the types underlying sets, unions, intersections, arrays etc..

resolves #2414

@ryoid @Southclaws @dvdsgl
P.S. The TypeScript effect schema language uses the same code in its generation and probably also needs these changes - I don't have a test set-up for that one however.

### DIFF
--- a/packages/quicktype-core/src/language/TypeScriptZod.ts
+++ b/packages/quicktype-core/src/language/TypeScriptZod.ts
@@ -6,7 +6,17 @@ import { RenderContext } from "../Renderer";
 import { BooleanOption, Option, OptionValues, getOptionValues } from "../RendererOptions";
 import { Sourcelike } from "../Source";
 import { TargetLanguage } from "../TargetLanguage";
-import { ClassProperty, EnumType, ObjectType, PrimitiveStringTypeKind, TransformedStringTypeKind, Type } from "../Type";
+import {
+    ArrayType,
+    ClassProperty,
+    ClassType,
+    EnumType,
+    ObjectType,
+    PrimitiveStringTypeKind,
+    SetOperationType,
+    TransformedStringTypeKind,
+    Type
+} from "../Type";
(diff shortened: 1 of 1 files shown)


kriswest:
doh! forked the wrong repo. Will reopen PR from a new fork.

PR #2419, opened 2023-10-02
fix: Typescript zod (and effect) language uses block scoped variables before they've been declared

kriswest:
By ensuring that referenced types are defined before they are referenced. Tested on a fairly large set of interconnected schemas for the [bridging API](https://fdc3.finos.org/docs/agent-bridging/spec) in the [FDC3 standard](https://fdc3.finos.org/docs/fdc3-standard).

Please let me know if there's a more idiomatic way to get to the types underlying sets, unions, intersections, arrays etc..

resolves https://github.com/glideapps/quicktype/issues/2414

@ryoid @Southclaws @dvdsgl
P.S. The TypeScript effect schema language uses the same code in its generation and probably also needs these changes - I don't have a test set-up for that one however.

### DIFF
--- a/packages/quicktype-core/src/language/TypeScriptZod.ts
+++ b/packages/quicktype-core/src/language/TypeScriptZod.ts
@@ -6,7 +6,17 @@ import { RenderContext } from "../Renderer";
 import { BooleanOption, Option, OptionValues, getOptionValues } from "../RendererOptions";
 import { Sourcelike } from "../Source";
 import { TargetLanguage } from "../TargetLanguage";
-import { ClassProperty, EnumType, ObjectType, PrimitiveStringTypeKind, TransformedStringTypeKind, Type } from "../Type";
+import {
+    ArrayType,
+    ClassProperty,
+    ClassType,
+    EnumType,
+    ObjectType,
+    PrimitiveStringTypeKind,
+    SetOperationType,
+    TransformedStringTypeKind,
+    Type
+} from "../Type";
 import { matchType } from "../TypeUtils";
 import { AcronymStyleOptions, acronymStyle } from "../support/Acronyms";
 import {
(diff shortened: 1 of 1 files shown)


kriswest:
@ryoid @Southclaws @dvdsgl any chance of a review on this PR? I believe it straightens out issues with generated typescript zod code in the current release.

dvdsgl:
Yep! Will review next weekend.

kriswest:
Much appreciated @dvdsgl

packages/quicktype-core/src/language/TypeScriptZod.ts
```diff
@@ -183,49 +193,142 @@ export class TypeScriptZodRenderer extends ConvenienceRenderer {
...
-                        ordinal = Math.max(ordinal, depIndex + 1);
+        //Items to process on this pass
+        let indices: number[] = [];
+        mapType.forEach((_, index) => {
+            indices.push(index);
+        });
+        //items to process on the next pass
+        let deferredIndices: number[] = [];
+
+        //defensive: make sure we don't loop foreever, even complex sets shouldn't require many passes
```
kriswest, line 282:
```suggestion
        //defensive: make sure we don't loop forever, even complex sets shouldn't require many passes
```

packages/quicktype-core/src/language/TypeScriptZod.ts
```diff
@@ -183,49 +193,142 @@ export class TypeScriptZodRenderer extends ConvenienceRenderer {
...
+            indices = deferredIndices;
+            deferredIndices = [];
+            passNum++;
 
-            // insert index
-            order.splice(ordinal, 0, index);
-        });
+            if (passNum > MAX_PASSES) {
+                //giving up
+                order.push(...deferredIndices);
```
kriswest, line 326:
```suggestion
                order.push(...deferredIndices);
                console.warn("Exceeded maximum number of passes when determining output order, output may contain forward references");
```

dvdsgl:
I’m pretty sure the type graph already provides this for other languages where declaration order matters, although I am not sure where that functionality is. I think we can still merge this however, and replace it if there’s a built-in way to do it.

dvdsgl:
Thank you!

PR #2426, opened 2023-10-17
fix: Expose the prefer-const-values Typescript language setting via CLI

### LINKED ISSUE #2345: Option parsing failed: Unknown option: --prefer-const-values.
The cli is not accepting `--prefer-const-values.` option. The reason as far as I can see is:
```typescript

export abstract class TypeScriptFlowBaseTargetLanguage extends JavaScriptTargetLanguage {
    protected getOptions(): Option<any>[] {
        return [
            tsFlowOptions.justTypes,
            tsFlowOptions.nicePropertyNames,
            tsFlowOptions.declareUnions,
            tsFlowOptions.runtimeTypecheck,
            tsFlowOptions.runtimeTypecheckIgnoreUnknownProperties,
            tsFlowOptions.acronymStyle,
            tsFlowOptions.converters,
            tsFlowOptions.rawType,
            tsFlowOptions.preferUnions,
            tsFlowOptions.preferTypes,
            tsFlowOptions.preferConstValues  // <- This is missing
        ];
    }
```

Worked when I added it in my node_modules.

kriswest:
resolves #2345 
A previous PR (#2314) introduced the ability to generate const values in Typescript but didn't expose that option correctly for use through the CLI. This one-line PR exposes the option (which works a treat!)

@DanielBretzigheimer

### DIFF
--- a/packages/quicktype-core/src/language/TypeScriptFlow.ts
+++ b/packages/quicktype-core/src/language/TypeScriptFlow.ts
@@ -51,7 +51,8 @@ export abstract class TypeScriptFlowBaseTargetLanguage extends JavaScriptTargetL
             tsFlowOptions.converters,
             tsFlowOptions.rawType,
             tsFlowOptions.preferUnions,
-            tsFlowOptions.preferTypes
+            tsFlowOptions.preferTypes,
+            tsFlowOptions.preferConstValues
         ];
     }
 


kriswest:
@dvdsgl @schani (or any other maintainer), is there any chance of getting this one-line correction merged? 

The Linux Foundation/FINOS' FDC3 Standard project uses quicktype to generate Typescript interfaces as part of its build pipeline and we're stuck with a manual process without this option (as const values resolved a problem for use in the generated TypeScript types - we found that single value string enums, the alternative to string consts in standardized messages, do not match the string values where actual typescript consts do). As we make use of quicktype for this, we're trying to contribute back on those things that we use.

kriswest:
Many thanks @schani, our automation will be back up and running again shortly!

PR #2443, opened 2023-11-16
Fix Python class member comment placement

usefulalgorithm:
## Description

For the variable descriptions to be correctly parsed as Python variable docstrings, they need to be place beneath the variable declarations.

Tested with 
[foo.json](https://github.com/glideapps/quicktype/files/13382029/foo.json)

### Before

截圖 2023-11-17 凌晨2 32 23

### After

截圖 2023-11-17 凌晨2 32 49


dvdsgl:
Thank you!

kopecn:
@usefulalgorithm  Many thanks for the fix.  

@dvdsgl Issue https://github.com/glideapps/quicktype/issues/1685 can be marked as dup/fixed.

PR #2454, opened 2023-12-06
fix: typecheck prs

airtonix:
fixes #2453 

I suspect that #2453 was allowed to occur because you're not typechecking


dvdsgl:
Builds fail?

Issue #2462, opened 2024-01-06
`golang` test failure on `test/inputs/json/misc/734ad.json`


dvdsgl:
```shell
$ FIXTURE=golang npm test -- test/inputs/json/misc/734ad.json
```

## All regressions

        "734ad.json",
        "1b409.json",
        "0a91a.json",
        "combinations1.json",
        "combinations2.json",
        "combinations3.json",
        "combinations4.json",

dvdsgl:
`null` values are not present in output in go:

<img width="1031" alt="CleanShot 2024-01-06 at 15 24 22@2x" src="https://github.com/glideapps/quicktype/assets/108197/64551fb8-3c74-4ebc-b438-5dd4cb74580c">

PR #2463, opened 2024-01-06
Skip regressed tests

dvdsgl:
* #2462


dvdsgl:
Something insane has happened – this passes locally.

PR #2490, opened 2024-02-12
Add useCow option for Rust

surma:
(I am not sure if there’s actually interest in this, but I thought I’d share an idea as a PR and we can take it from here.)

I am using QuickType to generate Rust structs so I can parse a [LDTK](https://ldtk.io) file (a JSON blob) that is bundled with my binary. Since the JSON blob is already embedded in the binary, it seems wasteful to use `String`. Instead, for most intents and purposes, it’s functionally identical to use `Cow<'_, str>` instead of `String`, but without having to copy all the strings to the heap.

This PR adds a new `--use-cow` option for Rust, which replaces the occurrences of `String>` with `Cow<'a, str>`. Because of the introduction of lifetimes, there is also a bit of machinery as a struct containing a string now needs to have a lifetime itself, and this needs to be recursively propagate up the type tree.

<details>
<summary>Example before/after</summary>
Before:

```rust
#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
pub struct World {
    /// Default new level height
    pub default_level_height: i64,

    /// Default new level width
    pub default_level_width: i64,

    /// User defined unique identifier
    pub identifier: String,

    /// Unique instance identifer
    pub iid: String,

    /// All levels from this world. The order of this array is only relevant in
    /// `LinearHorizontal` and `linearVertical` world layouts (see `worldLayout` value).
    /// Otherwise, you should refer to the `worldX`,`worldY` coordinates of each Level.
    pub levels: Vec<Level>,

    /// Height of the world grid in pixels.
    pub world_grid_height: i64,

    /// Width of the world grid in pixels.
    pub world_grid_width: i64,

    /// An enum that describes how levels are organized in this project (ie. linearly or in a 2D
    /// space). Possible values: `Free`, `GridVania`, `LinearHorizontal`, `LinearVertical`, `null`
    pub world_layout: Option<WorldLayout>,
}

```

After:

```rust
#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
pub struct World<'a> {    
    /// Default new level height
    pub default_level_height: i64,   
 
    /// Default new level width
    pub default_level_width: i64,

    /// User defined unique identifier
    pub identifier: Cow<'a, str>,

    /// Unique instance identifer
    pub iid: Cow<'a, str>,

    /// All levels from this world. The order of this array is only relevant in
    /// `LinearHorizontal` and `linearVertical` world layouts (see `worldLayout` value).
    /// Otherwise, you should refer to the `worldX`,`worldY` coordinates of each Level.
    pub levels: Vec<Level<'a>>,

    /// Height of the world grid in pixels.
    pub world_grid_height: i64,

    /// Width of the world grid in pixels.
    pub world_grid_width: i64,

    /// An enum that describes how levels are organized in this project (ie. linearly or in a 2D
    /// space). Possible values: `Free`, `GridVania`, `LinearHorizontal`, `LinearVertical`, `null`
    pub world_layout: Option<WorldLayout>,
}
```
</details>


dvdsgl:
We really appreciate the contribution, but our CI is in a broken state right now, which prevents us from evaluating new contributions.

PR #2494, opened 2024-02-14
Fixing failing unit test causing CI pipline to fail

### LINKED ISSUE #2462: `golang` test failure on `test/inputs/json/misc/734ad.json`

Adam724:
[This PR](https://github.com/glideapps/quicktype/pull/2449) introduced a new option to allow optional struct fields to be set to omitempty. However, it does not properly evaluate the boolean condition, and always behaves as if the option is set to true. This is causing some tests to fail unexpectedly when fields are set to null. The generated code contains the omitempty flag erroneously and thus ignores the null field when generating JSON output, causing the input/output JSON not to match. This PR fixes #2462.

### DIFF
--- a/packages/quicktype-core/src/language/Golang.ts
+++ b/packages/quicktype-core/src/language/Golang.ts
@@ -19,6 +19,7 @@ import { anyTypeIssueAnnotation, nullTypeIssueAnnotation } from "../Annotation";
 import { TargetLanguage } from "../TargetLanguage";
 import { ConvenienceRenderer } from "../ConvenienceRenderer";
 import { RenderContext } from "../Renderer";
+import _, { flatMap } from "lodash";
 
 export const goOptions = {
     justTypes: new BooleanOption("just-types", "Plain types only", false),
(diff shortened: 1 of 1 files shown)


dvdsgl:
THANK YOU!

Adam724:
Glad it is working! If you could take a look at [this PR](https://github.com/glideapps/quicktype/pull/2505) when you get a chance, that'd be great. Realized while testing a bit more today that the logic for `omitEmpty` is not right. It needs to still check whether the field is optional first before adding the tag. Ideally there would be a test case to catch this, but doesn't look like the current capabilities support passing CLI options to tests. I am looking into adding this, but will include it in a different PR since it's more involved.

PR #2509, opened 2024-02-16
Enhance testing capabilities

Adam724:
Currently, there is no way to pass cliOptions in the automated tests, making it difficult to test these flags. This PR aims to solve that with the following:
1. Adds a test configuration object in `test/lib/optionMap.ts`. This allows you to specify a test file path as key, and a configuration object as the value to customize the cliOptions, comparisonArgs, and target language for that test.
2. If the current test language does not match that specified in optionMap, the test is skipped. This is necessary since different languages support different CLI options.
3. Allows you to specify an input and expected output JSON file for tests. This is necessary since some cliOptions (such as omit-empty) are expected to produce a different JSON as output in some cases. I've adopted a standard file naming for input files matching this regex: `/^(.*)\.in\.(.*)$/`, and output files matching this regex: `/^\1\.out\.\2$/`. For example `myCustomTest.in.json` and `myCustomTest.out.json`. If an input file is found with no matching output, an error is thrown with a message indicating the issue.

No changes are necessary for any existing tests, this just adds the ability to write more customizable additional tests. Unsure if there is a better way to do this, any feedback is welcome!


test/fixtures/golang/main.go
```diff
@@ -1,11 +1,13 @@
...
-import "os"
+import (
+	"io"
+	"log"
+	"os"
+)
 
 func main() {
-	bytes, err := ioutil.ReadAll(os.Stdin)
+	bytes, err := io.ReadAll(os.Stdin)
```
Adam724, line 10: `ioutil.ReadAll` is depreacted from Go 1.16 and calls `io.ReadAll` instead. https://pkg.go.dev/io/ioutil#ReadAll

test/lib/optionMap.ts
```diff
@@ -0,0 +1,23 @@
...
+import { GoLanguage, Language } from "../languages";
+
+type TestOptions = {
+    cliOptions: RendererOptions;
+    language: Language;
+    comparisonArgs?: Partial<ComparisonArgs>;
+};
+
+const optionMap: Record<string, TestOptions> = {
+    "test/inputs/json/misc/227e2.in.json": {
```
Adam724, line 12: This means, for `227e2.in.json`, the test will be run with `omit-empty` set to true, strict comparison of input and expected output, and will only run for Golang language.

schani:
Thanks @Adam724 — this PR pointed at a real gap, and both of its ideas ended up in the codebase in some form.

The per-test CLI options half has since been superseded: `quickTestRendererOptions` in `test/languages.ts` now covers running specific inputs with specific renderer options, so the `optionMap` mechanism isn't needed anymore.

The other half — expected-output files so a test can assert output that legitimately differs from the input (your Go `omit-empty` case) — was still missing, and lives on in #2961, rebuilt on the JSON Schema fixtures' existing `.out.<feature>.json` convention and using your test input, with credit as co-author. Leaving this PR open for reference until that lands.

Issue #2516, opened 2024-02-19
const number type throws error


jonnytest1:
this schema doesnt work
(in my other example there was also type:"number" but it had the same effect)

```
{
    "type": "object",
    "properties": {
        "delay": {
            "const": 10
        }
    },
    "$schema": "http://json-schema.org/draft-07/schema#"
}

```

jonnytest1:
>> i tried adding this at the start of the Javascript.ts#nameStyle method which at least fixes the runtime error 
unfortunatly it is still treated as a string 

`
protected nameStyle(original: string, upper: boolean): string {
        if (typeof original === "number") {
            return original
        }
`

![image](https://github.com/glideapps/quicktype/assets/13507796/553e3ebd-9862-4d4b-bc48-f94913665c9d)

jonnytest1:
found it 
![image](https://github.com/glideapps/quicktype/assets/13507796/7dc07410-0d89-4543-a8e8-3909f139d69b)

jonnytest1:
🤔 now i just need to get out the `number |`

![image](https://github.com/glideapps/quicktype/assets/13507796/1dd54345-1e32-4e0b-bc77-75129745427d)

inferrinizzard:
Looks like a duplicate of #2333 #2371

jonnytest1:
true , but mine has a PR linked :3

PR #2517, opened 2024-02-19
fixed number const not working

jonnytest1:
PR for https://github.com/glideapps/quicktype/issues/2516

### DIFF
--- a/packages/quicktype-core/src/input/JSONSchemaInput.ts
+++ b/packages/quicktype-core/src/input/JSONSchemaInput.ts
@@ -943,7 +943,9 @@ async function addTypesInSchema(
                 if (!includedTypes.has(name)) continue;
 
                 const attributes = isNumberTypeKind(kind) ? numberAttributes : undefined;
+                if (!isConst) {
                 unionTypes.push(typeBuilder.getPrimitiveType(kind, attributes));
+                }
             }
 
             const stringAttributes = combineTypeAttributes(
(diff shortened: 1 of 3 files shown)


jonnytest1:
currently using "getStringType" even though its actually a number 🤔 i mean it works , but probably only because the other changes patched it a bit , maybe a number type is already somewhere but i didnt find it

jonnytest1:
oh i think const boolean should probably also be a case 🤔

schani:
Thanks for taking a crack at this — the underlying bug (#2516, non-string `const` crashing with `s.codePointAt is not a function`) is real and still reproduces on current master. But closing this PR, because after reviewing and testing it the fix doesn't hold up:

- The `JSONSchemaInput.ts` change is effectively dead code — both branches of the added `if (isConst && typeof schema.const === "number")` are identical, so the type model still stores a number inside a string-enum case set.
- Only the JavaScript/TypeScript renderer family is patched. Every other target language (verified with go, python, csharp) still crashes with the same error, since each has its own `nameStyle`.
- Even for TypeScript it only works with `--prefer-unions`, and only for the first enum case. Default TS mode emits `export enum Delay { 10 = "10", }`, which is invalid TypeScript. Boolean and null consts still crash everywhere.
- The branch conflicts unresolvably with master: `language/JavaScript.ts` and `language/TypeScriptFlow.ts` no longer exist (they were split into directories).

#2371 takes the better approach for the same bug — treating non-string consts as plain primitive types instead of string enums — which fixes the crash for all const types and all target languages with no per-renderer hacks. The real long-term fix is proper literal-type support for non-string consts. Keeping #2516 open until one of those lands.

PR #2518, opened 2024-02-20
Swift - CodingKeys improvements

brenoxp:
This PR enhances the Swift code generator within the `quicktype-core` package by incorporating user-defined access level settings into the CodingKeys enum, alongside its ability to conform to additional protocols. This dual enhancement not only broadens the scope of customization for generated Swift code but also ensures consistency in access control and protocol conformance, facilitating more refined and secure code generation strategies.

**Key Updates:**

- A new feature is introduced allowing the CodingKeys enum to inherit the access level specified by the user, ensuring that the enum's visibility aligns with the rest of the generated code.
- The capability to specify additional protocols for the CodingKeys enum has been expanded to include the applied access level, offering a more integrated approach to custom protocol conformance.

**How to Use:**

```bash
quicktype \
  --src person.json \
  --src-lang json \
  --lang swift \
  --access-level public\
  --coding-keys-protocol "CaseIterable"\
  --out Person.swift
```


dvdsgl:
Thank you!

Issue #2519, opened 2024-02-20
Go - time package is not being imported in certain cases


acline443:
In one of the recent releases of quicktype, native time type support has been added for Go. I have had some success with this but there seems to be a bug where in some circumstances the time.Time type is being referenced in the generated code but the import for "time" is not being added. 

Steps to reproduce, use the following schema:
```json
{
    "$schema": "http://json-schema.org/draft-06/schema#",
    "$id": "https://api.verosint.com/v1/events/request",
    "type": "object",
    "properties": {
        "timeRange": {
            "$ref": "#/definitions/timeRange"
        }
    },
    "additionalProperties": false,
    "required": [
        "timeRange"
    ],
    "definitions": {
        "timeRange": {
            "type": "array",
            "items": {
                "type": "string",
                "format": "date-time"
            },
            "maxItems": 2,
            "minItems": 1,
            "uniqueItems": true
        }
    }
}
```

Run generation command: 
```bash
quicktype -s schema --src schemas/*.json -o generated/models.go --package generated
```

The generated models.go file:
```go
// This file was generated from JSON Schema using quicktype, do not modify it directly.
// To parse and unparse this JSON data, add this code to your project and do:
//
//    models, err := UnmarshalModels(bytes)
//    bytes, err = models.Marshal()

package generated

import "encoding/json"

func UnmarshalModels(data []byte) (Models, error) {
	var r Models
	err := json.Unmarshal(data, &r)
	return r, err
}

func (r *Models) Marshal() ([]byte, error) {
	return json.Marshal(r)
}

type Models struct {
	TimeRange []time.Time `json:"timeRange"`
}
```

Notice: the import for "time" is missing despite the reference to the package in the struct definition.

PR #2528, opened 2024-03-09
fix: Go time import bug

rekram1-node:
fixes: [#2519](https://github.com/glideapps/quicktype/issues/2519)

If a json schema only has date-time referenced as an item of an array the generated Go code will not have the import for the time package. This adds an additional case to force time package imports

Changes:
* Added additional clause for time imports


dvdsgl:
Thank you!

Issue #2530, opened 2024-03-09
Go time import is misplaced in some circumstances.


rekram1-node:
My fix for [#2528](https://github.com/glideapps/quicktype/pull/2528) fixed the bug where Go code was missing the time import. However, it seems a new bug was introduced where in some cases the time import is not appearing at the top of the generated file.

Reproduction steps:
```json
{
    "type": "object",
    "properties": {
        "timeRange": {
            "type": "array",
            "items": {
                "type": "string",
                "format": "date-time"
            }
        }
    }
}
```

Generated Code:
```go
// This file was generated from JSON Schema using quicktype, do not modify it directly.
// To parse and unparse this JSON data, add this code to your project and do:
//
//    models, err := UnmarshalModels(bytes)
//    bytes, err = models.Marshal()

package generated

import "encoding/json"

func UnmarshalModels(data []byte) (Models, error) {
	var r Models
	err := json.Unmarshal(data, &r)
	return r, err
}

func (r *Models) Marshal() ([]byte, error) {
	return json.Marshal(r)
}

import "time"

type Models struct {
	TimeRange []time.Time `json:"timeRange,omitempty"`
}
```

Note the time import is not at the top of the file like the other import. This is a bug that causes the go code to not compile (syntax error).

Looking into a fix for this hopefully will have a Pull Request ready shortly.

rekram1-node:
Found the issue, basically just a 1 liner fix that I missed the first time around.

PR #2531, opened 2024-03-09
fix: Bug with misplaced Go time imports

rekram1-node:
fixes: [#2530](https://github.com/glideapps/quicktype/issues/2530)

My previous fix forced generated Go code to add time imports for []time.Time types, however, the change was missing an additional update included in this fix which will ensure that the import takes place at the top of the file (otherwise it would encounter a syntax error)

Changes:
* formatted a line
* Added additional clause for time imports

Reproduction to fix bug:

Schema:
```json
{
    "type": "object",
    "properties": {
        "timeRange": {
            "type": "array",
            "items": {
                "type": "string",
                "format": "date-time"
            }
        }
    }
}
```

Generated Code:
```go
// This file was generated from JSON Schema using quicktype, do not modify it directly.
// To parse and unparse this JSON data, add this code to your project and do:
//
//    models, err := UnmarshalModels(bytes)
//    bytes, err = models.Marshal()

package generated

import "time"

import "encoding/json"

func UnmarshalModels(data []byte) (Models, error) {
	var r Models
	err := json.Unmarshal(data, &r)
	return r, err
}

func (r *Models) Marshal() ([]byte, error) {
	return json.Marshal(r)
}

type Models struct {
	TimeRange []time.Time `json:"timeRange,omitempty"`
}
```

Notice: time import is where it should be, fixing  [#2530](https://github.com/glideapps/quicktype/issues/2530)

### DIFF
--- a/packages/quicktype-core/src/language/Golang.ts
+++ b/packages/quicktype-core/src/language/Golang.ts
@@ -307,7 +307,9 @@ export class GoRenderer extends ConvenienceRenderer {
 
         this.emitPackageDefinitons(
             false,
-            usedTypes.has("time.Time") || usedTypes.has("*,time.Time") || usedTypes.has("[],time.Time") ? new Set<string>(["time"]) : undefined
+            usedTypes.has("time.Time") || usedTypes.has("*,time.Time") || usedTypes.has("[],time.Time")
+                ? new Set<string>(["time"])
+                : undefined
         );
         this.emitDescription(this.descriptionForType(c));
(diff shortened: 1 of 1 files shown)


rekram1-node:
@dvdsgl Sorry to bug you just wanted to see if we could merge this change soon, this adds an additional fix that my [previous pull request](https://github.com/glideapps/quicktype/pull/2528) was missing

dvdsgl:
Thank you @rekram1-node !

PR #2606, opened 2024-05-25
fix(C#): added dependency usings based on types present in renderContext

sakets594:
## Description

change - forcing dependency namespace (`"System.Collections.Generic"`) to be added for list and dictionary in cases where common namespace addition(`CSharpRenderer.emitUsings`) was skipped

## Related Issue

https://github.com/glideapps/quicktype/issues/2523

## Motivation and Context

https://github.com/glideapps/quicktype/issues/2523

if you run the C# renderer with --features just-types --array-type list then it creates List<T> properties, but doesn't add the required using System.Collections.Generic;. It also happens without the --array-type list if a Dictionary<K,V> is generated as a property.

1. Example with dictionary 
command: `echo '{ "name": "David","keyval":{"1":{"prop1":1},"2":{"prop1":1},"3":{"prop1":3}} }' | script/quicktype -l csharp --features just-types`
output:
```cs
namespace QuickType
{

    public partial class TopLevel
    {
        public string Name { get; set; }
        public Dictionary<string, Keyval> Keyval { get; set; }
    }

    public partial class Keyval
    {
        public long Prop1 { get; set; }
    }
}
```

2. Example with array as list
command: `echo '{ "name": "David","arr":[1,2,3] }' | script/quicktype -l csharp --features just-types --array-type list` 
output:
```cs
namespace QuickType
{

    public partial class TopLevel
    {
        public string Name { get; set; }
        public List<long> Arr { get; set; }
    }
}
```

## New Behaviour / Output

with new change  `using System.Collections.Generic` gets added inside name space before class definition when required
1. Example with dictionary 
command: `echo '{ "name": "David","keyval":{"1":{"prop1":1},"2":{"prop1":1},"3":{"prop1":3}} }' | script/quicktype -l csharp --features just-types`
output:
```cs
namespace QuickType
{
    using System.Collections.Generic;

    public partial class TopLevel
    {
        public string Name { get; set; }
        public Dictionary<string, Keyval> Keyval { get; set; }
    }

    public partial class Keyval
    {
        public long Prop1 { get; set; }
    }
}
```

2. Example with array as list
command: `echo '{ "name": "David","arr":[1,2,3] }' | script/quicktype -l csharp --features just-types --array-type list` 
output:
```cs
namespace QuickType
{
    using System.Collections.Generic;

    public partial class TopLevel
    {
        public string Name { get; set; }
        public List<long> Arr { get; set; }
    }
}
```

## How Has This Been Tested?

Manual testing using command line, covered negative and positive cases of examples provided above
[EDIT] - tested with both array and dictionary present in json to validate single `using System.Collections.Generic;`


packages/quicktype-core/src/language/CSharp/CSharpRenderer.ts
```diff
@@ -387,4 +387,21 @@ export class CSharpRenderer extends ConvenienceRenderer {
...
+                _,
+                _arrayType => this._csOptions.useList ? nameSpaceForTypes.push("System.Collections.Generic") : undefined,
+                _classType => { },
+                _mapType => nameSpaceForTypes.push("System.Collections.Generic"),
+                _objectType => { },
+                _unionType => { }
+            )
+        });
+        nameSpaceForTypes = nameSpaceForTypes.filter((val, ind) => nameSpaceForTypes.indexOf(val) === ind );
+        nameSpaceForTypes.forEach(this.emitUsing.bind(this));
```
inferrinizzard, line 404: We don't need the array or filter steps, you can just directly `emitUsing` the `"Systems.Collections.Generic"` and return inside the `matchCompoundType`

sakets594: updated to handle only `using System.Collections.Generic;` case

inferrinizzard (approved):

PR #2623, opened 2024-06-29
Fix text in C++ default statement

Crayon2000:
## Description

* Fix text in C++ default statement
* Fix some typos in comments

## Related Issue

None that I know of.

## Motivation and Context

It fixes an error in the C++ output code. Problem was introduced in PR #2431.

## How Has This Been Tested?

Tested with this command: `quicktype --lang cpp --src-lang json ./test/inputs/json/samples/us-senators.json`

Before the changes we can see that we are getting a `[object Object]` for all enumeration names:
```cpp
    inline void to_json(json & j, const SenatorRankLabel & x) {
        switch (x) {
            case SenatorRankLabel::JUNIOR: j = "Junior"; break;
            case SenatorRankLabel::SENIOR: j = "Senior"; break;
            default: throw std::runtime_error("Unexpected value in enumeration \"[object Object]\": " + std::to_string(static_cast<int>(x)));
        }
    }

    inline void from_json(const json & j, Title & x) {
        if (j == "Sen.") x = Title::SEN;
        else { throw std::runtime_error("Input JSON does not conform to schema!"); }
    }

    inline void to_json(json & j, const Title & x) {
        switch (x) {
            case Title::SEN: j = "Sen."; break;
            default: throw std::runtime_error("Unexpected value in enumeration \"[object Object]\": " + std::to_string(static_cast<int>(x)));
        }
    }
```

## New Behaviour / Output

After the changes, we get the proper enumeration names:
```cpp
    inline void to_json(json & j, const SenatorRankLabel & x) {
        switch (x) {
            case SenatorRankLabel::JUNIOR: j = "Junior"; break;
            case SenatorRankLabel::SENIOR: j = "Senior"; break;
            default: throw std::runtime_error("Unexpected value in enumeration \"SenatorRankLabel\": " + std::to_string(static_cast<int>(x)));
        }
    }

    inline void from_json(const json & j, Title & x) {
        if (j == "Sen.") x = Title::SEN;
        else { throw std::runtime_error("Input JSON does not conform to schema!"); }
    }

    inline void to_json(json & j, const Title & x) {
        switch (x) {
            case Title::SEN: j = "Sen."; break;
            default: throw std::runtime_error("Unexpected value in enumeration \"Title\": " + std::to_string(static_cast<int>(x)));
        }
    }
```


Crayon2000:
For some reason the 'objective-c' test never run. Should I do something about this so my PR can be merged?

PR #2654, opened 2024-09-30
python: Fix for comments starting or ending with double quotes. Fixes #1239

### LINKED ISSUE #1239: python: invalid code due to comments starting or ending with double quotes
when generating python code, the generated syntax becomes invalid if there are comments (i.e. `"description"` fields in the JSON schema) that start or end with double quotes

# Example

for the JSON schema fragment:
```json
"properties": {
  "type": {
    "description": "type is \"output\"",
    "minLength": 1,
    "oneOf": [
	  {
	    "$ref": "#/definitions/MessageType"
	  }
    ]
  },
```
the generated (invalid) python 3.7 code fragment is
```python
"""type is "output""""
type: MessageType
```

# Solution

the generator for python should check the first and last character of a documentation-comment (i.e. the `"description"` field value) -- and if this is the case, the emitted code should "escape" these characters (e.g. prepend/append a newline or space character)

nithin-sudarsan:
… #1239

## Description

This fix handles both the cases where the comments might or end with a double quote `"`.

## Related Issue

#1239 

## Motivation and Context

The code earlier was replacing only the ending double quote, but the starting double quote was left unhandled. Moreover, while handling ending double quote, there was no whitespace to separating it at the end.

Same as above 👆

## New Behaviour / Output

If the line starts or end with double quote, both the cases are prepended and appended with whitespace accordingly.


russaa:
I think, with these changes, it should fix potential issues, when the comment starts with a double-quote too (actually, the problem with the previous implementation would have been, when comment would itself start with three double-quotes).

The escaping of all double-quotes in the comment-string should also take care of cases, when the comment contains three (or more) consecutive double-quotes which would be problematic in python.

so I would say, this should fix any potential issues for invalid comments in python due to double-quotes

(I tested these changes with the input that previously raised the issues with invalid python code for me, and the new output was in fact valid now)

nithin-sudarsan:
@russaa thank you for your feedback!
Can you help me with guiding on how do I get approval from a maintainer in order to complete the above workflow?

russaa:
sorry, I don't have any additional insight on how to best contribute to this project

schani:
Thanks for the contribution, and sorry it sat for so long. Closing this as superseded: master has since fixed #1239 via centralized comment escaping (28d1ca57 "Escape triple-quoted description comments", 19f35c7a "Centralize comment escaping in emitters", and 13b12f36 "Harden comment-injection escaping", which explicitly removed the Python trailing-quote code this PR modifies — so the branch now conflicts with master).

The centralized fix also covers cases this PR didn't: backslashes at the end of a description (which would still swallow the closing `"""`), embedded triple quotes, and multi-line docstrings. So there's nothing left to salvage here, unfortunately.

Verified while reviewing all open PRs; the underlying issue #1239 appears to be fixed on current master.

PR #2658, opened 2024-10-19
feat: expose various emit functions to subclasses

jhosteny:
## Description

This change makes the various Elixir `emit` functions `protected` instead of `private`, so that they can be overridden in customized code generators.

## Related Issue

https://github.com/glideapps/quicktype/issues/2657

## Motivation and Context

Allow these functions to be overridden in subclasses.

N/A

## New Behaviour / Output

N/A

## How Has This Been Tested?

`FIXTURE=Elixir npm test`

N/A


inferrinizzard:
Closed and re-opened to retrigger GH Actions

PR #2659, opened 2024-10-23
feat(typescript-effect-schema): support effect@^3.10.0

successkrisz:
## Context (-, typescript-effect-schema)

Input Format: -
Output Language: typescript-effect-schema

## Description

With the [effect v3.10.0 release](https://effect.website/blog/effect-3.10#effectschema-moved-to-effectschema), effect-schema became a stable version within the core `effect` package and `@effect/schema` should not be used anymore as newer effect versions are incompatible with it.

It already had the issue with the `Schema.Record` API moving from `Schema.Record(key, value)` to `Schema.Record({key, value})` this PR addresses that one as well

## Current Behaviour / Output

quicktype currently supports the usage of `@effect/schema@0.x` to generate the schemas which is deprecated as now effect/schema is part of the core `effect` library.

```typescript
import * as S from "@effect/schema/Schema"
// ...
S.Record(S.String, S.Number)
```

## Proposed Behaviour / Output

```typescript
import {Schema as S} from "effect"
// only other api change is how records are done:
S.Record({ key: S.String, value: S.Number })
```

## Solution

Changed import to use `effect` instead of `@effect/schema` for the imports and fixed the behaviour of Records

## Alternatives

No Alternatives

## Context

-

### DIFF
--- a/packages/quicktype-core/src/language/TypeScriptEffectSchema/TypeScriptEffectSchemaRenderer.ts
+++ b/packages/quicktype-core/src/language/TypeScriptEffectSchema/TypeScriptEffectSchemaRenderer.ts
@@ -76,7 +76,7 @@ export class TypeScriptEffectSchemaRenderer extends ConvenienceRenderer {
 
     protected emitImports(): void {
         this.ensureBlankLine();
-        this.emitLine(this.importStatement("* as S", '"@effect/schema/Schema"'));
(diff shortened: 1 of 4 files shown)


successkrisz:
Cancelling in favour of: https://github.com/glideapps/quicktype/pull/2646

successkrisz:
#2646

Issue #2662, opened 2024-11-04
[BUG]: [DEP0040] DeprecationWarning: The `punycode` module is deprecated


baraknaveh:
When running `quicktype` with `node` v22, it produces the following warning:
```
(node:42238) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
```

## Issue Type

The issue doesn't affect correctness, just pollutes the logs.

## Context (Environment, Version, Language)

Node version: v22
Input Format:  --src-lang schema
Output Language: --lang typescript

## Description

The issue doesn't affect correctness, just pollutes the logs.

## Input Data

N/A

## Expected Behaviour / Output

No warnings in log

## Current Behaviour / Output

```
(node:42238) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
```

## Steps to Reproduce

Run with node v22

davnotdev:
I am running into the same issue. Turns out, this has to do with dependencies downstream. Thankfully, the fix is quite simple, and just involves overriding a dependency's version. [PR here](https://github.com/glideapps/quicktype/pull/2673), [diff here](https://patch-diff.githubusercontent.com/raw/glideapps/quicktype/pull/2673.diff).

aguynamedben:
Bump! I'm new to Quicktype and also seeing this deprecation message.

baraknaveh:
Thanks for fixing!

PR #2673, opened 2025-01-08
Fix Runtime `punycode` Deprecation Error (#2662)

davnotdev:
Addresses #2662.

After the builtin `punycode` implementation was deprecated, `quicktype` now crashes at runtime from a deprecation error. This fix temporarily overrides `whatwg-url` to version `^13.0.0`, the earliest version that properly uses a userspace `punycode` implementation. 

`whatwg-url` is a dependency of `node-fetch 2.x` which is a dependency of `cross-fetch`. In the future, updating `cross-fetch` should fix the issue without having to override `whatwg-url` once [this issue](https://github.com/lquixada/cross-fetch/issues/177) is addressed.

Also, further work should be done to update `@vscode/vsce`, `eslint-config-canonical`, and `ajv` These are dev dependencies, so they should not affect runtime, but are still prone to the `punycode` error.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8487,7 +8487,6 @@
         },
         "node_modules/punycode": {
             "version": "2.3.1",
-            "dev": true,
             "license": "MIT",
             "engines": {
                 "node": ">=6"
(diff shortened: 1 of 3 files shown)


aguynamedben:
Thank you for this PR. +1. Workaround until this is merged is to add this to package.json.

```ts
  "//": "We can remove resolutions once this is merged: https://github.com/glideapps/quicktype/pull/2673",
  "resolutions": {
    "whatwg-url": "^13.0.0"
  }
```

aguynamedben (approved):
LGTM

inferrinizzard:
Fixed by #2744

PR #2675, opened 2025-01-14
Getters for optional values in cpp should return a const reference

### LINKED ISSUE #2674: [BUG]: Getters for optional values in C++ should return a const reference
Currently, the get_xyz() methods return something like:
`boost::optional<std::vector<Coordinate>> get_coordinates() const { return coordinates; }`
Which makes this code not behave as expected:
`for (auto coordinate: test.get_coordinates()) std::cout << "latitude" << coordinate.get_latitude() << std::endl;`

## Issue Type

This is an issue with quicktype C++ code generation

Input Format: Json Scheam
Output Language: C++

Version: Current

## Description

Returning a copy of the boost::optional means that user's can't iterate through the vector as they might initially expect, leading to bugs.  It's also presumably less efficient that returning a const reference to the underlying object.

## Input Data

`
{
  "$ref": "#/definitions/foo",
  "definitions": {
    "coordinate": {
      "type": "object",
      "required": [
          "latitude",
          "longitude"
      ],
      "properties":
      {
        "latitude": {
          "type": "number"
        },
        "longitude": {
          "type": "number"
        }
      }
    },
    "coordinates": {
      "type": "array",
      "items": {
          "$ref": "#/definitions/coordinate"
      }
    },
    "foo": {
      "type": "object",
       "required": [
       ],
       "properties": 
       {
         "coordinates": {
           "$ref": "#/definitions/coordinates"
         },
         "bar" : {
           "type": "number"
         }
      }
    }
  }
}
`

`const boost::optional<std::vector<Coordinate>>& get_coordinates() const { return coordinates; }`

## Current Behaviour / Output

`boost::optional<std::vector<Coordinate>> get_coordinates() const { return coordinates; }`

## Steps to Reproduce

Using above schema;
`quicktype --lang c++ -s schema --out ./test.h ./test.schema.json`
(or insert into app.quicktype.io)

flounderpinto:
**Overview**
Fixes #2674.  I can't entirely confirm that it doesn't have some unintended side effects, but this does fix the return type of optional values.

**Test file:**
`{
  "$ref": "#/definitions/foo",
  "definitions": {
    "coordinate": {
      "type": "object",
      "required": [
          "latitude",
          "longitude"
      ],
      "properties":
      {
        "latitude": {
          "type": "integer",
          "minimum": 0,
          "maximum": 999999
        },
        "longitude": {
          "type": "number",
          "minimum": 0,
          "maximum": 999999
        },
        "baz": {
          "type": "string",
          "minLength": 0,
          "maxLength": 999,
          "pattern": "*"
        }
      }
    },
    "coordinates": {
      "type": "array",
      "items": {
          "$ref": "#/definitions/coordinate"
      }
    },
    "foo": {
      "type": "object",
       "required": [
       ],
       "properties": 
       {
         "coordinates": {
           "$ref": "#/definitions/coordinates"
         },
         "bar" : {
           "type": "number"
         }
      }
    }
  }
}
`

**Diff of generated file changes:**
Screenshot from 2025-01-16 09-53-48
Screenshot from 2025-01-16 09-55-05

### DIFF
--- a/packages/quicktype-core/src/language/CPlusPlus/CPlusPlusRenderer.ts
+++ b/packages/quicktype-core/src/language/CPlusPlus/CPlusPlusRenderer.ts
@@ -1829,6 +1829,40 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
                 this.ensureBlankLine();
             }
         );
+
+        this.ensureBlankLine();
+
+        this.emitBlock(
+            [
+                "inline void ",
+                checkConst,
+                "(",
+                this._stringType.getConstType(),
+                " name, ",
+                this.withConst(classConstraint),
+                " & c, ",
+                "const ",
+                this._optionalType,
+                "<",
+                cppType,
+                "> & value)"
+            ],
+            false,
+            () => {
+                this.emitBlock(
+                    ["if (value)"],
+                    false,
+                    () => {
+                        this.emitLine(
+                            checkConst,
+                            "(name, c, *value);"
+                        );
+                    }
+                );
+                this.ensureBlankLine();
+            }
+        );
+
         this.ensureBlankLine();
     }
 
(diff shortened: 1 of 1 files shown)


schani:
Thanks for this fix — while reviewing all open PRs we confirmed the bug (#2674) is real: by-value optional getters let `for (auto& c : *obj.get_coordinates())` iterate a destroyed temporary (we reproduced a stack-use-after-scope under AddressSanitizer), and mutations through the getter are silently lost. Your fix was right in substance, including the crucial detail of adding the `checkConstraint` overloads for optionals that the const-ref setter path needs — without them, constrained optional properties wouldn't compile.

The branch conflicts with master by now (the renderer was reformatted since), so we re-ported the change onto current master in #2873, together with a schema fixture and runtime tests covering getter reference stability, mutation via `get_mutable_*`, roundtrips with optionals present/absent, and constraint enforcement through the new overloads (verified with both `std::optional` and `boost::optional`). You're credited as co-author on the fix commit.

Closing this in favor of #2873.

Issue #1632, opened 2021-01-17
C#: support nullable contexts


stephtr:
Starting with version 8.0 [C# supports nullable reference types](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-reference-types), which quicktype currently doesn't.

mixa3607:
Maybe this help https://github.com/mixa3607/quicktype

mindplay-dk:
@mixa3607 this looks good - you have support for partial classes in there as well. Why don't you submit a PR?

schani:
This is now implemented with `--csharp-version 8`, which is also the current default. Generated output enables the nullable context, uses `?` for optional or nullable reference-valued properties, and leaves required non-null references unannotated. The behavior works with both Newtonsoft.Json and System.Text.Json, so this issue can be closed.

PR #2692, opened 2025-03-26
Adds .NET 8 C# version and expands nullable types (#1632)

Flern:
## Description

Relaxes nullability check for .NET 8+ targeting.  .NET 8 introduced nullable reference types, thereby removing the restriction that only value types can be made nullable.

## Related Issue

(https://github.com/glideapps/quicktype/issues/1632)

## Motivation and Context

Specifically when using System.Text.Json, the new nullable ignore attribute does not trigger Roslyn warning CS8601, leading to potential runtime null reference exceptions.

Previously, reference types such as `string` would not include the `?` nullability indicator.

## New Behaviour / Output

Now, all C# types (reference and value) will include `?` when optional, since the serialized data may contain null values for those properties per the specification.

## How Has This Been Tested?

Ran before and after examples in a Docker environment.

image


Flern:
I have decided I don't like this approach.  Nullable reference types aren't tied explicitly to .NET 8 (in fact I believe they were introduced in .NET 5...).  I'll update this to introduce a nullable option instead of assuming nullable types based on a .NET version.

PR #2694, opened 2025-03-28
C#:  Adds version 8 and expands nullability to reference types (#1632)

Flern:
## Description

Relaxes nullability check for C# 8+ targeting.  C# 8 introduced nullable reference types, thereby removing the restriction that only value types can be made nullable.

## Related Issue

(https://github.com/glideapps/quicktype/issues/1632)

## Motivation and Context

Specifically when using System.Text.Json, the new nullable ignore attribute does not trigger Roslyn warning CS8601, leading to potential runtime null reference exceptions.

Previously, reference types such as `string` would not include the `?` nullability indicator.

## New Behaviour / Output

Now, all C# types (reference and value) will include `?` when optional, since the serialized data may contain null values for those properties per the specification.

## How Has This Been Tested?

Ran before and after examples in a Docker environment.

image


TheRubble:
Hey all....

We've just bumped into this issue due to using the models generated from the schema in a .NET API due to lack of support of nullable reference types the binder fails if the null, this looks like the fix we need. 

Is there anything we can do to support this PR?

Thanks

Issue #2719, opened 2025-04-29
[BUG]: Dart enum parsing throws on nullable values


jessicamrbr:
## Issue Type

Output Generation

## Context (Environment, Version, Language)

**Input Format:** JSON Schema  
**Output Language:** Dart  

**CLI, npm, or app.quicktype.io:** app.quicktype.io (also reproducible via CLI)  
**Version:** Latest as of April 2025

## Description

When generating Dart classes from a JSON Schema where a property is a nullable enum (e.g., `CommandShortcutType?`), the generated `fromMap()` function forcefully accesses the enum map with a non-null assertion operator (`!`), which throws a runtime exception if the value is `null` in the input JSON.

This makes the output unusable for any optional enum fields, even though the Dart class correctly marks the property as nullable.

## Input Data

```json
{
  "type": "object",
  "properties": {
    "shortcutType": {
      "type": "string",
      "enum": ["standard", "view"]
    }
  },
  "required": []
}
```

## Expected Behaviour / Output

Generated Dart code should respect nullability in fromMap() by safely handling null values for optional enum fields. Avoid using ! when accessing the enum map if the corresponding property is nullable.

Expected snippet:

```dart
shortcutType: json["shortcutType"] != null
    ? shortcutTypeValues.map[json["shortcutType"]]
    : null,
```
or
```dart
shortcutType: shortcutTypeValues.map[json["shortcutType"]]
// expect(json["shortcutType"], null)
// expect(shortcutTypeValues.map[null], null)
```

## Current Behaviour / Output

The current output forcefully unwraps the enum value, even when the Dart type is nullable:

```dart
commandShortcutType: commandShortcutTypeValues.map[json["commandShortcutType"]]!,
```

This causes a runtime crash when the input JSON omits or explicitly sets the value to null.

## Steps to Reproduce

Open https://app.quicktype.io/ . Paste the provided JSON Schema. Select output language as Dart. Observe that the generated fromMap() uses ! with optional enum types.

## Possible Solution

packages/quicktype-core/src/language/Dart/DartRenderer.ts
```
--                    this._options.nullSafety ? "]!" : "]"
++                    (this._options.nullSafety && !isNullable) ? "]!" : "]"
```

vitordeap:
Up!

Please review @jessicamrbr's PR, it's simple and it would fix this issue.

Thanks for the good work with this repo

djnugent:
[...]

PR #2720, opened 2025-04-29
Update DartRenderer.ts

jessicamrbr:
## Related Issue
(https://github.com/glideapps/quicktype/issues/2719)

## How Has This Been Tested?
no tested


packages/quicktype-core/src/language/Dart/DartRenderer.ts
```diff
@@ -469,7 +469,7 @@ export class DartRenderer extends ConvenienceRenderer {
                     defined(this._enumValues.get(enumType)),
                     ".map[",
                     dynamic,
-                    this._options.nullSafety ? "]!" : "]",
+                    (this._options.nullSafety && !isNullable) ? "]!" : "]"
```
vitordeap, line 472:
```suggestion
                     (this._options.nullSafety && !enumType.isNullable && !isNullable)? "]!" : "]",
```

In my tests, this is the code that worked

djnugent (approved):
Tested on my project and it works

schani:
I pushed three commits on top of this PR (thanks for enabling maintainer edits):

**1. `2119d71` — merge of current `glideapps/quicktype` master**, so CI runs against today's tree.

**2. `bd71c35` — a regression test for #2719**: `test/inputs/schema/optional-enum.schema`, a schema with an *optional* enum property, plus two samples (property present / property absent). The Dart schema fixture compiles and runs the generated code, so this exercises the bug end-to-end:

- At the merge base (before this PR's fix), the absent-property sample crashes at runtime:
  ```
  Unhandled exception:
  Null check operator used on a null value
  #0      new TopLevel.fromJson (TopLevel.dart:22)
  ```
- With the fix, `fromJson` emits `shortcutValues.map[json["shortcut"]]` (no `!`) for the nullable field and the sample round-trips as `"shortcut": null`.

The fixture is skipped for the same languages that already skip `enum.schema` (Crystal, CJSON, Swift, Scala3, Haskell, TypeScript-Zod, TypeScript-EffectSchema). I verified locally that Dart (`dart:stable` docker), Python, JavaScript, and Go round-trip both samples, and that every other schema-fixture language at least generates code for it cleanly. (I looked at un-skipping `enum.schema` for Dart now that the crash is fixed, but it still can't round-trip: absent optional *lists* come back as `[]`, the same issue that keeps `optional-const-ref.schema` skipped.)

**3. `e18a918` — a small addition to the fix itself.** Every sibling nullable-handling branch in `fromDynamicExpression`/`toDynamicExpression` (`mapList`, `mapMap`, the DateTime branch) also gates on `!this._options.requiredProperties`, because `--required-props` makes `dartType()` declare the field **non**-nullable even when the type is nullable. With only `!isNullable` in the condition, `null-safety` + `required-props` on the optional-enum schema generated non-compiling Dart:

```dart
Shortcut shortcut;                                  // non-nullable field
shortcut: shortcutValues.map[json["shortcut"]],     // expression type is Shortcut?
```

`dart analyze` flags it: `The argument type 'Shortcut?' can't be assigned to the parameter type 'Shortcut'. - argument_type_not_assignable` — whereas master's unconditional `]!` compiled for that combo. The commit changes the condition to

```ts
this._options.nullSafety && (!isNullable || this._options.requiredProperties)
```

which keeps this PR's behavior for default options (verified: still no `!` for nullable enums) and restores master's `]!` under `--required-props` (verified with `dart analyze`: the type error is gone). It also restores the trailing comma before `];` that the original change dropped.

Note on test coverage for the `required-props` combo: it can't be exercised by the round-trip harness — any data that makes an enum nullable necessarily contains null/absent values, which `--required-props` code rejects at runtime by design (same as master). So that path is verified by the `dart analyze` evidence above, while CI covers the main fix via the new fixture.

PR #2787, opened 2025-06-19
Fix missing/broken error messages

uuf6429:
## Description

Error messages were showing up like so:

> Error: Internal error: .

This is because of a problem with handling strings in the custom interpolation function.

## How Has This Been Tested?

Tested it manually, that after the change the original error showed up correctly.

----

PS: This is a rather critical bug IMO - it's decreasing the quality/trustworthiness of the project.


packages/quicktype-core/src/Messages.ts
```diff
@@ -286,6 +286,8 @@ export function messageError<Kind extends ErrorKinds>(
             valueString = value.message;
         } else if (typeof value !== "string") {
             valueString = JSON.stringify(value);
+        } else {
+            valueString = value;
```
uuf6429, line 290: Essentially, none of the cases above actually handled string, with the last else-if cases handling "everything else than strings".

inferrinizzard:
@uuf6429 Can you give an example of a reproduction and the resulting error messages before and after ?

uuf6429:
@inferrinizzard simply call `panic("test");` from anywhere, it will show the following in the console: `Error: Internal error: .` With my fix it would be `Error: Internal error: test.`

PS: I'm guessing this stuff is not tested, otherwise the original problem would have been long caught. I also don't know how/where to test it, the project is pretty big and the test code somewhat messy.

S15:
An overly simple repro is is:
```javascript
const fakeRef = { toString() { return "file://base"; } };

messageError("SchemaFetchError", {
  address: "http://example.com/missing",  // plain string
  base: fakeRef,                           // object with toString()
});
```
```
QuickTypeError: Could not fetch schema , referred to from file://base
    <...>
  errorMessage: 'Could not fetch schema ${address}, referred to from ${base}',
  messageName: 'SchemaFetchError',
  userMessage: 'Could not fetch schema , referred to from file://base',
  properties: {
    address: 'http://example.com/missing',
    base: { toString: [Function: toString] }
  }
```

Any schema with an external reference that doesn't exist should trigger this, since address is just a string.

Issue #2753, opened 2025-05-19
[BUG]: "Option parsing failed: ." when providing input/src file as positional argument


Braincompiler:
Running `quicktype` with:

```cli
npx quicktype src/systemregler.schema.json --src-lang schema --top-level Configuration --lang python --out src/parameters/types.py
```
or
```cli
npx quicktype --src-lang schema --top-level Configuration --lang python --out src/parameters/types.py src/systemregler.schema.json
```
... fails with the meaningless error: `Option parsing failed: .` After some digging in I found out the problem seems to be the `messageError` function at https://github.com/glideapps/quicktype/blob/master/src/index.ts#L731 or the `exceptionToString` function from the `ts-necessities` package. The exception message was actually "Unknown value: src/systemregler.schema.json" which is the name of the input file as a positional argument. Which in turn seems to be another bug, since according to the README, positional arguments for the input are allowed: https://github.com/glideapps/quicktype?tab=readme-ov-file#using-quicktype. If not, the documentation needs to be updated.

It was working before, so maybe the bug is a regression with the latest update.

## Issue Type

input parsing

## Context (Environment, Version, Language)

Input Format: JSONSchema
Output Language: Python

CLI, npm, or app.quicktype.io: npm
Version: 23.2.5

## Description

Running `quicktype` with:

```cli
npx quicktype --src-lang schema --top-level Configuration --lang python --out src/parameters/types.py --src src/systemregler.schema.json
```

... workarounds the problem and quicktype is working as expected.

## Expected Behaviour / Output

Positional arguments should work again or the error needs to be more meaningful.

## Current Behaviour / Output

Meaningless error: `Option parsing failed: .`

## Steps to Reproduce

1. Run quicktype with positional argument instead of `--src` option.

stewartadam:
It's not limited to positional arguments either it seems, I'm seeing this error whenever using `--just-types`.

0gust1:
We were hit by this one too, and to contrast with OP's message, suppressing positional args doesn't help.

The command

`npx quicktype --src-lang schema --lang ts --top-level MyTypeName --prefer-unions --no-runtime-typecheck --no-runtime-typecheck-ignore-unknown-properties --src ./MySchema.schema.json --out ./mytype.generated.types.ts`

stills ends up with `Error: Option parsing failed: .`

I tried to wrap all arguments values with quote, but still the same message.

RyanGaudion:
Same issue here running:

[...]

PR #2792, opened 2025-06-24
chore: Add unit tests for options functions and cli

inferrinizzard:
## Description

- split top level cli bin index into multiple files #2762
- add unit tests for options parsing and inference
- generate unit tests for `makeQuicktypeOptions`
- tweak flags boolean parsing
- rename misc

## Related Issue

#2753 

## Motivation and Context

Previously testing coverage for CLI functionality was lacking

N/a

## New Behaviour / Output

image

## How Has This Been Tested?

`vitest` locally, vitest version `^3.1.4`


schani:
Thanks for this — splitting `src/index.ts` into focused modules is a direction we want, and the new unmocked `test/cli/options.test.ts` is genuinely useful. A few things need fixing before this can merge:

**1. Inference flags default to OFF for every invocation (functional bug).** In `src/inference.ts`, `inferCLIOptions` sets `options[flagName] = !opts[cliName]` (e.g. `inferMaps = !opts.noMaps`). Then `makeQuicktypeOptions` in `src/quicktype.options.ts` picks up that already-computed `options[flagName]` as `positiveValue` and negates it again (`quicktypeOptions[flagName] = !value`). Net effect: every inference flag (maps, enums, dates, uuids, integer-strings, …) is inverted from what the user asked for, and defaults to disabled. This is exactly what broke branch CI on `schema-java`: `uuid.1.fail.uuid.json` stopped failing because UUID inference was silently off, which then fail-fast-cancelled most of the other language jobs.

**2. `test/cli/quicktype.options.test.ts` mocks `quicktype-core` wholesale**, including a hand-rolled `inferenceFlagNames: ["inferMaps", "inferEnums"]` and a `negatedInferenceFlagName` mock that produces `noInferMaps` — not the real convention (`noMaps`, per `negatedInferenceFlagName` in `src/utils.ts`). Because it never exercises the real `inferCLIOptions` → `makeQuicktypeOptions` path, it validated the mock rather than the pipeline, and masked bug 1 completely. Needs rewriting against real flag names/values, unmocked like `test/cli/options.test.ts`.

**3. `test/languages.ts`**: `quickTestRendererOptions` is narrowed from `(RendererOptions | [string, RendererOptions])[]` to `RendererOptions[]`, dropping the pinned-input tuple form current master relies on.

**4. Real conflicts with current master** in 8 files (`package.json`, `package-lock.json`, `src/GraphQLIntrospection.ts`, `src/index.ts`, `test/languages.ts`, `test/lib/deepEquals.ts`, `test/lib/multicore.ts`, `test/test.ts`). Master's `src/index.ts` has since gained fixes (boolean `--help` type-label fix, int-range-aware number inference, Biome cleanups) that aren't reflected in the new split modules (`usage.ts`, `input.ts`/`sources.ts`) — since this PR relocates that file, resolving means effectively redoing the split on top of current master.

**5. This overlaps almost entirely with your other open PR, #2761**, which does the same CLI refactor/split. We'd rather not merge both — could you consolidate into a single PR?

Given the overlap and the master drift, would you be up for redoing this as one refactor branched off current master, folding in #2761? The bug-1 fix and the mock-test rewrite are hard requirements, and we'll need a green fixture-test CI run (all languages, not fail-fast-cancelled) before merge.

Issue #2796, opened 2025-06-27
[FEATURE]: Improve C++ std::runtime_error message for failed enum deserialization


cconverse711:
Currently, a vague exception message is provided ("Input JSON does not conform to schema!"), which leaves no useful guidance for what part of the JSON does not conform. Providing the enum name, and possibly listing the possible options would help users fix the non-conformant json.

## Context (Input, Language)

Input Format: schema
Output Language: C++

## Description

I would like to more easily indicate to the user what part of the JSON failed to deserialize. For example, if using a json config file for an application, it would be helpful to indicate which part is does not conform to the schema, so that they can fix it more easily. 

## Current Behaviour / Output

std::runtime_error is raised with message `Input JSON does not conform to schema!`

## Proposed Behaviour / Output

std::runtime_error raised with message like `Cannot deserialize to enumeration "FooBar".`
It may also be helpful to include a comma-separate list of the actual allowed enum options, but that may be unwieldy if there are many.

## Solution

Adjust the exception error message to include enumeration name, and possibly enum options.

schani:
The generated C++ `from_json` code now throws `std::runtime_error("Cannot deserialize to enumeration \"FooBar\"")` when enum conversion fails. This applies to both ordinary enums and the map-based path used for large enums; I also verified the generated code at runtime with an invalid value. Regenerating with a current quicktype release provides the improved diagnostic.

PR #2797, opened 2025-06-27
Add enum name to default case deserialization exception for C++

cconverse711:
## Description

Includes enumeration name in std::runtime_error message for enum deserialization failure.

## Related Issue

#2796 

## Motivation and Context

#2796 

std::runtime_error is raised with message `Input JSON does not conform to schema!`

## New Behaviour / Output

std::runtime_error raised with message like `Cannot deserialize to enumeration "FooBar"`.

## How Has This Been Tested?

Existing C++ unit tests


schani:
I've resolved the merge conflict by merging `master` into this branch (no rebase/force-push, so the original commit is untouched).

**What conflicted:** #2846 moved the enum `from_json` throw out of the small-enum `if/else` branch so a single `else { throw ... }` now covers both the small-enum (if/else chain) and the new large-enum (>15 cases, map-based lookup) code paths — the same lines this PR edits.

**Resolution:** kept #2846's structure (single throw after the if/else) and applied this PR's improved message. As a result, `Cannot deserialize to enumeration "<EnumName>"` is now thrown for **both** small and large enums — extending the PR's intent to the code path #2846 added.

**Verified:**
- Generated C++ from `test/inputs/schema/enum-large.schema` (which has both a 20-case enum `Callsign` and a 3-case enum `Priority`); both `from_json` functions emit the new throw in the right place:
  - large: `else { throw std::runtime_error("Cannot deserialize to enumeration \"Callsign\""); }` (after the map lookup)
  - small: `else { throw std::runtime_error("Cannot deserialize to enumeration \"Priority\""); }` (after the if/else chain)
- Compiled the generated header with `g++ -std=c++17` against nlohmann/json and ran it: valid values parse correctly, and unknown values throw with exactly the new message for both enum sizes.
- `tsc --noEmit` on `quicktype-core` passes; biome shows no new findings on the touched file vs master.

**Test note:** the fixture harness's fail samples (`callAndExpectFailure` in `test/utils.ts`) only assert that parsing *fails*, not the exception message content, so the message itself can't be pinned by a fixture test with the current infra. The existing `enum-large.1.fail.enum.json` / `enum-large.2.fail.enum.json` samples do exercise both throw paths (bad callsign → large-enum path, bad priority → small-enum path) and continue to pass.

schani:
CI note: the failed run was an infrastructure flake, not this PR — the cjson fixture's setup downloads `hashtable.h` from raw.githubusercontent.com with a bare `curl` (no `--fail`, no retries); the runner got HTTP 429 and the literal error body was saved as the header and compiled (`hashtable.h:1:1: error: ... 429: Too Many Requests`). Every other red job was a fail-fast cancellation, and master runs from the same hour are green. Failed jobs have been re-run.

PR #2805, opened 2025-07-11
fix(C#): Enhance number handling in SystemTextJsonCSharpRenderer

### LINKED ISSUE #2804: [BUG]: quicktype output bug (C# System.Text.Json renderer)
## Issue Type

quicktype output bug (C# System.Text.Json renderer)

Input Format: JSON
Output Language: C# (System.Text.Json)
CLI, npm, or app.quicktype.io: Any
Version: master (latest)

## Description

When generating C# code for schemas with unions of integer and double (e.g., `integer | double`), the generated deserializer did not correctly distinguish between integer and floating-point JSON numbers. Both were handled as separate cases, but both matched the same `Number` token, causing only one branch to be taken and the other ignored.

## Input Data

```json
[1, 2.5, 3]
```
or any schema with a union of integer/double.

The generated C# code should correctly distinguish between integer and floating-point values in a union, deserializing them to the correct type.

## Current Behaviour / Output

Both integer and double union branches matched the same `Number` token, so only one branch was ever taken, leading to incorrect deserialization (e.g., all numbers treated as one type).

## Steps to Reproduce

1. Generate C# code with quicktype for a schema with a union of integer and double.
2. Attempt to deserialize a JSON array with both integer and floating-point numbers.
3. Observe that only one type is handled.

## Possible Solution

The fix is to check, when both integer and double transformers are present, whether the JSON number is an integer (using `reader.TryGetInt64`) or a double, and dispatch accordingly. This was implemented by:

- Adding a check for both transformers.
- Using `reader.TryGetInt64` to distinguish integer from double.
- Dispatching to the correct transformer for each case.

mikecrowe-otp:
## Description

* Added support for handling both integer and double types in a union case.
* Introduced a variable generator to manage variable naming dynamically.
* Updated decoder transformer cases to accommodate the new logic for number types.
* Ensured backward compatibility for existing single type handling.

## Related Issue

Fixes #2804

## Motivation and Context

C# code would not compile

```log
error CS0152: The switch statement contains multiple cases with the label value '8'
```

## New Behaviour / Output

Code compiles

## How Has This Been Tested?

In our production code


schani:
Hi @mikecrowe-otp — I'm the repo maintainer's assistant. Thanks for the fix! I've pushed a regression-test commit on top of your branch (900977d4), since the PR didn't include a test that would guard against this bug coming back.

**Why no existing test caught this:** the schema fixtures in `test/inputs/schema` were only ever run through the Newtonsoft C# language; `csharp-SystemTextJson` only ran the JSON-input fixture, and JSON inference can never produce an integer|double union. So the converter code path you fixed was never compiled in CI — even though the existing `integer-float-union.schema` already triggers the bug.

**What the commit does:**
- Adds a `JSONSchemaFixture` for `CSharpLanguageSystemTextJson` (`schema-csharp-SystemTextJson`) and enables it in the CI matrix, so every schema fixture is now generated, compiled, and round-tripped with `--framework SystemTextJson`.
- Adds `union-int-double.schema` (an `anyOf` integer|number|string union) with samples for an integer, a fractional number, a large integer, and a string, exercising your `TryGetInt64` dispatch at runtime.
- Skips seven schemas that hit *pre-existing* System.Text.Json renderer bugs unrelated to your fix (each documented next to its skip in `test/languages.ts`): Newtonsoft-style code emitted for transformed string types (CS0103), a `JsonSerializer` naming collision (CS0120), CS8602 build warnings polluting the test driver's stdout, and `check-required` not being implemented.

**Before/after evidence (verified locally with the dotnet 6 SDK):**
- At the merge-base (without your fix), the generated converter for the new fixture contains two `case JsonTokenType.Number:` labels and fails to compile: `error CS0152: The switch statement contains multiple cases with the label value '8'`. Running the new fixture through the actual test harness at the merge-base fails with exactly that error.
- With your fix, the generated converter has a single `Number` case with the `TryGetInt64` dispatch; it compiles cleanly and round-trips all four samples (`123`, `3.141592`, `123456789012345`, `"hello"`).
- The full `schema-csharp-SystemTextJson` fixture (48 schemas after skips) passes end to end on this branch, and the Newtonsoft output for the same fixture is byte-identical before/after your change (distinct `JsonToken.Integer`/`JsonToken.Float` cases, as expected).

schani:
Diagnosed the CI failure: the only real failure was `schema-cjson` (the other 12 red jobs were fail-fast cancellations). The CJSON renderer has the very same duplicate-number-dispatch bug this PR fixes for System.Text.Json — its `from_json` emits `if (cJSON_IsNumber)` for the integer member and an unreachable `else if (cJSON_IsNumber)` for the double member, so `3.141592` round-trips as `3` and the harness diff fails on `union-int-double.2.json`. CJSON already skips `integer-float-union.schema` for exactly this reason; the new schema now gets the same skip (plus Pike, cosmetically).

Also merged current master into the branch (clean merge). Locally verified on the merge: the STJ fixture passes all four `union-int-double` samples plus the schemas recently added on master, Newtonsoft output is unchanged, and C++/Python/Rust were spot-checked green — so CI should pass now.

schani:
Second CI round diagnosed: only the csharp job genuinely failed (the other red jobs were fail-fast cancellations — rust just started late before being cancelled, and `union-int-double.schema` itself PASSED under the new STJ fixture). The failure was `optional-constraints.schema`, which landed on master after the STJ fixture was added here and hits two pre-existing STJ renderer gaps unrelated to this PR: a CS8602 warning that `dotnet run` prints to stdout ahead of the JSON (same issue already documented for `minmaxlength.schema`), and missing integer min/max + optional-string pattern checks so expected-failure samples don't fail. Pushed skips for it and for `optional-const-ref.schema` (verified locally it would fail next for the same reasons); the other five master-added schemas (`enum-large`, `comment-injection*`) were locally verified green under the STJ fixture, so this should be the last round. The underlying STJ constraint gaps deserve their own issue.

PR #2828, opened 2025-09-09
Allow build for web

### LINKED ISSUE #2763: [BUG]: Module build failed: UnhandledSchemeError: Reading from "node:fs" is not handled by plugins (Unhandled scheme).
in my react web application project, i upgrade quicktype-core@23.2.5, build the project, has a error

```
fatal - Error: ERROR in node:fs
Module build failed: UnhandledSchemeError: Reading from "node:fs" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
```

vexdev:
## Description

Revert imports from node:fs to fs to allow the package.json directive 
```
"browser": {
  "fs": false
}
```
to prevent the fs import to be packaged when building for web.

## Related Issue

Fixes #2763

## Motivation and Context

Fixes #2763

Builds for web were failing, unable to find node:fs

## New Behaviour / Output

Builds working as expected

## How Has This Been Tested?

Run a build in a sample react project with webpack


schani:
I pushed a small regression-guard commit on top of this branch (f2083b32) so the `node:fs` breakage can't sneak back in.

**What it does:** `test/check-no-node-imports.ts` scans `packages/quicktype-core/src` for `node:`-prefixed imports/requires and fails with an explanation of why they're forbidden there: quicktype-core's `"browser": { "fs": false }` package.json field only stubs the *bare* `fs` specifier, so `node:fs` breaks webpack/browserify bundles (#2763) even though the code is identical at runtime under Node. The check is scoped to quicktype-core only — the Node-only CLI in the root `src/` keeps using `node:` imports freely.

**Where it runs in CI:** it's called at the top of `main()` in `test/test.ts`, so every "Test PR" matrix job executes it (`npm test` → `script/test` → `ts-node test/test.ts`) before running any fixture. No workflow files were touched.

**Verified fail/pass:**
- With `NodeIO.ts` restored to the merge-base state (`import * as fs from "node:fs"`), both the standalone check and `FIXTURE=... npm test` exit 1:
  ```
  error: found "node:"-prefixed imports in packages/quicktype-core/src:

      packages/quicktype-core/src/input/io/NodeIO.ts:1: import * as fs from "node:fs";
  ```
- On this branch (with your fix), both exit 0.

Also worth noting: biome's `useNodejsImportProtocol` lint rule is what pushes code toward `node:` imports, so an automated fix could easily reintroduce this — hence a hard guard rather than relying on review.

schani:
Merged current master into this branch to resolve the conflict that appeared when #2780 (relative symlink support) landed. Worth noting: #2780 added an `import * as path from "node:path"` to `NodeIO.ts` — which reintroduces exactly the web-bundler breakage this PR fixes (the `browser` field can only stub bare specifiers), and the guard added in the previous commit caught it immediately. The resolution keeps bare `fs`/`path` imports, and a follow-up commit stubs `path` in the `browser` field alongside `fs` (it's only used under `isNode`), so web builds keep working after the symlink fix. Verified locally: the no-`node:`-imports guard passes and `tsc --noEmit` on quicktype-core is clean.

Issue #2629, opened 2024-07-05
[FEATURE]: CSharp - make emitting DateOnlyConverter & TimeOnlyConverter optional


Thomas-Avery:
Add a `rendererOptions` section for C# to exclude the generated `DateOnlyConverter` and `TimeOnlyConverter` when using the `System.Text.Json` serialization framework.

For C# projects targeting the [.NET Standard](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0) `DateOnly` and `TimeOnly` can't be used since these types are only available in .NET 6+.

## Context (Input, Language)

Input Format:
 JSON Schema

Output Language:
C#

## Description

When generating C# code from a JSON schema, I would like to be able to use the C# code in a .NET library targeting the .NET Standard while using the`System.Text.Json` serialization framework.

.NET Standard provides wide support for consumers of a library (.NET Framework 4.6.1+ and .NET & .NET Core 2.0+)

## Current Behavior / Output

Go to https://app.quicktype.io/

Choose the JSON schema pokedex-schema.json example.

Select the following language options: 

- Language: C#
- Serialization framework: System Text Json
- Generated namespace: QuickType

Create a .NET library project

```
dotnet new classlib --framework netstandard2.0
dotnet add package System.Text.Json
```

Copy over the C# code from QuickType.

Build the library 

```
dotnet build
```

Note build errors.
```
Error CS0246 : The type or namespace name 'DateOnly' could not be found (are you missing a using directive or an assembly reference?)
Error CS0246 : The type or namespace name 'TimeOnly' could not be found (are you missing a using directive or an assembly reference?)
```

## Proposed Behavior / Output

Go to https://app.quicktype.io/

Choose the JSON schema pokedex-schema.json example.

Select the following language options: 

- Language: C#
- Serialization framework: System Text Json
- Generated namespace: QuickType
- **Support .NET Standard: true**

Create a .NET library project

```
dotnet new classlib --framework netstandard2.0
dotnet add package System.Text.Json
```

Copy over the C# code from QuickType.

Build the library 

```
dotnet build
```

The .NET library builds successfully. 

## Solution

Add a language configuration option to indicate if the output should support .NET Standard.

When that configuration option is selected, the `DateOnlyConverter` and `TimeOnlyConverter` are not emitted.

## Alternatives

[...]

PR #2842, opened 2025-10-27
feat(csharp): add options to disable DateOnly and TimeOnly generation for System.Text.Json

findyoucef:
## Description

Adds two new C# renderer options to control whether `DateOnly` and `TimeOnly` types are emitted when generating System.Text.Json code. Also updates the renderer to conditionally register their converters and refreshes `ajv` and `esbuild` dependencies.

## Related Issue

https://github.com/glideapps/quicktype/issues/2629

## Motivation and Context

Projects targeting .NET Standard lack `DateOnly` and `TimeOnly`, causing compilation errors in generated code. This change adds `--no-dateonly` and `--no-timeonly` options so users can disable these types for better cross-platform compatibility.

Generated code always included:
```csharp
public DateOnly StartDate { get; set; }
options.Converters.Add(new DateOnlyConverter());
```
Which fails on .NET Standard.

## New Behaviour / Output

With flags:
```bash
quicktype --lang csharp --framework system-text-json --no-dateonly --no-timeonly
```
Output now omits `DateOnly`/`TimeOnly` and related converters:
```csharp
public DateTime StartDate { get; set; }
// No DateOnlyConverter registration
```

## How Has This Been Tested?

- Verified .NET 7 (default behavior) and .NET Standard 2.0 (flags enabled) compile successfully.  
- Generated C# code compatible with .NET Standard by disabling unsupported `DateOnly`/`TimeOnly` types.
---

### Summary

Adds `no-dateonly` and `no-timeonly` C# options for System.Text.Json codegen.

### Impact

Allows generating C# code compatible with .NET Standard by disabling unsupported `DateOnly`/`TimeOnly` types.


schani:
Thank you @findyoucef for this contribution — the underlying problem (the System.Text.Json renderer's unconditional `DateOnlyConverter`/`TimeOnlyConverter` emission breaking compilation on .NET Standard / pre-.NET 6 targets, #2629) is real, and your conditional-emission approach is the right fix.

Unfortunately this branch could no longer be rebased onto master: it conflicts with the C# 8 default work that has landed since, and it carries unrelated dependency churn (`package.json`/`package-lock.json`, `.vscode`) that made the diff unmergeable as-is.

I've opened #2962 as a successor that reimplements the live core of your change on current master, with your authorship credited via a `Co-authored-by` trailer on the commit. Leaving this PR open for reference until the successor lands.

Issue #2843, opened 2025-10-29
[BUG]: generated php code is re-declaring functions on global level


PssbleTrngle:
when having multiple properties with `"type": "object"`, quicktype will generate PHP code with local `from()` and `to()` functions. Since the `function` always lives in the global scope in PHP, calling this twice will throw a redeclaration exception.

## Issue Type

output

## Context (Environment, Version, Language)

Input Format: JSON Schema
Output Language: PHP

CLI, npm, or app.quicktype.io:
Version: npm@10.9.3

## Description

When generating a PHP file from a JSON Schema, that contains two properties with simply the type `object`, but not defined properties, quicktype does not generate a class for them but simply inline the generated `from()` and `to()` functions within the block to directly call them afterwards.

Since in PHP the `function` keyword always creates a function on the global (or namespace) scope, calling either of these "wrapping" functions that define the `from()` function twice, or simply ones for two properties will fail with `Cannot redeclare to()`.

You probably expected that the defined `function to()` does only live within the scope of the parent function it's defined in, as is the case in a lot of language, but this is not how it works in PHP. It works more similar to how the `var` keyword works in JavaScript.

Instead of inlining these named function `from` and `to` here, only to call them directly afterwards, it would make more sense (and actually work without throwing an error) to directly create the `\stdClass` here and return it.
An example of how this would look like has been provided under "Expected Behaviour/Output".

## Input Data

```json
{
  "$schema": "http://json-schema.org/draft-06/schema#",
  "type": "object",
  "properties": {
    "a": {
      "type": "object"
    },
    "b": {
      "type": "object"
    }
  }
}
```

## Expected Behaviour / Output

Not defining the function `from` here, since it does not as you probably expected only live

```php
/**
     * @throws Exception
     * @return ?stdClass
     */
    public function toB(): ?stdClass {
        if (Coordinate::validateB($this->b))  {
            if (!is_null($this->b)) {
                $out = new stdClass();
                foreach ($my as $k => $v) {
                    $my->$k = $v; /*any*/
                }
                return $out;
            } else {
                return  null;
            }
        }
        throw new Exception('never get to this Coordinate::b');
    }
```

## Current Behaviour / Output

```php
/**
     * @throws Exception
     * @return ?stdClass
     */
    public function toB(): ?stdClass {
        if (Coordinate::validateB($this->b))  {
            if (!is_null($this->b)) {
                function to($my): stdClass {
                    $out = new stdClass();
[...]

PR #2844, opened 2025-10-29
Directly return cloned object without defining nested from/to functions

PssbleTrngle:
## Description

As described in the linked issue, the code that is currently generated for simple `object` properties is invalid.
Instead of emitting the `to()` and `from()` just to call them immediately, I have opted to simply build and return the created `$out` object.

## Related Issue

#2843

## How Has This Been Tested?

I was unable to correctly setup the project environment and could not test the new output.


PssbleTrngle:
sorry, it seems like I also staged the two changes in the github workflow files, which bump the `ubuntu` version to latest, since the one you are using currently is no longer supported.
I can remove those changes again if you want to, but how you have it currently these workflows cannot be executed anymore

schani:
I pushed three commits to this branch to get it mergeable and covered by CI (thanks @PssbleTrngle for the fix and the excellent analysis in #2843):

**1. `51cfa6e` — merge of `master`, resolving the conflict.** The conflict came entirely from the incidental CI workflow edits this branch carried (`ubuntu-22.04` → `ubuntu-latest` in `master.yaml`/`test-pr.yaml`); master has since split and rewritten those workflow files. Since those edits are unrelated to the PHP fix, I dropped them and took `.github/workflows/*` wholesale from master. The `PhpRenderer.ts` fix itself merged cleanly and is unchanged.

**2. `7e93291` — regression test.** Added `test/inputs/json/priority/nested-objects.json` with two properties that quicktype infers as maps (bare objects → `stdClass` in PHP), mirroring the two-`"type": "object"`-properties schema from #2843, and whitelisted it in PHP's `includeJSON` so the PHP CI fixture roundtrips it through `test/fixtures/php/main.php`.

Evidence (php:8.2-cli, driver = `test/fixtures/php/main.php`):
- **Before** (master's renderer): `Fatal error: Cannot redeclare from() (previously declared in TopLevel.php:24) in TopLevel.php on line 95`
- **After** (this PR's fix): roundtrip output is byte-identical to the input.

I also reran all of PHP's other `includeJSON` samples (`bitcoin-block`, `pokedex`, `simple-object`, `getting-started`, `uuids`) through the driver with the fixed renderer — all pass (pokedex differs only by omitted `null`s, which the harness allows via `allowMissingNull`). The new input generates fine for other languages that run priority samples (spot-checked go/csharp/typescript/rust/python/java/kotlin/swift; e.g. Go infers `map[string]int64` for both properties).

**3. `ee7d5a2` — same fix for `phpSampleConvert`.** The nested `function sample(): stdClass` had the identical global-redeclaration problem: with the two-map-property input, `TopLevel::sample()` fataled with `Cannot redeclare sample()` even after this PR's from/to fix. Applied the same inlining there. Verified: `TopLevel::sample()` can now be called repeatedly and emits sane bodies (`$out = new stdClass(); ...; return $out;`). Note the CI driver never calls `sample()`, so this path isn't covered by the fixture test — hence the manual verification.

Checks: `tsc --noEmit` passes for `quicktype-core` and the test project; `biome check` on the touched files reports the same pre-existing findings as master (none new).

One pre-existing limitation worth noting (not made worse by this PR): for maps *of maps*, both the old and new code ignore the assignment target in the recursive case, so nested map-of-map values were and remain broken — that's a separate issue from the redeclaration fatal fixed here.

PR #2845, opened 2025-10-31
feat(kotlin): Support for date and datetime from JSON schema

### LINKED ISSUE #2460: Kotlin date and datetime types are generated as strings from json schema
Kotlin generated data classes use string instead of date or datetime
Here is the part of the schema: 
      "properties": {
          "PlannedDateTime" : {
          "type" : "string",
          "format" : "date-time",
          "examples" : [ "2020-02-03T08:00:00.000+01:00" ],
          "description" : "When the planing process is finished, this timestamp is set and never changed afterwards (frozen value). It is expressed as a local time plus timezone offset, so UTC can be derived by a simple calculation."
        },
        "pokemon": {
          "type": "array",
          "description": "All pokémon contained in the pokédex",
          "items": {
            "$ref": "#/definitions/Pokemon"
          }
                        
        }
      },

Generated result is: 

data class PokedexSchema (
    /**
     * When the planing process is finished, this timestamp is set and never changed afterwards
     * (frozen value). It is expressed as a local time plus timezone offset, so UTC can be
     * derived by a simple calculation.
     */
    val plannedDateTime: String,

    /**
     * All pokémon contained in the pokédex
     */
    val pokemon: List<Pokemon>
)

nikhilunni:
## Description
Add support for JSON Schema `format: date-time`, `format: date`, and `format: time` in the Kotlin code generator. These formats now generate proper `java.time.*` types instead of `String`.

This implementation follows the same pattern as Java's date-time support, using:
- `OffsetDateTime` for `format: "date-time"`
- `LocalDate` for `format: "date"`
- `OffsetTime` for `format: "time"`

The feature works across all Kotlin frameworks: `just-types`, `jackson`, `klaxon`, and `kotlinx`.

## Related Issue
Fixes #2460

## Motivation and Context
We prefer Kotlin generation to Java generation for nullability propagation, but did not support datetime formatting, as Java does. Plus, saw there was a longstanding issue for this.

For the given JSON schema:
```json
{
  "type": "object",
  "properties": {
    "eventDate": {
      "type": "string",
      "format": "date"
    },
    "eventTime": {
      "type": "string",
      "format": "time"
    },
    "plannedDateTime": {
      "type": "string",
      "format": "date-time"
    }
  }
}
```

It would output:
```kotlin
package quicktype

data class MySchema (
    val eventDate: String,
    val eventTime: String,
    val plannedDateTime: String
)
```

## New Behaviour / Output
```kotlin
package quicktype

import java.time.OffsetDateTime
import java.time.LocalDate
import java.time.OffsetTime

data class MySchema (
    val eventDate: LocalDate,
    val eventTime: OffsetTime,
    val plannedDateTime: OffsetDateTime
)
```

### Manual Testing

- Tested with the existing test/inputs/schema/date-time.schema file
- Verified output for all Kotlin frameworks:
  - `--framework just-types`
  - `--framework jackson`
  - `--framework klaxon`
  - `--framework kotlinx`
- Tested with the exact schema from issue #2460
- Verified proper handling of:
  - Simple date-time properties
  - Date-time types in unions
  - Date-time types in arrays
  - Complex nested structures with date-time types

### Automated Testing
- Added "date-time" feature to KotlinLanguage test configuration
- Added "date-time" feature to KotlinJacksonLanguage test
configuration
- Follows the same pattern as the Golang date-time feature
implementation (commit 1f89a971)


nikhilunni:
Could somebody please take a look at this?

schani:
Update on the CI failures: the fresh run on 379134eb confirmed two defects in the feature, and while fixing them we found a third:

1. **Klaxon generation crash** (masked in CI by fail-fast): `matchType` in `KotlinKlaxonRenderer.ts` is missing the `transformedStringType` matcher, so generating from `test/inputs/schema/date-time.schema` dies with "Unsupported type date-time in non-exhaustive match".
2. **Jackson runtime failure** (the one CI shows): the generated mapper can't deserialize `java.time.LocalDate` — nothing handling `java.time` is registered.
3. **Duplicate union guards** in both renderers: `when` branches like `is TextNode -> ...` repeated per date type, so only the first union member could ever parse.

We've prepared verified fixes for all three (round-trip byte-identical for both klaxon and jackson locally, fail-case correctly rejected, no regression on non-date output): **nikhilunni/quicktype#1** targets your `master`, so merging it on your side updates this PR and should turn kotlin CI green. Cherry-picking works too (branch: `glideapps/quicktype:kotlin-datetime-fixes-2845`) — or say the word and we'll push directly to this PR.

test/languages.ts
```diff
@@ -1156,6 +1156,17 @@ export const KotlinLanguage: Language = {
         "af2d1.json",
         "32431.json",
         "bug427.json",
+        // Contain date-time values with trailing zeros in the fractional
+        // seconds, which java.time does not re-serialize identically
+        "54d32.json",
```
schani: We should not detect unserializable date-times! Don't we already have a facility for that?

schani:
Thank you so much, both for your contribution and your patience!

PR #2846, opened 2025-11-20
Fix: throw error for undefined enum for both small and large enum in C++

HavardNJ-Laerdal:
Current version throws for an unknown enum if the enum is considered small and handled with if-else
```
    inline void from_json(const json & j, Extrasystole & x) {
        if (j == "coupled_pvc") x = Extrasystole::CoupledPvc;
         ...
        else { throw std::runtime_error("Input JSON does not conform to schema!"); }
```

For a large enum implemented with map, the throw is missing
```
    inline void from_json(const json & j, BasicRhythm & x) {
        static std::unordered_map<std::string, BasicRhythm> enumValues {
            {"sinus", BasicRhythm::Sinus},
            ...
        };
        auto iter = enumValues.find(j.get<std::string>());
        if (iter != enumValues.end()) {
            x = iter->second;
        }
}
```
resulting in the property being a valid value which is inconsistent with the json.

This change adds the throw at the end of the large enum.
```
        ....
        if (iter != enumValues.end()) {
            x = iter->second;
        }
        else { throw std::runtime_error("Input JSON does not conform to schema!"); }
```

### DIFF
--- a/packages/quicktype-core/src/language/CPlusPlus/CPlusPlusRenderer.ts
+++ b/packages/quicktype-core/src/language/CPlusPlus/CPlusPlusRenderer.ts
@@ -1986,10 +1986,10 @@ export class CPlusPlusRenderer extends ConvenienceRenderer {
                         );
                         onFirst = false;
                     });
-                    this.emitLine(
-                        'else { throw std::runtime_error("Input JSON does not conform to schema!"); }',
-                    );
                 }
+                this.emitLine(
+                    'else { throw std::runtime_error("Input JSON does not conform to schema!"); }',
+                );
             },
         );
         this.ensureBlankLine();
--- a/test/inputs/schema/enum-large.1.fail.enum.json
+++ b/test/inputs/schema/enum-large.1.fail.enum.json
@@ -1,0 +1,4 @@
+{
+  "callsign": "zulu",
+  "priority": "low"
+}
--- a/test/inputs/schema/enum-large.1.json
+++ b/test/inputs/schema/enum-large.1.json
@@ -1,0 +1,4 @@
+{
+  "callsign": "alpha",
+  "priority": "low"
+}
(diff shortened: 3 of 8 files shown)


schani:
I pushed a regression test on top of this branch (380f27e7), since the repo had no fixture exercising an enum above the 15-case `isLargeEnum` threshold — which is why this bug went unnoticed.

**What the test adds**

- `test/inputs/schema/enum-large.schema`: an object with a 20-case enum (`callsign`, takes the map-lookup path this PR fixes) and a 3-case enum (`priority`, takes the if/else path).
- Three valid samples covering first/middle/last cases, which must round-trip.
- Two failure samples gated to the `enum` test feature:
  - `enum-large.1.fail.enum.json` — unknown value for the **large** enum. This is the actual bug.
  - `enum-large.2.fail.enum.json` — unknown value for the **small** enum, as a regression control for the path that already threw.

**Before/after evidence** (generated C++ compiled with `g++ -std=c++17` against nlohmann/json):

- At this branch's merge-base (without the fix), `{"callsign": "zulu", ...}` is accepted and the behavior is undefined — in one build it silently round-tripped as `{"callsign": "alpha", ...}` (uninitialized enum happened to be 0), in another it crashed later in `to_json` with garbage value 22993. The unknown small-enum value was correctly rejected.
- With this PR's fix, both failure samples deterministically throw `Input JSON does not conform to schema!` in `from_json`, and all valid samples still round-trip. The `schema-cplusplus` fixture passes end-to-end locally.

**Gating decision**: the fail samples run for every language that declares the `enum` feature and doesn't skip the schema. I mirrored the existing `enum.schema` skip list exactly (cjson, crystal, dart, haskell, scala3, swift, typescript-zod, typescript-effect-schema — languages that skip it because they don't validate enum values or have other enum issues). No renderer other than C++ special-cases enums by size, so any language already rejecting the existing `enum.1.fail.enum.json` rejects these too. I verified the `schema-typescript` and `schema-golang` fixtures pass locally, and that the generated Python driver rejects both fail samples.

Issue #2850, opened 2025-12-02
[BUG]: JSON to Java : Quicktype fails to generate the expected enum value for a given string - super weird


NeoAyush:
## Issue Type

This issue is with quicktype output

## Context (Environment, Version, Language)

Version : Eclipse Temurin JDK 21.0
Language : java

Input Format: json
Output Language: Java

CLI, npm, or app.quicktype.io: npm
Version: 23.2.6

## Description

We have a business requirement where we define an enum in a json file. I had a task to enhance this enum by adding another enum constant in it with the value "MULTI_SPA_IN_GROUP_REJECTED", however the generated Enum in java does some wierd case conversion and generates - "MULTI_Spa_IN_GROUP_REJECTED". This only happens for the substring - "SPA", it is getting automatically coverted to "Spa" ( loosing the original upper casing) which is super weird as the other enum value generations are working as expected, without any messing up of the casing.

## Input Data

`"MessageCode": {
      "title": "MessageCode",
      "type": "object",
      "required": [
        "code"
      ],
      "properties": {
        "code": {
          "enum": [
               "SOME_OTHER_VALUE",
            **"MULTI_SPA_IN_GROUP_REJECTED"** [newly added]
          ],
          "type": "string",
          "description": "the code of the message",
          "examples": [
            "ITEM_REMOVED"
          ]
        }
      }
    }`

## Expected Behaviour / Output

The enum constants that contain the string "SPA" SHOULD NOT LOOSE the case in the generated code and show up as defined in the json schema definition.

`public enum MessageCode {
   SOME_OTHER_VALUE, MULTI_SPA_IN_GROUP_REJECTED;

    @JsonValue
    public String toValue() {
        switch (this) {
            case SOME_OTHER_VALUE: return "SOME_OTHER_VALUE";
            case MU_SPA: return "MULTI_SPA_IN_GROUP_REJECTED";
        }
        return null;
    }`

## Current Behaviour / Output

The enum constants that contain the string "SPA" LOOSES the case in the generated code and does not show up as defined in the json schema definition.

For the quicktype command, we are using below options:
`        
            --lang=java \
            --array-type=list \
            --datetime-provider java8 \
            --lombok-copy-annotations \
            --lombok \
            --alphabetize-properties \
            --acronym-style=camel
`

Currently generated java code :  

`public enum MessageCode {
   SOME_OTHER_VALUE, MULTI_Spa_IN_GROUP_REJECTED;

    @JsonValue
[...]

PR #2851, opened 2025-12-05
Fix: Keep acronyms uppercase in Java enum constants (#2850)

mohamedelhabib:
Enum constants containing known acronyms (like "SPA") were being incorrectly converted when using --acronym-style=camel, e.g. "MULTI_SPA_IN_GROUP_REJECTED" became "MULTI_Spa_IN_GROUP_REJECTED".

For Java enum constants (UPPER_UNDERSCORE style), acronyms should always remain uppercase regardless of the --acronym-style setting.

- Modified javaNameStyle() to use allUpperWordStyle for acronyms in enum constants
- Added test case for enum with "SPA" acronym


schani:
Thanks for the fix — the one-line change in `javaNameStyle` is correct, and I've kept it as-is. I pushed three commits to this branch to get it ready to merge:

**1. Merged current `master`** (`d2e1c43`) — the branch was a few months behind.

**2. Replaced the test artifacts** (`b1479e5`, `6066b9d`):

- `test/inputs/json/priority/enum-spa.json` was a JSON Schema document added as a JSON *sample*, so it ran through every language fixture with default options and never exercised the `acronym-style` code path — it couldn't catch this bug (and priority samples run in every CI job, which is likely what made CI red here).
- `test/fixtures/java/src/main/java/io/quicktype/MessageCode.java` was a hand-written file that nothing ever compared against generated output.
- More fundamentally, the fixture harness *cannot* catch this regression: the mangled constants are self-consistent identifiers, so the generated Java still compiles, and (de)serialization uses the raw JSON names, so round-tripping succeeds either way.

The replacement is `test/check-java-acronym-names.ts`, following the existing `check-no-node-imports.ts` pattern: it generates Java from a schema whose enum contains the known acronym `SPA`, for **all four** `--acronym-style` settings, and asserts on the emitted enum constant **identifier** (extracted from the `toValue()` switch — just grepping the output for `MULTI_SPA_IN_GROUP_REJECTED` would always pass, because the raw JSON name appears in string literals even when the identifier is mangled). It runs standalone via ts-node and as a pre-fixture check in `test/test.ts`, so it runs in every CI matrix job.

Verified both ways: with your fix temporarily reverted, the check fails with

```
acronym-style=camel: got "MULTI_Spa_IN_GROUP_REJECTED", expected "MULTI_SPA_IN_GROUP_REJECTED"
acronym-style=lowerCase: got "MULTI_spa_IN_GROUP_REJECTED", expected "MULTI_SPA_IN_GROUP_REJECTED"
```

and with the fix present it passes for all four styles.

**3. Reverted the unrelated re-indentation** of the neighboring ternary in `Java/utils.ts` — it failed `biome check`; the diff is now just the semantic change plus a comment.

PR #2875, opened 2026-07-06
Fix crash on non-string const in JSON Schema

### LINKED ISSUE #2333: CLI type generation fails - s.codePointAt is not a function

### LINKED ISSUE #2516: const number type throws error

schani:
## The bug

Any JSON Schema containing a non-string `const` — e.g.

```json
{ "type": "object", "properties": { "version": { "const": 1 } } }
```

crashes quicktype for **every** target language with:

```
Error: s.codePointAt is not a function.
```

Reproduced on current master before this change. Fixes #2333. Fixes #2516.

## Root cause

In `packages/quicktype-core/src/input/JSONSchemaInput.ts` (`addTypesInSchema` → `convertToType`), the string-enum path was taken whenever the schema had a `const`, regardless of the const's type:

```ts
if (needStringEnum || isConst) {
    const cases = isConst ? [schema.const] : ...;
    unionTypes.push(typeBuilder.getStringType(..., StringTypes.fromCases(cases)));
```

The non-string const value then reaches enum-case name generation (`splitIntoWords` → `codePointAt`), which can only handle strings, and throws.

## The fix

Gate the string-enum path on the const actually **being a string** (`typeof schema.const === "string"`):

- a **string** const still becomes a single-case enum — output is byte-identical to master (verified below);
- a **non-string** const now degrades to its plain primitive type (integer / double / bool). The literal value is dropped; generating proper literal types is future work (noted in a FIXME, preserving the intent of the original PR's comment).

### Relationship to #2371

This supersedes #2371 by @DanielBretzigheimer, which had the right approach but rotted against master and left one hole: it gated on the *declared* type, so the degenerate `{"type": "string", "const": 5}` still crashed (the declared `"string"` kept the enum path alive with a non-string case). Gating on `typeof schema.const` instead also fixes that case — it now falls through to a plain `string` type. Daniel is credited as co-author on both commits.

## Tests (written first)

New schema fixture `test/inputs/schema/const-non-string.schema` + samples, exercising:

- a bare integer const (`{"const": 1}` — the exact repro from #2333; becomes `double`),
- an integer const with declared type (`{"type": "integer", "const": 5}` — the degenerate-adjacent case; becomes `integer`),
- a number const (`{"const": 3.5}`),
- a boolean const (`{"const": true}`),
- a string const (`{"const": "widget"}`) to pin the existing single-case-enum behavior, plus a `.fail.enum.json` sample checking that enum-verifying languages still reject other values.

**Tests-first evidence:** on unpatched master this fixture cannot even be generated — quicktype crashes with `s.codePointAt is not a function` for every language — so the fixture itself is the failing test (first commit), fixed by the second commit.

`{"const": null}` is deliberately left out of the fixture: with the fix it no longer hits the internal `TypeError`, but it produces an empty union and quicktype reports `Trying to make an empty union - do you have an impossible type in your schema?`. Proper null-const support is part of the literal-types future work.

## Verification done locally

- **Generation succeeds for all 26 CLI target languages** on the new fixture (was: crash in all of them).
- **Full test harness run and passing** for `schema-typescript`, `schema-javascript`, `schema-golang`, `schema-rust` on the new fixture (round-trip + expected-failure sample).
- **Manual driver round-trips** (generate → compile → parse sample → re-serialize → compare; fail sample rejected where the language checks enums): Python 3.12 (test driver, but no `mypy` locally so the compile lint step didn't run), C++ (g++ 13, `--no-boost` variant since boost headers aren't installed locally; CI runs the boost variant), C# Newtonsoft (`dotnet` 6 via Docker), Java (Maven via Docker, with `-Dmaven.compiler.source=8` override for the modern JDK), Dart (via Docker).
- **Not run locally** (generation-only): swift, kotlin, scala3, elm, ruby, php, haskell, crystal, elixir, objc, flow, pike, cjson and the remaining harness languages — CI covers these; the fixture's shapes (plain primitives + a single-case string enum) are the same shapes existing `enum.schema` and number fixtures already exercise, so no per-language skips were added.
- **String-const output byte-identical to master** across typescript, go, python, csharp, java, rust, c++, swift, kotlin, ruby, dart for a string-const + string-enum schema.
- **Degenerate case** `{"type": "string", "const": 5}`: crashed on master (and would still crash under #2371's gating) → now generates a plain `string` property.
- `npx tsc --noEmit -p packages/quicktype-core` clean; `npx biome check` on the touched file shows no new findings (only pre-existing ones at shifted line numbers).


schani:
CI caught a real gap in the new fixture: `const-non-string.1.fail.enum.json` expects the invalid string-const value to be rejected, but CJSON (the one genuinely failing job — the rest were fail-fast cancellations) doesn't validate enum values, which is exactly why `enum.schema`/`enum-large.schema` are in its `skipSchema`. Pushed a commit adding `const-non-string.schema` to the same eight skip lists that already skip both enum schemas (Crystal, CJSON, Swift, Scala3, Dart, Haskell, TS-Zod, TS-EffectSchema) — the proven-green pattern from #2846.

PR #2886, opened 2026-07-07
Handle Windows absolute paths for schema inputs

### LINKED ISSUE #2869: [BUG]: in Windows it doesn't work with absolute paths and interprets them as a link
## Issue Type

path to file

```powershell
quicktype -s schema C:\Users\Cat\axe\projects\neo-app\data\schemes\bookmark.schema.json -o C:\Users\Cat\axe\projects\neo-app\data\typescript\bookmark.ts --just-types
```
```
Error: Could not fetch schema , referred to from c:\Users\Cat\axe\projects\neo-app\data\schemes\bookmark.schema.json#items.
```

If you run it as a relative path, everything works.

```powershell
quicktype -s schema .\data\schemes\bookmark.schema.json -o .\data\typescript\bookmark.ts --just-types
```

The problem arose when I wrote a PowerShell script that takes all the files in a folder and generates types from them. It had an absolute path, which prevented it from working.

I solved this problem in the script by replacing it with a relative path, but that seems like a hack.

schani:
## Problem

On Windows, giving quicktype a schema input as an absolute path fails:

```
quicktype -s schema C:\Users\Cat\...\bookmark.schema.json -o out.ts --just-types
Error: Could not fetch schema , referred to from c:\Users\Cat\...\bookmark.schema.json#items.
```

Schemas without relative `$ref`s fail too, with `Internal error: Defined value expected`. Relative paths work.

## Root cause

Schema addresses are parsed with urijs, and a Windows absolute path is not a valid URI: urijs parses the drive letter of `C:\Users\me\top.schema.json` as a URI *scheme* (lowercasing it, and keeping the backslashes as an opaque path). The mangled address breaks schema-store lookups, and relative `$ref`s resolved against it via `absoluteTo` yield bogus addresses like `c:///item.schema.json`, which `is-url` in NodeIO then classifies as HTTP URLs and tries to fetch.

## The fix

- `packages/quicktype-core/src/input/JSONSchemaInput.ts`: a new `fixWindowsPath` helper converts Windows absolute paths (drive-letter `C:\...`/`C:/...` and UNC `\\server\share\...` forms) to `file:` URIs, applied at the three places where address strings are parsed into URIs (`normalizeURI`, `Ref.root`, `Ref.parse`). urijs handles `file:` URIs correctly, so normalization, schema-store keying, and relative-`$ref` resolution all work unchanged (drive-letter casing is preserved because the letter is now in the URI path, not the scheme).
- `packages/quicktype-core/src/input/io/NodeIO.ts`: `readableFromFileOrURL` now reads `file:` URIs from disk instead of letting `is-url` classify them as HTTP URLs. Deliberately not `url.fileURLToPath` (its result is platform-dependent): a drive-letter URI maps to `C:/dir/x.json`, which is the correct absolute path on Windows (Node's fs accepts forward slashes there) and a cwd-relative path on POSIX, which is what makes the bug testable on Linux CI.

Plain POSIX paths never hit either code path, and other input kinds (JSON, GraphQL) don't go through the schema-address machinery at all.

## Tests (written first)

`test/check-windows-schema-paths.ts`, run by `test/test.ts` before the fixtures in every CI job (same pattern as the #2850 check). It builds a literal `C:/Users/quicktype/` directory tree in a temp dir, chdirs into it, and runs the quicktype-core pipeline (`JSONSchemaInput` + `FetchingJSONSchemaStore`, exactly as the CLI wires it) with three schema-path spellings, where the top-level schema `$ref`s a sibling file so relative-ref resolution is asserted too:

- Windows absolute path with backslashes
- Windows absolute path with forward slashes
- POSIX absolute path (guards against regressing existing behavior)

The check was added and run **before** the fix, failing with the issue's exact symptom on both Windows cases (POSIX case passed):

```
error: schema inputs given as absolute paths must work (issue #2869):
    Windows absolute path with backslashes: quicktype failed: Error: Internal error: Defined value expected, but got undefined
    Windows absolute path with forward slashes: quicktype failed: Error: Internal error: Defined value expected, but got undefined
```

After the fix it passes, and the issue's original command (simulated on Linux via a literal `C:/Users/Cat/` tree, run as `quicktype -s schema 'C:\Users\Cat\bookmark.schema.json' --lang typescript --just-types`) now produces the same output as the relative-path invocation.

## Verification notes

- `FIXTURE=schema-golang script/test` on all 63 local schema samples: green. (`ref-remote.schema` was excluded — it fails identically on unmodified master in this environment because remote fetching via `global.fetch` doesn't work with the local `getStream`; CI swaps in cross-fetch via `env.sh`. Unrelated to this change.)
- `FIXTURE=golang` on JSON priority samples: green (NodeIO's non-schema path unaffected).
- Not verifiable locally: behavior on an actual Windows machine (no Windows CI job exists). The drive-letter and UNC → `file:` URI mappings follow the standard Windows file-URI convention, and the POSIX-side simulation exercises the full pipeline including `$ref` resolution.

Fixes #2869


packages/quicktype-core/src/input/io/NodeIO.ts
```diff
@@ -39,6 +39,28 @@ function parseHeaders(httpHeaders?: string[]): HttpHeaders {
...
 }
 
+// "file:" URIs come from JSONSchemaInput, which converts Windows absolute
+// schema paths to them because they are not valid URIs (issue #2869). We
+// don't use `url.fileURLToPath` because its result is platform-dependent:
+// drive-letter URIs must also be readable on POSIX (as paths relative to the
+// working directory), which is how the tests exercise this, and Node's fs
+// accepts forward slashes on Windows anyway. The addresses were URI-decoded
+// when they were normalized, so there's no percent-decoding to do here.
+function filePathFromFileURI(fileURI: string): string {
```
schani: Put that in a utility file and add comprehensive unit tests.

packages/quicktype-core/src/input/JSONSchemaInput.ts
```diff
@@ -136,22 +136,41 @@ function checkJSONSchema(x: unknown, refOrLoc: Ref | (() => Ref)): JSONSchema {
 
 const numberRegexp = /^[0-9]+$/;
 
+// Windows absolute paths are not valid URIs: urijs parses the drive letter
+// of e.g. "C:\Users\me\top.schema.json" as a URI scheme (and lowercases it,
+// and treats the backslashes as an opaque path), so relative refs resolve to
+// bogus addresses (issue #2869). Convert drive-letter and UNC paths to
+// "file:" URIs, which NodeIO knows how to read.
+function fixWindowsPath(pathOrURI: string): string {
```
schani: Put that in a shared utility file and add comprehensive unit tests.

schani (commented):
We should probably first convert our more regular tests to vitest, then we can add these unit tests.

packages/quicktype-core/src/input/io/NodeIO.ts
schani: Moved to `support/WindowsPaths.ts` together with `fixWindowsPath` (they're the two halves of the same protocol), with unit tests in `test/unit/windows-paths.test.ts` covering drive-letter/UNC/POSIX conversions, passthrough cases, and round-trips — 2d98579f.

packages/quicktype-core/src/input/JSONSchemaInput.ts
schani: Moved to the shared `support/WindowsPaths.ts` (together with NodeIO's `filePathFromFileURI`) and added comprehensive unit tests in `test/unit/windows-paths.test.ts` — including the lookalikes that must pass through unchanged (`CC:\x`, `1:\x`, `C:x`, single backslash, existing URIs) — 2d98579f. Now that #2892 landed, these are proper Vitest tests, which unblocks the point from the review body.

Issue #2874, opened 2026-07-06
[BUG]: quicktype-core prints \"=== RUNNING IN CI, USE FETCH.CI ===\" on every import (>= v23.3.1)


LangLangBart:
## Issue Type

Runtime

## Context (Environment, Version, Language)

**CLI, npm, or app.quicktype.io:** npm package (`quicktype-core`)
**Version:** >= 23.3.1

## Description

Every import prints `=== RUNNING IN CI, USE FETCH.CI ===` to stdout before any user code runs.

Issue started with `23.3.1`

## Expected Behaviour / Output

No unexpected console output on library import.

## Current Behaviour / Output

```txt
=== RUNNING IN CI, USE FETCH.CI ===
```

printed on every import of the package

## Steps to Reproduce

1. `npm init -y && npm install quicktype-core@latest`
2. `node -e "require('quicktype-core')"`
3. Observe the spurious log line

```bash
podman run --quiet --rm node:alpine sh -c '
cd /tmp && npm init -y > /dev/null
npm install quicktype-core@23.3.0 > /dev/null
node -e "require(\"quicktype-core\")" && echo OK'
# OK

podman run --quiet --rm node:alpine sh -c '
cd /tmp && npm init -y > /dev/null
npm install quicktype-core@23.3.1 > /dev/null
node -e "require(\"quicktype-core\")" && echo OK'
# === RUNNING IN CI, USE FETCH.CI ===
# OK
```

schani:
Confirmed reproducible on current master (v23.3.0, commit `68ecb284`) — this is a triage finding from a systematic sweep of the issue backlog (assisted by Claude); root cause verified locally (local `dist` build is clean since `CI` is unset there — the bug is in what actually gets *published* to npm).

**Root cause:** `packages/quicktype-core/env.sh`, run as part of the build, rewrites `$fetch` imports to `$fetch.ci` whenever `$CI` is set, unless `PUBLISH=true` is also set:
```sh
if [[ $PUBLISH == true ]]; then
  exit 0
fi
if [[ $CI ]]; then
  grep -rl '$fetch' src | xargs sed -i -e 's/$fetch/$fetch.ci/g'
fi
```
[...]

PR #2887, opened 2026-07-07
Fix URL inputs broken by WHATWG fetch streams on Node 18+

### LINKED ISSUE #2613: [BUG]: Cannot fetch external links from $ref due inputStream.once error
When generating an output in Go, Java or Python for a JsonSchema that contains an external link reference using the $ref key, Quicktype should resolve it by fetching that URL to download the schema. Still, it's showing the following error instead:

```bash
Error: Cannot read from file or URL https://raw.githubusercontent.com/cloudevents/spec/v1.0.2/cloudevents/formats/cloudevents.json: TypeError: inputStream.once is not a function
```

**Important note**: this issue is not reproducible in lower versions like `23.0.106`.

## Issue Type

Issue with input parsing of JSON schemas with external link references (using $ref)

Input Format: JSONSchema
Output Language: Go, Java, Python

CLI, npm, or app.quicktype.io: npm package
Version: 23.0.170

## Description

Autogenerate code to use JSON schemas in different languages like Go, Java or Python.
In lower versions of the Quicktype package, like `23.0.106,` this issue does not occur, and everything works as expected, but updating to the latest (`23.0.170`) reproduced the issue.

## Input Data

```json
{
    "description": "test schema",
    "type": "object",
    "allOf": [
      {
        "$ref": "https://raw.githubusercontent.com/cloudevents/spec/v1.0.2/cloudevents/formats/cloudevents.json"
      }
    ]
  }
 
```

As it happens with lower versions, it should generate the code in the expected language correctly.

## Current Behaviour / Output

Outputs an error and doesn't generate the code.

## Steps to Reproduce

1. Using quicktype as npm package dependency version 23.0.170
2. Running an npm script with "quicktype -s schema test_schema.json -o schema.go" where `test_schema.json` has an external link (similar to the input data provided).

### LINKED ISSUE #2678: [BUG]: Syntax error in input JSON flutter: readStream.setEncoding is not a function
```bash
quicktype https://storage.flutter-io.cn/flutter_infra_release/releases/releases_linux.json -o flutter.ts
(node:92203) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
Error: Syntax error in input JSON flutter: readStream.setEncoding is not a function.
```

### LINKED ISSUE #2821: [BUG]: Generating TS types from JSON url doesn't work
## Issue Type

Runtime

Input Format: json
Output Language: ts

CLI, npm, or app.quicktype.io: CLI
Version: 23.2.6

## Description

Using `npx quicktype https://storage.flutter-io.cn/flutter_infra_release/releases/releases_linux.json -o test.ts` should generate TS types but instead it throws `Error: Syntax error in  JSON : .`. I'm assuming the error handling is broken somehow but logging the entire error shows

```
QuickTypeError: Syntax error in  JSON :
    at messageError (node_modules/quicktype-core/dist/Messages.js:97:11)
    at messageParseError (node_modules/quicktype-core/dist/input/Inputs.js:23:40)
    at JSONInput.<anonymous> (node_modules/quicktype-core/dist/input/Inputs.js:68:24)
    at Generator.throw (<anonymous>)
    at rejected (node_modules/quicktype-core/dist/input/Inputs.js:6:65)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
  errorMessage: 'Syntax error in ${description} JSON ${address}: ${message}',
  messageName: 'MiscJSONParseError',
  userMessage: 'Syntax error in  JSON : ',
  properties: {
    description: 'input',
    address: 'releases_linux',
    message: 'readStream.setEncoding is not a function'
  }
}
```

Related #2678

schani:
## Problem

Every URL-based input is broken when quicktype runs on Node >= 18 with the native fetch:

- JSON from a URL fails with `Syntax error in input JSON …: readStream.setEncoding is not a function` (#2678, #2821)
- A schema URL or a remote `$ref` fails with `inputStream.once is not a function`, surfaced as `Could not fetch schema …` because `JSONSchemaStore.get` swallows the underlying error (#2613)

## Root cause

`packages/quicktype-core/src/input/io/$fetch.ts` prefers `global.fetch` when it exists. On Node >= 18 that is the native (undici) fetch, whose `response.body` is a **WHATWG `ReadableStream`**, not a Node `Readable`. `readableFromFileOrURL` in `NodeIO.ts` cast that body straight to a Node `Readable` (`as unknown as Readable`), so every consumer that calls Node stream APIs (`setEncoding`, `once`, `pipe`) blew up.

## The fix

`readableFromResponseBody` feature-detects a web stream (`typeof body.getReader === "function"`) and wraps it in a Node-style `Readable` via an async-generator adapter and `Readable.from` (`readable-stream` implements `from`, but `@types/readable-stream` doesn't declare it, hence a typed cast). The cross-fetch fallback (node-fetch), whose body already is a Node stream, passes through unchanged.

**Also removed the CI cross-fetch build substitution** (`env.sh` + `$fetch.ci.ts`, added in #2744/#2745): it was a workaround for exactly this bug, and it meant CI never tested the fetch path that ships. Worse, the current publish workflow doesn't set `PUBLISH=true`, so the substitution leaked into releases — `quicktype-core` 23.3.17 on npm ships `NodeIO.js` requiring `$fetch.ci`, forcing cross-fetch and printing `=== RUNNING IN CI, USE FETCH.CI ===` on import (see #2874). With the real bug fixed, CI and releases build and test the same code.

## Tests (written first, failed before the fix)

The fixture harness only feeds quicktype local files, so it could not catch this (and `test/inputs/urls/github.json` is not referenced by any harness code). New `test/check-url-input.ts` runs from `test/test.ts` before the fixtures — so it runs in every CI matrix job. It serves a JSON sample plus a JSON Schema with a relative `$ref` from a local HTTP server and runs quicktype against the URLs.

The check was run **before** the fix and failed with the exact reported errors:

```
    JSON from URL: quicktype threw: Error: Syntax error in input JSON TopLevel: readStream.setEncoding is not a function
    schema from URL: quicktype threw: Error: Internal error: Defined value expected, but got undefined
```

After the fix it passes. With env.sh gone, CI builds use `$fetch.ts` (native fetch on Node 22 per `.nvmrc`), so the check exercises the WHATWG path in CI too.

## Verification

- `npm run build` passes (exit 0).
- Regression check passes (native fetch, Node 24 locally).
- CLI-level repros pass: `--lang ts http://…/sample.json`, `-s schema --lang go http://…/main.schema` (remote relative `$ref` resolved), and issue #2678's exact command against `https://raw.githubusercontent.com/…/package.json`.
- `FIXTURE=typescript-zod script/test`: 210/210 pass.
- Not verified locally: other language fixtures (CI matrix covers them); the browser/app.quicktype.io bundle (browsers take the same `getReader` wrapping path, and both the adapter and `readable-stream` are browser-safe, but I could not run that build here); `npm run lint` (no eslint config exists in the repo — lint is broken independently of this change).

Fixes #2613
Fixes #2678
Fixes #2821


schani:
Rebased onto master and fixed the CI failure.

**Why CI failed:** `test/test.ts` is re-executed by every forked cluster worker, so `checkURLInput()` ran in all ~16 workers concurrently. In a Node cluster worker, `server.listen(0)` gives every worker the **same** shared ephemeral port, with incoming connections round-robined between the sibling processes — the two failing worker logs even show the same port. A worker's fetch for the relative `$ref` (`referenced.schema`) could be answered by a sibling's server that was already closing, so the check failed nondeterministically with "Could not fetch schema referenced.schema". The check now runs only in the cluster primary (before any workers are forked), where it has its own private port.

**Rebase notes:** master's #2882 reworked the CI fetch shim's banner; this PR deletes the shim and the `env.sh` substitution entirely, so the deletion was kept, and the now-stale comment in `test/check-clean-import.ts` (which described the substitution) was updated. Both master's `checkCoreImportKeepsStdoutClean` and this PR's `checkURLInput` are kept in `test/test.ts`.

Verified locally: `CI=true QUICKTEST=true FIXTURE=javascript npm test` passes (previously reproduced the exact CI failure); the standalone check passes, still fails against master's unfixed `NodeIO.ts`, and `quicktype --src http://…/smoke.json` works end-to-end with the built CLI on native fetch.

PR #2888, opened 2026-07-07
typescript-zod: emit z.lazy() for recursive schemas instead of silently dropping them

### LINKED ISSUE #2508: converting a recursive json schema into a zod type breaks
example self referential schema:
```
{
  "id": "http://json-schema.org/geo",
  "description": "A geographical coordinate",
  "type": "object",
  "definitions":{
    "Mydef1":{
        "type":"object",
        "properties":{
            "test":{
                "$ref":"#/definitions/Mydef1"
            }
            
        }
    }
  },
  "properties": {
    "latitude": {
       "$ref":"#/definitions/Mydef1"
    }
  }
}
```

produces:

```
export const Mydef1Schema = z.object({
    "test": z.union([Mydef1Schema, z.null()]).optional(),
});
export type Mydef1 = z.infer<typeof Mydef1Schema>;

export const CoordinateSchema = z.object({
    "latitude": z.union([Mydef1Schema, z.null()]).optional(),
});
export type Coordinate = z.infer<typeof CoordinateSchema>;
```
but it should be 
https://github.com/colinhacks/zod#recursive-types
```
export const Mydef1Schema = z.object({
    "test": z.union([z.lazy(() =>Mydef1Schema), z.null()]).optional(),
});
export type Mydef1 = z.infer<typeof Mydef1Schema>;

export const CoordinateSchema = z.object({
    "latitude": z.union([Mydef1Schema, z.null()]).optional(),
});
export type Coordinate = z.infer<typeof CoordinateSchema>;
```

schani:
## Problem

Converting a recursive JSON Schema (or recursive JSON input) with `--lang typescript-zod` produced broken output. Originally (#2508) the recursive reference was emitted without `z.lazy()`, which throws at runtime; on current master it is even worse — the output degenerates to just `import * as z from "zod";` plus an "Exceeded maximum number of passes" warning, with **all** schemas silently dropped.

## Root cause

`emitSchemas()` orders schemas topologically so every schema is defined before it is referenced. Types that participate in a reference cycle can never be placed by that loop. After 999 passes the bailout pushed `deferredIndices` — which had just been reset to `[]` — instead of the stuck `indices`, so recursive types *and everything depending on them* vanished from the output. And even with the bailout fixed, the renderer had no `z.lazy()` support, so recursive schemas referenced consts before their definition.

## The fix

- Detect object types that participate in a reference cycle by walking child type refs (`packages/quicktype-core/src/language/TypeScriptZod/TypeScriptZodRenderer.ts`).
- Emit those types with an explicitly declared TS type and a `z.lazy()`-wrapped schema, as zod cannot infer recursive types ([zod docs](https://github.com/colinhacks/zod#recursive-types)):

  ```ts
  export type Mydef1 = {
      "test"?: Mydef1;
  };
  export const Mydef1Schema: z.ZodType<Mydef1> = z.lazy(() =>
      z.object({
          "test": Mydef1Schema.optional(),
      })
  );
  ```

  Recursive types are emitted first: inside `z.lazy()` their references are deferred to parse time, so they don't constrain the topological order, while other schemas can reference them directly. With `--just-schema` the annotation falls back to `z.ZodType<any>`.
- Fix the bailout to push the stuck `indices` so schemas are never silently dropped even if ordering ever fails.

## Tests (written first)

**The regression test was added and shown to fail before the fix:**

- Wired up a `schema-typescript-zod` `JSONSchemaFixture` (`test/fixtures.ts`) and added it to the CI matrix (`.github/workflows/test-pr.yaml`). Running `FIXTURE=schema-typescript-zod script/test test/inputs/schema/mutually-recursive.schema` against the unfixed renderer failed: the generated `TopLevel.ts` contained only the zod import, so the test driver errored on both samples ("Could not run command for JSON output", exit 1). The issue's own schema likewise produced only `import * as z from "zod";`. After the fix both pass.
- Un-skipped the typescript-zod JSON fixture inputs that were skipped as "Does not handle recursive" (`direct-recursive.json`, `list.json`, `bug790.json`, `recursive.json`) and the "Uses generated schema before it's defined" group (`be234`, `76ae1`, `6de06`, `29f47`, `spotify-album`, `reddit`) — all pass now and guard against regressions in the existing CI fixture.
- New skips, each verified to be unrelated to this fix: `constructor.schema` (zod v3 validates the inherited `Object.prototype.constructor` when an optional `"constructor"` property is absent from the parsed object), `date-time.schema` / `github-events.json` (`z.coerce.date()` round-trips as ISO UTC with milliseconds, not the input string), `2df80.json` (the test driver can't find the top-level schema among the disambiguated `Fluffy`/`PurpleTopLevelSchema` names).

## Verification

- `FIXTURE=typescript-zod script/test`: all 210 tests pass locally (node + tsc).
- `FIXTURE=schema-typescript-zod script/test`: all schema inputs pass locally **except** `ref-remote.schema`, which could not be verified locally — its remote fetch from raw.githubusercontent.com fails in my environment for *all* languages (including master's plain typescript). A local simulation (same `$ref` target `list.schema`, via a file ref) generates correct `z.lazy()` output, and `list.schema` itself passes the fixture, so CI is expected to pass it.
- `vega-lite.schema` (large, heavily recursive) now generates, compiles, and validates.

Fixes #2508

### DIFF
--- a/packages/quicktype-core/src/language/TypeScriptZod/TypeScriptZodRenderer.ts
+++ b/packages/quicktype-core/src/language/TypeScriptZod/TypeScriptZodRenderer.ts
@@ -33,6 +33,11 @@ import { legalizeName } from "../JavaScript/utils";
 import type { typeScriptZodOptions } from "./language";
 
 export class TypeScriptZodRenderer extends ConvenienceRenderer {
+    /** TypeRefs of object types that participate in a reference cycle.
+     * These must be emitted as z.lazy() schemas with an explicit type
+     * annotation, since zod cannot infer recursive types. */
+    private _recursiveTypeRefs: Set<number> = new Set();
+
     public constructor(
         targetLanguage: TargetLanguage,
         renderContext: RenderContext,
@@ -140,7 +145,59 @@ export class TypeScriptZodRenderer extends ConvenienceRenderer {
         return match;
     }
 
+    /** TypeScript type for `t`, used to annotate recursive schemas for
+     * which zod cannot infer the type. */
+    protected underlyingTypeFor(t: Type): Sourcelike {
+        if (["class", "object", "enum"].includes(t.kind)) {
+            return this.nameForNamedType(t);
+        }
+
+        return matchType<Sourcelike>(
+            t,
+            (_anyType) => "any",
+            (_nullType) => "null",
+            (_boolType) => "boolean",
+            (_integerType) => "number",
+            (_doubleType) => "number",
+            (_stringType) => "string",
+            (arrayType) => [
+                "Array<",
+                this.underlyingTypeFor(arrayType.items),
+                ">",
+            ],
+            (_classType) => panic("Should already be handled."),
+            (mapType) => [
+                "Record<string, ",
+                this.underlyingTypeFor(mapType.values),
+                ">",
+            ],
+            (_enumType) => panic("Should already be handled."),
+            (unionType) => {
+                const children = Array.from(unionType.getChildren()).map(
+                    (type: Type) => this.underlyingTypeFor(type),
+                );
+                return arrayIntercalate(" | ", children);
+            },
+            (_transformedStringType) => {
+                if (_transformedStringType.kind === "date-time") {
+                    return "Date";
+                }
+
+                return "string";
+            },
+        );
+    }
+
+    protected isRecursive(t: ObjectType): boolean {
+        return this._recursiveTypeRefs.has(t.typeRef);
+    }
+
     protected emitObject(name: Name, t: ObjectType): void {
+        if (this.isRecursive(t)) {
+            this.emitLazyObject(name, t);
+            return;
+        }
+
         this.ensureBlankLine();
         this.emitLine("\nexport const ", name, "Schema = ", "z.object({");
         this.indent(() => {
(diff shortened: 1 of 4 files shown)


packages/quicktype-core/src/language/TypeScriptZod/TypeScriptZodRenderer.ts
```diff
@@ -140,7 +145,59 @@ export class TypeScriptZodRenderer extends ConvenienceRenderer {
         return match;
     }
 
+    /** TypeScript type for `t`, used to annotate recursive schemas for
+     * which zod cannot infer the type. */
+    protected underlyingTypeFor(t: Type): Sourcelike {
```
schani, line 158: Is this shared with the regular Typescript renderer? Would it help to unify, or would that be a distraction?

schani: Not shared, and I looked at unifying — I think it would be a distraction. The regular TypeScript renderer's counterpart is `sourceFor` in `TypeScriptFlowBaseRenderer`. It differs on essentially every axis:

- It returns `MultiWord` and lives in a different hierarchy (`JavaScriptRenderer` subclass; the zod renderer extends `ConvenienceRenderer` directly).
- Its output is steered by ts/flow-only options that don't exist here: `preferConstValues` (single-case enums become string literals), `declareUnions` (unions may be named types), `T[]`/`parenIfNeeded` array style, `{ [key: string]: T }` maps.
- `underlyingTypeFor` has a stricter contract: it must produce exactly the type `z.infer` would derive from the schema `typeMapTypeFor` emits — e.g. `z.coerce.date()` ⇒ `Date`, `z.record(z.string(), T)` ⇒ `Record<string, T>` — otherwise the `z.ZodType<T>` annotation on the lazy schema stops typechecking. So its real coupling is to `typeMapTypeFor` right above it, not to the TS renderer.

A shared helper would need parameters for all of the above, coupling the two language families for ~30 lines of `matchType` dispatch. I've added a comment on `underlyingTypeFor` recording this (2e0f6dbf) so the next reader doesn't have to re-derive it. Happy to unify anyway if you feel strongly.

PR #2892, opened 2026-07-09
Modernize Node support and migrate unit tests to Vitest

schani:
## Summary

- require Node 20 or newer for published npm packages while using Node 24 for development, full CI fixtures, and publishing
- build and run unit tests on both Node 20 and Node 24 in CI
- remove direct fetch polyfills and cover native fetch with local HTTP tests for URL input, remote schema references, and GraphQL introspection
- migrate standalone regression checks to Vitest while retaining the custom cross-language fixture harness
- update the JavaScript PropTypes fixture to use the maintained prop-types API without the legacy fetch dependency chain
- bump the release line to 24.0.0 and add migration documentation
- retain TypeScript 5.8 for builds and TypeScript 4.9 for the TypeScript-to-schema integration

## Verification

- clean install with npm ci
- full build on Node 20 and Node 24
- 4 Vitest files / 9 tests on Node 20 and Node 24
- focused JavaScript, TypeScript, and JavaScript PropTypes fixtures
- test TypeScript typecheck and Biome formatting checks
- VS Code extension bundle build
- npm package dry runs for the CLI and all three library packages

The full cross-language matrix remains in GitHub Actions because it requires the external language toolchains described in AGENTS.md.


MIGRATING.md
```diff
@@ -0,0 +1,17 @@
+# Migrating to quicktype 24
```
schani, line 1: This should be referred-to from the README.md

schani: Added a direct link from the Node 20 requirement in README.md to MIGRATING.md in 4188aae8.

PR #2926, opened 2026-07-13
Adopt Biome lint rules

schani:
# Adopt Biome lint rules

This PR turns Biome's linter on for real, on top of the formatter-only setup from #2923, and adopts rules deliberately:

1. **First commit** enables the linter with the `recommended` preset. Of the 218 recommended rules, 30 reported diagnostics on the codebase; those were temporarily set to `"off"` so the commit lands with lint green and zero code changes.
2. **One commit per rule** then re-enables each of those 30 and fixes all of its violations — except three that misfit the codebase (see the rejected table).
3. The **non-recommended rules** of Biome 2.5 were then evaluated group by group (correctness, complexity, performance, security, style, suspicious): 29 more rules were adopted with fixes, and 27 already-clean guard rules were enabled in per-group batches. The `nursery` group was skipped as unstable, and `a11y` plus the framework-specific rules (React/Vue/Qwik/Solid/Next/DOM/CSS/GraphQL) don't apply to this library/CLI.

Every commit keeps `npm run lint`, `npm run build`, and the Vitest unit suite green. Lint fixes are behavior-preserving; where a fix would have been risky or wrong, a targeted `biome-ignore` suppression with a real explanation (or a scoped config override) is used instead. Notably, `noUndeclaredDependencies` caught a genuine issue: the CLI imported `wordwrap` without declaring it, relying on hoisting from quicktype-core — it is now a declared dependency.

**Result: 270 rules active** (215 recommended + 55 opted-in), 3 recommended rules off, 3 scoped overrides.

### Recommended preset (re-enabled one by one)

| Rule | Violations fixed | Notes |
|---|---|---|
| correctness/noUnusedImports | 1 | unused lodash import |
| correctness/noUnusedVariables | 5 | unused catch bindings → optional catch |
| correctness/useParseIntRadix | 3 | single-char parses, radix 10 |
| correctness/noSwitchDeclarations | 20 | case blocks scoped |
| complexity/noBannedTypes | 13 | `{}` → `Record<string, never>` |
| complexity/noExtraBooleanCast | 2 | redundant `!!` |
| complexity/noUselessConstructor | 7 | super-forwarding constructors removed |
| complexity/noUselessContinue | 1 | |
| complexity/noUselessSwitchCase | 1 | case adjacent to identical default |
| complexity/noUselessTernary | 1 | `x ? false : true` → `!x` |
| complexity/noUselessUndefinedInitialization | 24 | `= undefined` initializers |
| complexity/useDateNow | 2 | `+new Date()` → `Date.now()` |
| complexity/useOptionalChain | 1 | |
| style/noNonNullAssertion | 2 | replaced `!` with narrowing via destructured consts |
| style/useConst | 1 | |
| style/useExponentiationOperator | 6 | `Math.pow` → `**` |
| style/useNodejsImportProtocol | 6 | `node:` prefix; quicktype-core exempted via override (its `browser` field only remaps bare specifiers — enforced by `core-package.test.ts`) |
| style/useTemplate | 40 | concatenation → template literals, escapes hand-verified |
| suspicious/noExplicitAny | 25 | 6 became `unknown`; deliberate anys (heterogeneous maps, vendored get-stream, untyped tree-sitter) suppressed with reasons |
| suspicious/noGlobalIsNan | 6 | all call sites already number-typed |
| suspicious/noImplicitAnyLet | 3 | explicit annotations |
| suspicious/noPrototypeBuiltins | 2 | already-safe `.call` sites suppressed; `Object.hasOwn` unavailable under core's es6 lib |
| suspicious/noShadowRestrictedNames | 2 | `toString` param renamed; collection-utils `hasOwnProperty` import suppressed |
| suspicious/noTsIgnore | 1 | → `@ts-expect-error` with explanation |
| suspicious/noUselessEscapeInString | 7 | `\"` in template literals |
| suspicious/useIsArray | 1 | `instanceof Array` → `Array.isArray` |
| suspicious/useIterableCallbackReturn | 19 | forEach expression arrows → statement bodies |

### Non-recommended rules (opted in, violations fixed)

| Rule | Violations fixed | Notes |
|---|---|---|
| correctness/noUndeclaredDependencies | 2 | **real catch**: `wordwrap` now declared for the CLI; `vscode` host module suppressed |
| correctness/noGlobalDirnameFilename | 1 | protects the ESM dual build; CJS test harness suppressed |
| complexity/noUselessReturn | 24 | trailing `return;` |
| style/useCollapsedElseIf | 16 | one site hand-collapsed to keep a comment |
| style/useCollapsedIf | 5 | one site hand-collapsed to keep a comment |
| style/useThrowNewError | 15 | `throw Error(...)` → `throw new Error(...)` |
| style/useConsistentBuiltinInstantiation | 0 | overlapped entirely with useThrowNewError |
| style/useObjectSpread | 17 | `Object.assign({}, ...)` → spread; 3 sites needed type annotations |
| style/noYodaExpression | 7 | |
| style/noSubstr | 11 | `substring` → `slice`; one site got an explicit guard to preserve edge-case behavior |
| style/useExplicitLengthCheck | 11 | one numeric-fallback site suppressed (autofix would have produced a boolean) |
| style/noMultiAssign | 9 | chains split |
| style/useConsistentObjectDefinitions | 6 | shorthand |
| style/useForOf | 2 | replaced eslint-disabled index loops |
| style/noUnusedTemplateLiteral | 2 | |
| style/useShorthandAssign | 1 | |
| style/noInferrableTypes | 1 | |
| style/useDefaultParameterLast | 2 | both sites are exported API (`DartRenderer`'s protected dynamic-expression methods), so per review the defaults are kept with targeted suppressions — the rule now only guards new code |
| style/useConsistentMethodSignatures | 2 | property-style interface signatures |
| style/useUnifiedTypeSignatures | 1 | compose() overloads unified |
| style/noNamespace | 1 | obsolete `declare namespace Math` (fround is in ES2015 lib) |
| style/useNodeAssertStrict | 1 | same strict assert object |
| style/useReadonlyClassProperties | 13 | type-level only |
| suspicious/noEqualsToNull | 1 | value is `Type \| null`, never undefined |
| suspicious/noUnusedExpressions | 2 | deliberate `satisfies` checks suppressed |
| suspicious/useArraySortCompare | 3 | string sorts where UTF-16 order is intended, suppressed |
| suspicious/noMisplacedAssertion | 10 | script-style `test/release-version.ts` exempted via override; one test helper suppressed |
| suspicious/noShadow | 29 | one real shadow renamed; CJSONRenderer.ts exempted (28 same-named nested emit-lambda params; bulk rename too risky) |

### Already-clean guard rules enabled (no violations)

correctness: `noUnusedInstantiation`, `useSingleJsDocAsterisk` · complexity: `noDivRegex`, `noUselessCatchBinding`, `noUselessStringConcat`, `useArrayFind`, `useWhile`, `noRedundantDefaultExport`, `noExcessiveNestedTestSuites` · performance: `noDelete` · style: `useThrowOnlyError`, `useTrimStartEnd`, `useSymbolDescription`, `useSpreadOverApply`, `useNumberNamespace`, `useGlobalThis`, `useErrorCause` · suspicious: `noConstantBinaryExpressions`, `noVar`, `noFocusedTests`, `noSkippedTests`, `noReturnAssign`, `noDuplicateTestHooks`, `noDeprecatedImports`, `noDuplicateDependencies`, `useErrorMessage`, `noEmptySource`

### Recommended rules kept off

| Rule | Violations | Rationale |
|---|---|---|
| suspicious/noTemplateCurlyInString | 48 | quicktype's error messages use `${placeholder}` templates in plain strings by design (`Messages.ts`, substituted at runtime by `messageError`), and the Kotlin renderers emit Kotlin's own `${}` interpolation syntax. The rule flags the codebase's intended core patterns wholesale. |
| correctness/noVoidTypeReturn | 13 | rejected in review: the owner prefers `return panic(...)` in void functions — the `return` makes the control-flow termination explicit at the call site. Initially adopted, then reverted. |
| correctness/noConstructorReturn | 6 | rejected in review, same idiom objection as noVoidTypeReturn: all six violations were `return panic(...)` in a constructor. Reverting restored the `@ts-expect-error` on `queryType` that the fix had made obsolete. |

### Non-recommended rules not adopted (off by default; rationale)

| Rule | Violations | Rationale |
|---|---|---|
| style/noUselessElse | 20 | rejected in review: the owner prefers explicit else blocks after returning branches over flattened straight-line code. Initially adopted, then reverted; kept explicitly `"off"`. |
| suspicious/noEmptyBlockStatements | 25 | empty method bodies are deliberate no-op default hooks throughout the renderer class hierarchy |
| suspicious/useAwait | 7 | async methods implement async-by-contract interfaces (`Input.addSource` etc.) with sync bodies |
| suspicious/noParametersOnlyUsedInRecursion | 5 | the fix removes parameters from protected renderer API signatures |
| suspicious/noUnnecessaryConditions | 97 | flags defensive checks; high false-positive rate |
| suspicious/noConsole | 71 | the CLI writes to the console by design |
| suspicious/noBitwiseOperators | 56 | hashing/Markov/Mersenne-Twister code uses bitwise math deliberately |
| suspicious/noUndeclaredEnvVars | — | CLI/test harness read arbitrary env vars by design |
| complexity/noVoid | 5 | `void promise` deliberately marks fire-and-forget calls |
| complexity/noImplicitCoercions | 11 | `+str` / `!!x` are deliberate idioms |
| complexity/noForEach | 41 | forEach is used pervasively; useIterableCallbackReturn guards actual misuse |
| complexity/noExcessiveCognitiveComplexity, noExcessiveLinesPerFunction, useMaxParams | 105/273/70 | arbitrary size thresholds; language renderers are legitimately large |
| complexity/useSimplifiedLogicExpression | 20 | its De Morgan rewrites reduce readability |
| correctness/noUndeclaredVariables, noUnresolvedImports | 0/12 | redundant with the TypeScript compiler |
| correctness/useImportExtensions | 1189 | extension policy is owned by TS NodeNext resolution |
| correctness/noNodejsModules | 38 | it's a Node CLI; core's browser-compat is guarded by its own test |
| performance/noAwaitInLoops | 14 | sequential awaits are intentional (ordering, IO) |
| performance/noNamespaceImport, noBarrelFile, noReExportAll | 39/31/35 | `import *` and barrel index files are the package structure |
| performance/useTopLevelRegex | 20 | hoisting regexes risks shared `lastIndex` state and hurts locality |
| security/noSecrets | 121 | entropy heuristic false-positives on code-generation strings and fixtures |
| security/noDangerouslySetInnerHtml* | — | DOM/JSX only, N/A |
| style/noMagicNumbers | 1328 | absurd churn for a code generator full of structural constants |
| style/useConsistentMemberAccessibility | 949 | repo convention is explicit `public`; the rule's default demands the opposite |
| style/noTernary, useBlockStatements, noContinue, noNestedTernary, noEnum, noIncrementDecrement, noNegationElse | 484/437/46/18/18/97/76 | ban or churn idioms the codebase uses deliberately |
| style/useNamingConvention | 424 | underscore-prefixed privates and renderer naming clash with the rule |
| style/useNumericSeparators | 195 | churn incl. vendored PRNG constants |
| style/noParameterProperties | 154 | constructor parameter properties are a core idiom here |
| style/useDestructuring, useConsistentArrowReturn | 99/41 | mechanical churn, marginal benefit |
| style/noParameterAssign | 75 | parameter normalization (opts defaulting) is deliberate |
| style/useExportsLast, useFilenamingConvention, noExcessiveLinesPerFile, noExcessiveClassesPerFile | 61/47/64/26 | structural churn / arbitrary thresholds |
| style/noProcessEnv, correctness/noProcessGlobal | 16/43 | CLI and test harness use process/env by design |
| style/useConsistentArrayType, useConsistentTypeDefinitions | 33/10 | both forms used deliberately; pure churn |
| style/useAtIndex | 3 | `.at()` is not in quicktype-core's es6 lib |
| style/noCommonJs | 5 | build/test scripts legitimately use require |
| style/noDefaultExport | 3 | existing default exports are module API |
| style/noUselessUndefined | 109 | explicit `undefined` returns/arguments are used deliberately with strict optional typing |
| nursery group | — | skipped as unstable |
| a11y group + React/Vue/Qwik/Solid/Next/GraphQL/CSS rules | — | no DOM/JSX/framework code in this repo |

## Verification

- `npm run lint` exits 0 with zero diagnostics
- `npm run build` (all workspaces + CLI) succeeds
- `npm run test:unit` — 70/70 Vitest tests pass
- after every one of the 68 commits, all three of the above were green

### DIFF
--- a/packages/quicktype-core/src/attributes/TypeAttributes.ts
+++ b/packages/quicktype-core/src/attributes/TypeAttributes.ts
@@ -158,9 +158,8 @@ export function combineTypeAttributes(
         if (attrs.length === 1) return attrs[0];
         if (union) {
             return kind.combine(attrs);
-        } else {
-            return kind.intersect(attrs);
         }
+        return kind.intersect(attrs);
     }
 
     return mapFilterMap(attributesByKind, combine);
--- a/packages/quicktype-core/src/attributes/TypeAttributes.ts
+++ b/packages/quicktype-core/src/attributes/TypeAttributes.ts
@@ -36,9 +36,7 @@ export class TypeAttributeKind<T> {
         _schema: { [name: string]: unknown },
         _t: Type,
         _attrs: T,
-    ): void {
-        return;
-    }
+    ): void {}
 
     public children(_: T): ReadonlySet<Type> {
         return new Set();
--- a/packages/quicktype-core/src/attributes/TypeAttributes.ts
+++ b/packages/quicktype-core/src/attributes/TypeAttributes.ts
@@ -158,8 +158,9 @@ export function combineTypeAttributes(
         if (attrs.length === 1) return attrs[0];
         if (union) {
             return kind.combine(attrs);
+        } else {
+            return kind.intersect(attrs);
         }
-        return kind.intersect(attrs);
     }
 
     return mapFilterMap(attributesByKind, combine);
--- a/packages/quicktype-core/src/attributes/TypeAttributes.ts
+++ b/packages/quicktype-core/src/attributes/TypeAttributes.ts
@@ -118,7 +118,7 @@ export class TypeAttributeKind<T> {
 }
 
 // FIXME: strongly type this
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
+// biome-ignore lint/suspicious/noExplicitAny: heterogeneous by design; attribute values are typed by their kind
 export type TypeAttributes = ReadonlyMap<TypeAttributeKind<any>, any>;
 
 export const emptyTypeAttributes: TypeAttributes = new Map();
@@ -154,7 +154,7 @@ export function combineTypeAttributes(
     const attributesByKind = mapTranspose(attributeArray);
 
     // FIXME: strongly type this
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    // biome-ignore lint/suspicious/noExplicitAny: heterogeneous by design; attribute values are typed by their kind
     function combine(attrs: any[], kind: TypeAttributeKind<any>): any {
         assert(attrs.length > 0, "Cannot combine zero type attributes");
         if (attrs.length === 1) return attrs[0];
--- a/packages/quicktype-core/src/input/CompressedJSON.ts
+++ b/packages/quicktype-core/src/input/CompressedJSON.ts
@@ -56,15 +56,15 @@ export abstract class CompressedJSON<T> {
 
     private _ctx: Context | undefined;
 
-    private _contextStack: Context[] = [];
+    private readonly _contextStack: Context[] = [];
 
-    private _strings: string[] = [];
+    private readonly _strings: string[] = [];
 
-    private _stringIndexes: { [str: string]: number } = {};
+    private readonly _stringIndexes: { [str: string]: number } = {};
 
-    private _objects: Value[][] = [];
+    private readonly _objects: Value[][] = [];
 
-    private _arrays: Value[][] = [];
+    private readonly _arrays: Value[][] = [];
 
     public constructor(
         public readonly dateTimeRecognizer: DateTimeRecognizer,
--- a/packages/quicktype-core/src/input/CompressedJSON.ts
+++ b/packages/quicktype-core/src/input/CompressedJSON.ts
@@ -182,7 +182,7 @@ export abstract class CompressedJSON<T> {
     }
 
     protected commitString(s: string): void {
-        let value: Value | undefined = undefined;
+        let value: Value | undefined;
         if (this.handleRefs && this.isExpectingRef) {
             value = this.makeString(s);
         } else {
--- a/packages/quicktype-core/src/input/CompressedJSON.ts
+++ b/packages/quicktype-core/src/input/CompressedJSON.ts
@@ -105,6 +105,7 @@ export abstract class CompressedJSON<T> {
     }
 
     protected internString(s: string): number {
+        // biome-ignore lint/suspicious/noPrototypeBuiltins: Object.hasOwn is not in quicktype-core's es6 lib
         if (Object.prototype.hasOwnProperty.call(this._stringIndexes, s)) {
             return this._stringIndexes[s];
         }
--- a/packages/quicktype-core/src/input/CompressedJSON.ts
+++ b/packages/quicktype-core/src/input/CompressedJSON.ts
@@ -154,7 +154,9 @@ export abstract class CompressedJSON<T> {
             this._rootValue = value;
         } else if (this._ctx.currentObject !== undefined) {
             if (this._ctx.currentKey === undefined) {
-                panic("Must have key and can't have string when committing");
+                return panic(
+                    "Must have key and can't have string when committing",
+                );
             }
 
             this._ctx.currentObject.push(
@@ -165,7 +167,7 @@ export abstract class CompressedJSON<T> {
         } else if (this._ctx.currentArray !== undefined) {
             this._ctx.currentArray.push(value);
         } else {
-            panic("Committing value but nowhere to commit to");
+            return panic("Committing value but nowhere to commit to");
         }
     }
 
@@ -256,7 +258,7 @@ export abstract class CompressedJSON<T> {
     protected finishObject(): void {
         const obj = this.context.currentObject;
         if (obj === undefined) {
-            panic("Object ended but not started");
+            return panic("Object ended but not started");
         }
 
         this.popContext();
@@ -271,7 +273,7 @@ export abstract class CompressedJSON<T> {
     protected finishArray(): void {
         const arr = this.context.currentArray;
         if (arr === undefined) {
-            panic("Array ended but not started");
+            return panic("Array ended but not started");
         }
 
         this.popContext();
@@ -359,7 +361,7 @@ export class CompressedJSONFromString extends CompressedJSON<string> {
 
             this.finishObject();
         } else {
-            panic("Invalid JSON object");
+            return panic("Invalid JSON object");
         }
     }
 }
--- a/packages/quicktype-core/src/input/CompressedJSON.ts
+++ b/packages/quicktype-core/src/input/CompressedJSON.ts
@@ -153,9 +153,7 @@ export abstract class CompressedJSON<T> {
             this._rootValue = value;
         } else if (this._ctx.currentObject !== undefined) {
             if (this._ctx.currentKey === undefined) {
-                return panic(
-                    "Must have key and can't have string when committing",
-                );
+                panic("Must have key and can't have string when committing");
             }
 
             this._ctx.currentObject.push(
@@ -166,7 +164,7 @@ export abstract class CompressedJSON<T> {
         } else if (this._ctx.currentArray !== undefined) {
             this._ctx.currentArray.push(value);
         } else {
-            return panic("Committing value but nowhere to commit to");
+            panic("Committing value but nowhere to commit to");
         }
     }
 
@@ -257,7 +255,7 @@ export abstract class CompressedJSON<T> {
     protected finishObject(): void {
         const obj = this.context.currentObject;
         if (obj === undefined) {
-            return panic("Object ended but not started");
+            panic("Object ended but not started");
         }
 
         this.popContext();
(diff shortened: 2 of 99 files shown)


packages/quicktype-core/src/attributes/TypeAttributes.ts
```diff
@@ -154,15 +152,14 @@ export function combineTypeAttributes(
...
 
     // FIXME: strongly type this
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    // biome-ignore lint/suspicious/noExplicitAny: heterogeneous by design; attribute values are typed by their kind
     function combine(attrs: any[], kind: TypeAttributeKind<any>): any {
         assert(attrs.length > 0, "Cannot combine zero type attributes");
         if (attrs.length === 1) return attrs[0];
         if (union) {
             return kind.combine(attrs);
-        } else {
```
schani, line 163: I don't like this rule that forbids us from using `else` in this way. Disable that rule and undo any changes we had to make for it.

packages/quicktype-core/src/input/CompressedJSON.ts
```diff
@@ -153,9 +154,7 @@ export abstract class CompressedJSON<T> {
             this._rootValue = value;
         } else if (this._ctx.currentObject !== undefined) {
             if (this._ctx.currentKey === undefined) {
-                return panic(
-                    "Must have key and can't have string when committing",
-                );
+                panic("Must have key and can't have string when committing");
```
schani: I don't like this. Whatever the rule is that forces this change, disable it and undo the changes.

packages/quicktype-core/src/attributes/TypeAttributes.ts
schani: This was `style/noUselessElse`. Disabled (explicitly `"off"` in biome.json to document the rejection) and all 20 of its changes reverted in d7918833. PR body tables updated.

packages/quicktype-core/src/input/CompressedJSON.ts
schani: This was `correctness/noVoidTypeReturn` (a recommended rule). Disabled and all 13 of its changes reverted in 831388f0. Note: `correctness/noConstructorReturn` (still enabled) made 6 similar `return panic(...)` → `panic(...)` changes in the GQLSchemaFromJSON constructor — there the change let TypeScript's definite-assignment analysis work and removed a `@ts-expect-error`, so it's left as is; say the word if you want that reverted too.

packages/quicktype-core/src/language/Dart/DartRenderer.ts
```diff
@@ -419,9 +419,8 @@ export class DartRenderer extends ConvenienceRenderer {
     // If the first time is the unionType type, after nullableFromUnion conversion,
     // the isNullable property will become false, which is obviously wrong,
     // so add isNullable property
-    // eslint-disable-next-line @typescript-eslint/default-param-last
     protected fromDynamicExpression(
-        isNullable = false,
+        isNullable: boolean,
```
schani: `DartRenderer` is exported by `quicktype-core`, so this protected method is available to downstream subclasses. Removing `= false` makes existing calls that explicitly pass `undefined` fail type-checking. The same regression exists in `toDynamicExpression` below. Since this PR is intended to keep lint fixes behavior- and API-preserving, please retain both defaults and add targeted `biome-ignore lint/style/useDefaultParameterLast` suppressions instead.

schani (commented):
One API-compatibility issue found.

packages/quicktype-core/src/input/CompressedJSON.ts
schani: Done — `correctness/noConstructorReturn` is now also disabled (explicit `"off"`) and all 6 of its changes reverted in 889d3198, including restoring the `@ts-expect-error` on `queryType`.

packages/quicktype-core/src/language/Dart/DartRenderer.ts
schani: Fixed in 3738f48d: both `isNullable = false` defaults restored on `fromDynamicExpression` and `toDynamicExpression`, each with a targeted `biome-ignore lint/style/useDefaultParameterLast` suppression on the parameter. I audited the rest of that rule's commit: these two were its only violations, so no other exported/protected API was affected — the rule stays enabled and now only guards new code.

PR #2929, opened 2026-07-13
Optimize Markov chain representation

schani:
## Summary

- replace the gzip/Base64 JSON trie with a Base91 + rANS payload
- quantize transition probabilities into a dense five-level table for allocation-free lookup
- transpose the table and cluster 4,225 columns into 16 shared entropy models
- calibrate the compact scores to preserve every original map-inference gate in the fixture corpus
- remove the direct `pako` and `js-base64` dependencies
- add reproducible size/speed and decision-fidelity benchmarks

## Why

The previous runtime inflated and parsed a roughly 1.9 MB JSON trie into nested JavaScript objects and 128-slot arrays. The first iteration replaced it with exact typed-array count tables, but its encoded string was still 110 KB. The new representation spends the available error budget on five calibrated probability levels and entropy-codes the dense table with contextual rANS models.

## Benchmark

Measured on Node 24.6.0, darwin/arm64:

| Metric | Original | First iteration | Current |
| --- | ---: | ---: | ---: |
| Encoded string | 197,592 chars | 110,385 chars | **47,047 chars** |
| Browser bundle, including decoder | 271,049 bytes | 112,365 bytes | **49,451 bytes** |
| Retained memory | 3,449,830 bytes | ~293,904 bytes | **~275,489 bytes** |
| Decode median | 20.5 ms | 4.24 ms | **6.03 ms** |
| Inference throughput | 3.46M/s | 13.85M/s | **15.82M/s** |

The current decoder remains substantially faster than the original implementation while the complete uncompressed browser bundle is below 50,000 bytes.

## Decision fidelity

Compared against the original exact chain across the full JSON fixture corpus:

- 75,172 exact nonzero transitions audited
- 7,616 unique property names measured
- 32,745 map-inference score gates compared
- **0 changed score-gate decisions**
- explicit regressions cover the closest class-side and map-side boundaries

The compact name scores have a median multiplicative error of 1.12x and p95 of 1.52x, but the calibrated representation preserves every observed decision made by the original chain.

## Validation

- `npm run build`
- `npm run test:unit` (75 tests)
- `CPUs=2 QUICKTEST=true FIXTURE=javascript npm run test:fixtures` (56 fixtures)
- focused JavaScript fixtures for `77392.json`, `617e8.json`, and `recursive.json`
- `npm run benchmark:markov-chain`
- `npm run benchmark:markov-chain:fidelity -- /tmp/markov.json test/inputs/json`
- Biome checks
- encoder regeneration idempotence


packages/quicktype-core/src/MarkovChain.ts
```diff
@@ -1,67 +1,251 @@
 import { encodedMarkovChain } from "./EncodedMarkovChain.js";
-import { assert, inflateBase64, panic } from "./support/Support.js";
```
schani, line 2: Use `assert` and `panic` from `Support.js`!!!!!

packages/quicktype-core/src/MarkovChain.ts
```diff
@@ -176,28 +382,27 @@ function testWord(mc: MarkovChain, word: string): void {
 
 export function test(): void {
```
schani: This should completely move to a unit test, no?

PR #2939, opened 2026-07-13
fix(core): enforce rendererOptions keys in the type system

### LINKED ISSUE #2933: [BUG]: quicktype-core silently ignores unknown rendererOptions keys
## Issue Type

quicktype-core API (developer experience).

Input Format: JSON
Output Language: any

CLI, npm, or app.quicktype.io: npm (`quicktype-core`)
Version: 25.0.0-pre1 (also reproduces on 24.0.2)

## Description

Passing an unrecognized key in `rendererOptions` is silently ignored — no error, no warning. A typo (or an option name that exists for a different language) produces default output with no signal that the option had no effect:

```js
const { quicktype, InputData, jsonInputForTargetLanguage } = require("quicktype-core");

const jsonInput = jsonInputForTargetLanguage("typescript");
await jsonInput.addSource({ name: "X", samples: ['{"a": 1}'] });
inputData.addInput(jsonInput);

// "totally-bogus-option" is accepted without complaint
const result = await quicktype({
    inputData,
    lang: "typescript",
    rendererOptions: { "totally-bogus-option": "yes" },
});
```

The CLI does the right thing here (`Error: Option parsing failed: Unknown option: --just-typs.`), so the core API is strictly less safe than the CLI for the same mistake. This bit me for real while testing: I passed C#'s `just-types` spelling to a language that names it differently and got silently un-customized output.

## Expected Behavior

`quicktype()` should throw (or at minimum warn via `debugPrintTimes`-style logging) when a `rendererOptions` key doesn't match any option of the target language. Since `RendererOptions` is already typed per-language in newer versions, a runtime check would also catch untyped/dynamic callers (plain JS, options loaded from config files).

🤖 Filed with [Claude Code](https://claude.com/claude-code)

schani:
Fixes #2933.

Unknown `rendererOptions` keys were silently ignored by the quicktype-core API. Per review, this is now enforced **entirely in the type system** — no runtime validation.

## What this does

- Makes `quicktype()`, `quicktypeMultiFile()`, and `quicktypeMultiFileSync()` generic over `Lang extends LanguageName` (inferred from `lang`), and carries the generic through `Options`. With a literal `lang`, `rendererOptions` is typed as that language's option map, so unknown option names and invalid enum values are compile errors:

```ts
await quicktype({
    inputData,
    lang: "csharp",
    rendererOptions: { "totally-bogus": "x" },  // error TS2353: unknown property
});
```

- Fixes `OptionKey`, which fell through to the `Option` class type for non-enum options — even valid usage like `{ namespace: "Acme" }` failed to typecheck before this. String options are now `string`, boolean options `boolean | "true" | "false"` (matching what `BooleanOption.getValue` accepts), enum options keep their key union.

- Callers that don't pin `lang` to a literal — or omit it — keep the old, unchecked behavior, so plain-JS and dynamic callers are unaffected.

## Testing

- `test/unit/renderer-options.test-d.ts` covers valid options, unknown names, another language's option name, invalid enum values, boolean string forms, and the permissive non-literal fallback, using `@ts-expect-error` assertions run by vitest's typecheck mode (newly enabled in `vitest.config.ts`). Verified the harness fails when an expected error doesn't occur.
- The dead renderer options this work uncovered in `test/languages.ts` (C++ `unions`, TypeScript/Flow `declare-unions`, Zod/Effect `array-type`, PropTypes `runtime-typecheck`) stay removed.

### DIFF
--- a/packages/quicktype-core/src/RendererOptions/types.ts
+++ b/packages/quicktype-core/src/RendererOptions/types.ts
@@ -54,7 +54,13 @@ export type OptionValue<O> =
 export type OptionKey<O> =
     O extends EnumOption<string, Record<string, unknown>, infer EnumKey>
         ? EnumKey
-        : O;
+        : O extends Option<string, infer Value>
+          ? Value extends boolean
+              ? // `BooleanOption.getValue` also accepts the strings
+                // "true" and "false", which is what the CLI passes.
+                boolean | `${boolean}`
+              : Value
+          : never;
 
 // FIXME: Merge these and use camelCase user-facing keys (v24)
 export type OptionMap<T> = {
--- a/packages/quicktype-core/src/Run.ts
+++ b/packages/quicktype-core/src/Run.ts
@@ -113,10 +113,11 @@ export interface NonInferenceOptions<Lang extends LanguageName = LanguageName> {
      */
     outputFilename: string;
     /** Options for the target language's renderer */
-    rendererOptions: RendererOptions<Lang>;
+    rendererOptions: Partial<RendererOptions<Lang>>;
 }
 
-export type Options = NonInferenceOptions & InferenceFlags;
+export type Options<Lang extends LanguageName = LanguageName> =
+    NonInferenceOptions<Lang> & InferenceFlags;
 
 const defaultOptions: NonInferenceOptions = {
     lang: "ts",
@@ -600,15 +601,15 @@ class Run implements RunContext {
  * @param options Partial options.  For options that are not defined, the
  * defaults will be used.
  */
-export async function quicktypeMultiFile(
-    options: Partial<Options>,
-): Promise<MultiFileRenderResult> {
+export async function quicktypeMultiFile<
+    Lang extends LanguageName = LanguageName,
+>(options: Partial<Options<Lang>>): Promise<MultiFileRenderResult> {
     return await new Run(options).run();
 }
 
-export function quicktypeMultiFileSync(
-    options: Partial<Options>,
-): MultiFileRenderResult {
+export function quicktypeMultiFileSync<
+    Lang extends LanguageName = LanguageName,
+>(options: Partial<Options<Lang>>): MultiFileRenderResult {
     return new Run(options).runSync();
 }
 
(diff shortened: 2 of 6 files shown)


schani:
We shouldn't be validating this at runtime! It should just be a type system thing. If that's not possible, then let's scrap this PR.

schani:
It's possible — reworked in f306fa6 as a pure type-system check; the runtime validation is gone.

The pieces were mostly already there: `RendererOptions<Lang>` is a per-language mapped type, and TypeScript's excess-property checking rejects unknown keys in it. What was missing:

1. **`quicktype()` never inferred `Lang`** — `Options` dropped the generic that `NonInferenceOptions<Lang>` already had, so `rendererOptions` was always typed against the all-languages union, which collapses to an unchecked object. `quicktype`, `quicktypeMultiFile`, and `quicktypeMultiFileSync` are now generic, so a literal `lang: "csharp"` narrows `rendererOptions` to C#'s options and `{"totally-bogus": "x"}` is a compile error — as is an invalid enum value like `framework: "GSON"`.
2. **`OptionKey` was broken for non-enum options** — it fell through to the `Option` class type itself, so even the *valid* `{ namespace: "Acme" }` didn't typecheck. Now string options are `string` and boolean options are `boolean | "true" | "false"` (matching `BooleanOption.getValue`).

Callers that don't pin `lang` to a literal (or omit it) keep the old unchecked behavior, so plain-JS and dynamic callers are unaffected — they just don't get the safety, which is the nature of a type-system-only check.

The type-level behavior is tested with vitest's typecheck mode (`*.test-d.ts` with `@ts-expect-error` assertions); I verified the harness fails when an expected error doesn't materialize. The `test/languages.ts` cleanups stay: those dead options (`declare-unions`, C++ `unions`, Zod/Effect `array-type`, PropTypes `runtime-typecheck`) were real findings.

schani:
Review result: no correctness findings in the type-level implementation.

I verified that the PR builds cleanly and that all 78 unit tests plus the new typechecking suite pass. The GitHub Test workflow is also successful.

One non-blocking cleanup: `DriverNoSamplesForTopLevel` remains in `Messages.ts` but has no callers, apparently left over from the removed runtime-validation implementation. It can be deleted before merge, but I would not block the PR on it.

schani:
Good catch on the symptom, but the message isn't actually dead — `DriverNoSamplesForTopLevel` belongs to the merged empty-samples fix (#2937, commit ea22d3d), which calls it twice in `Inputs.ts`. This branch was just based on a pre-merge master, and my revert of `Messages.ts` restored the file from post-merge master, so the branch showed the message without its caller.

Fixed properly in e1a1404 by rebasing onto current master (and squashing out the discarded runtime-validation commit): the branch no longer touches `Messages.ts` at all, and the message has its callers. All unit + type tests pass.

PR #2914, opened 2026-07-12
Fix nested intersections in union flattening

ankrgyl:
## Description

Fix union flattening when an intersection is nested inside another union.

`FlattenUnions` previously checked only whether the types being unified were themselves `IntersectionType` instances. An intersection nested inside a `UnionType` was therefore passed to `unifyTypes`, eventually causing `Internal error: Unknown type intersection`.

The fix recursively checks union members for intersections and uses the existing reconstitution path when one is found.

## Related Issue

Fixes #2913

## Previous Behaviour / Output

Returned an error

## New Behaviour / Output

Generates the types

## How Has This Been Tested?

Added a unit test with the repro from #2913 and ran it on the full OpenAI OpenAPI spec

PR #2944, opened 2026-07-14
Fix nested intersections in union flattening

### LINKED ISSUE #2913: FlattenUnions fails when unifying an array element union that indirectly contains an intersection.
This shows up while trying to convert the latest OpenAI specs.

## Issue Type

input parsing

- **Environment:** macOS, Apple Silicon, Node.js v24.15.0
- **Quicktype versions:** 23.3.25 and 24.0.2
- **Also reproduced on:** upstream `master` at commit `7a86b5e1042f7bc387876fbbb0c0547d2ca22cbc`
- **Input language:** JSON Schema draft-07
- **Output language:** Rust

## Description

I'm trying to parse OpenAI's OpenAPI spec and create generated types

## Input Data

```json
{
  "type": "object",
  "properties": {
    "input": {
      "type": "array",
      "items": {
        "oneOf": [
          { "$ref": "#/definitions/Message" },
          { "$ref": "#/definitions/Item" }
        ]
      }
    }
  },
  "definitions": {
    "Message": {
      "type": "object",
      "properties": {
        "content": { "type": "string" }
      },
      "required": ["content"]
    },
    "Item": {
      "oneOf": [
        { "$ref": "#/definitions/FunctionOutput" },
        { "$ref": "#/definitions/ReasoningItem" }
      ]
    },
    "FunctionOutput": {
      "allOf": [
        { "$ref": "#/definitions/BaseItem" },
        {
          "type": "object",
          "properties": {
            "output": { "type": "string" }
          },
          "required": ["output"]
        }
      ]
    },
    "BaseItem": {
      "type": "object",
      "properties": {
        "id": { "type": "string" }
      },
      "required": ["id"]
    },
    "ReasoningItem": {
      "type": "object",
      "properties": {
        "summary": { "type": "string" }
      },
      "required": ["summary"]
    }
  }
}
```

Something like

```rust
#[derive(Serialize, Deserialize)]
pub struct TopLevel {
    pub input: Option<Vec<Item>>,
}

#[derive(Serialize, Deserialize)]
pub struct Item {
    pub content: Option<String>,
    pub id: Option<String>,
    pub output: Option<String>,
    pub summary: Option<String>,
}
```

## Current Behaviour / Output

```
Error: Internal error: Unknown type intersection.
```

## Steps to Reproduce

Save the above json schema to `nested-intersection.schema.json` and then run

```
npx quicktype@24.0.2 \
  --src-lang schema \
  --lang rust \
  nested-intersection.schema.json
```

## Possible Solution

Will send a PR

### LINKED ISSUE #2941: Kestra 0.19 schema fails with Unknown type intersection for all output languages
## Summary

quicktype fails to process the real-world Kestra 0.19 JSON Schema with:

    Error: Internal error: Unknown type intersection.

The same failure occurs for both TypeScript and Rust output, so it appears to happen before renderer-specific code generation.

This is likely related to #2913. I am reporting it separately because Kestra is a large, real-world schema that can serve as a regression and performance test after the underlying intersection issue is fixed.

## Input schema

- Hosted benchmark copy: https://hostr.flingit.run/s/quicktype-repros/kestra-0.19.0.schema.json
- Upstream SchemaStore source: https://raw.githubusercontent.com/SchemaStore/schemastore/master/src/schemas/json/kestra-0.19.0.json
- Size: 7,597,597 bytes
- SHA-256: d2e7749043c56583f2258002cbd905e868ca656e484c15aaac5403945cb5f726

The hosted copy was downloaded again after upload and its size and SHA-256 were verified against the benchmark input.

## Steps to reproduce

    node dist/index.js --lang typescript --src-lang schema --out /tmp/kestra.ts kestra-0.19.0.schema.json

The Rust renderer fails identically:

    node dist/index.js --lang rust --src-lang schema --out /tmp/kestra.rs kestra-0.19.0.schema.json

## Actual result

TypeScript:

    Error: Internal error: Unknown type intersection.
    real 4.38 seconds

Rust:

    Error: Internal error: Unknown type intersection.
    real 4.39 seconds

## Expected result

quicktype should generate output successfully. If this schema construct is intentionally unsupported, it should return an actionable schema diagnostic that identifies the unsupported construct or location rather than an internal error.

## Environment

- Repository: glideapps/quicktype
- Commit: f103f0a3864c58d0925791db928a26aba3ad02c8
- Branch: master
- quicktype version reported by the built CLI: 24.0.0
- Node.js: 24.6.0
- macOS: 14.8.3
- Architecture: arm64

## Benchmark context

This was discovered while evaluating large real-world JSON Schemas for whole-pipeline performance coverage. The Kestra schema is the largest representative schema in that set, but no phase breakdown can currently be collected because the pipeline aborts with the internal error.

schani:
## Credit

The core fix in this PR was authored by @ankrgyl (Ankur Goyal) in #2914. This PR is a maintainer-managed continuation of that work: all four commits were cherry-picked from the original branch with authorship preserved (the fix commit is still authored by Ankur Goyal). Thank you @ankrgyl!

## The bug

`flattenUnions` only checked whether the types being unified were themselves `IntersectionType` instances. An intersection nested inside a `UnionType` member slipped through to `unifyTypes`, which then failed with:

```
Internal error: Unknown type intersection.
```

This is easy to hit with JSON Schema inputs where a `oneOf` branch produces a union whose member is an intersection (e.g. `allOf` inside `anyOf`/`oneOf`), and it also occurs on large real-world schemas such as the Kestra 0.19.0 JSON Schema (#2941).

## The fix

`flattenUnions` (in `packages/quicktype-core/src/rewrites/FlattenUnions.ts`) now recursively checks union members for nested intersections and, when one is found, takes the existing reconstitution path (deferring flattening to a later pass) instead of attempting to unify a set that still contains an intersection.

Additional commits convert the original unit test into a schema fixture (`test/inputs/schema/nested-intersection-union.schema`) and skip that fixture for the cjson and cplusplus test targets, which have known limitations with these inputs.

Fixes #2913
Fixes #2941

## Verification

- `npm run build` succeeds and `npm run test:unit` passes (71 tests, 9 files).
- `quicktype --lang typescript --src-lang schema test/inputs/schema/nested-intersection-union.schema` generates successfully; on unpatched master the same invocation fails with `Internal error: Unknown type intersection.`
- Verified against the 7.6MB real-world Kestra 0.19.0 schema from #2941: both TypeScript and Rust output generate successfully in ~10s with the patch applied.


ankrgyl:
Thanks for doing this 🙏

PR #2947, opened 2026-07-14
feat(swift): add final classes option

### LINKED ISSUE #1241: Please generate "final" Swift classes
All of the generated Swift classes should be "final class" for both performance and clarity:

https://forums.swift.org/t/how-effective-is-final-for-code-optimization/24934

It's easier to remove "final" if you should need to subclass one of the classes, than it is to add "final" to them all :-)

coyaSONG:
## Description

Add an opt-in `--final-classes` Swift renderer option. When enabled, it emits every generated Swift reference type as a `final class`:

- models explicitly emitted as classes via `--struct-or-class class`
- models promoted to classes to break recursive value-type cycles
- the generated `JSONNull`, `JSONCodingKey`, and `JSONAny` Codable helpers

The option defaults to `false`, so existing Swift output remains unchanged. Struct output is unchanged in either mode.

## Related Issue

Closes #1241

## Motivation and Context

Final classes communicate that generated models are not intended as subclassing surfaces and can avoid dynamic subclass dispatch. Making the behavior opt-in provides that benefit without introducing a breaking change for users who currently subclass generated types.

The renderer previously had no option for this behavior: it selected only `class` or `struct` for model declarations and hard-coded bare `class` declarations for helper types.

```swift
class Person: Codable {
    // ...
}

class JSONAny: Codable {
    // ...
}
```

## New Behaviour / Output

With `--final-classes`:

```swift
final class Person: Codable {
    // ...
}

final class JSONAny: Codable {
    // ...
}
```

## How Has This Been Tested?

- focused final-class tests: 3/3 passed, including default-off compatibility
- `npm run test:unit`: 17 files and 130 tests passed with no type errors
- `FIXTURE=swift npm run test:fixtures`: all 241 Swift fixture runs passed with Swift 6.2.3, including the explicit `struct-or-class: class, final-classes: true` variant
- `npm run build`
- Biome on all four changed files
- `git diff --check`
- manual CLI generation confirmed bare `class` by default and `final class` with `--final-classes`

The repository's requested Node 24.6.0 was not installed in this environment; validation used Node 25.6.1. Implementation and validation were assisted by OpenAI Codex and reviewed against the repository guidance.

Not applicable; this changes generated source output.

### DIFF
--- a/packages/quicktype-core/src/language/Swift/SwiftRenderer.ts
+++ b/packages/quicktype-core/src/language/Swift/SwiftRenderer.ts
@@ -500,6 +500,7 @@ export class SwiftRenderer extends ConvenienceRenderer {
 
         const isClass = this._options.useClasses || this.isCycleBreakerType(c);
         const structOrClass = isClass ? "class" : "struct";
+        const finalPrefix = isClass ? "final " : "";
 
         if (isClass && this._options.objcSupport) {
             // [Michael Fey (@MrRooni), 2019-4-24] Swift 5 or greater, must come before the access declaration for the class.
@@ -508,6 +509,7 @@ export class SwiftRenderer extends ConvenienceRenderer {
 
         this.emitBlockWithAccess(
             [
+                finalPrefix,
                 structOrClass,
                 " ",
                 className,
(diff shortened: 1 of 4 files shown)


schani:
Could you put this behind an option that's default off please?  It's a breaking change otherwise.

PR #2959, opened 2026-07-20
Modernize Elm target to 0.19

schani:
Modernizes the Elm target from Elm 0.18 to **Elm 0.19** and brings the Elm fixtures back into CI.

## Breaking: Elm 0.18 output is removed

The renderer now emits Elm 0.19 code only. There was never a version option, so nothing was deleted from the options surface — the generated code simply targets 0.19 now, and 0.18 projects can stay on older quicktype releases.

## Generated-output changes

- Decode pipelines are seeded with `Jdec.succeed Foo` instead of `Jpipe.decode Foo` (`decode` was removed in NoRedInk/elm-json-decode-pipeline 1.0; install instructions in the header comment updated accordingly).
- The 0.19 elm/json encoder API (`Json.Encode.list`/`array` take the element encoder, `Json.Encode.dict` exists) makes the generated `makeArrayEncoder`/`makeListEncoder`/`makeDictEncoder` helpers unnecessary — they are gone; only `makeNullableEncoder` remains.
- Imports no longer expose `map`/`toList` (ambiguous under 0.19 shadowing rules): `import Dict exposing (Dict)`, plus `import Array exposing (Array)` only with `--array-type array`; `List` needs no import.
- Elm 0.19 forbids parameters shadowing top-level definitions, so the parameter names used by generated functions (`x`, `y`, `r`, `f`, `m`, `str`, `somethingElse`) are now forbidden as generated global identifiers (caught by `priority/unions.json`, which has a type named `x`).
- Generated identifiers can no longer start with `_` (an 0.19 syntax error, caught by `blns-object.json`).
- String literals escape non-printables with 0.19's `\u{XXXX}` syntax instead of `\uXXXX`/`\UXXXXXXXX` (caught by `identifiers.json` / `nst-test-suite.json`).

## `array-type` now defaults to `list`

Defaults-audit recommendation #13, deferred until this modernization: `List` is the idiomatic Elm collection, and with the 0.19 encoder API the `list` path needs no helper at all. `--array-type array` still works and stays covered in CI by a quicktest entry (`quickTestRendererOptions`).

## Fixture revival

The driver in `test/fixtures/elm/` is rewritten for 0.19:

- `elm-package.json` → `elm.json` (elm/core, elm/json, NoRedInk/elm-json-decode-pipeline 1.0.1).
- `Main.elm` uses `Platform.worker` (`Platform.program` is gone) and `Jdec.errorToString` for decode errors.
- `runner.js` uses the 0.19 embedding API (`Elm.Main.init()` instead of `Elm.Main.worker()`).
- A `Warmup.elm` module is compiled once by the setup command so all package dependencies are downloaded/built into the shared `ELM_HOME` cache before per-sample compiles run in parallel (elm 0.19.1 can corrupt its package registry on concurrent cold-cache downloads). The `sysconfcpus` hack for the 0.18 compiler is gone.

## Skip lists (every entry re-verified against elm 0.19.1)

Un-skipped (fixed by the renderer changes above): `identifiers.json`, `simple-identifiers.json`, `blns-object.json`, `nst-test-suite.json`, `keywords.json`.

Still skipped, with verified reasons:

- `recursive.json`, `direct-recursive.json`, `bug427.json`, `bug790.json`, `list.json` and the recursion schemas (`union-list`, `list`, `ref-remote`, `mutually-recursive`, `postman-collection`, `vega-lite`, `simple-ref`, plus newly `recursive-union-flattening`): Elm type aliases cannot be recursive.
- `constructor.schema`, `keyword-unions.schema`: compile and run now, but elm/json's field decoder uses the JS `in` operator, which finds inherited `Object.prototype` members — decoding `{}` yields a non-null `"constructor"`.
- `nested-intersection-union.schema`: the generated decoder accepts invalid union members because all class properties decode via `Jpipe.optional` (same limitation as documented for C++).
- The five newly enabled identifier-heavy JSON inputs are in `skipDiffViaSchema` (schema round-trip changes inferred names/types); all 20 pre-existing `skipDiffViaSchema` entries were re-checked and still differ.

## CI

`- elm,schema-elm` is re-enabled in the test matrix; the workflow's elm 0.19.1 install step already existed. Locally validated with elm 0.19.1: the **full** corpus (215 JSON tests incl. misc, 67 schema tests) passes for both `FIXTURE=elm` and `FIXTURE=schema-elm`, as do `npm run test:unit` and `npm run lint`.


schani:
Follow-up: the toolchain moved to elm 0.19.2 in #2960 (0.19.2 refuses `"elm-version": "0.19.1"` applications and the release assets were renamed; no language changes, and the cold-cache warmup is still needed — the corruption reproduces with 0.19.2).

PR #2969, opened 2026-07-20
fix(haskell): don't over-rename enum cases when the enum suffix already disambiguates

### LINKED ISSUE #2868: [BUG]: Forbidden identifiers too strict for Haskell enums
## Issue Type

Quicktype output

Input Format: JSON Schema
Output Language: Haskell

CLI, npm, or app.quicktype.io: CLI and app.quicktype.io
Version:

## Description

`Error` is marked as a forbidden global identifier, so enums named `error` are renamed to avoid clashing, but this is unnecessary for enums, which are already suffixed.

cc @algas

## Input Data

```json
{
  "id": "http://json-schema.org/geo",
  "description": "A geographical coordinate",
  "type": "string",
  "enum": ["ok", "error"]
}
```

```hs
data Health
    = ErrorHealth
    | OkHealth
    deriving (Show)
```

## Current Behaviour / Output

```hs
data Health
    = HealthErrorHealth
    | OkHealth
    deriving (Show)
```

## Possible Solution

1. Aeson imports should be excluded from the forbidden list; the generator should import `Data.Aeson` qualified
2. Forbidden identifiers should be renamed _only_ when it exactly matches. If a suffix is being added, it should _not_ be renamed

schani:
## Bug

For a string enum like `["ok", "error"]`, the Haskell renderer produced
`HealthErrorHealth | OkHealth` instead of the expected `ErrorHealth | OkHealth`.

## Root cause

`Error` is listed as a globally forbidden identifier in
`packages/quicktype-core/src/language/Haskell/constants.ts`. When
`ConvenienceRenderer.makeNameForEnumCase` styles the enum case name to
`Error`, it detects a collision with that forbidden global identifier and
picks an alternative (`Health_error` -> `HealthError`) *before* the
enum-name suffix (`Health`) is textually appended at emit time in
`HaskellRenderer.emitEnumDefinition`. Since the suffix is only appended
after name resolution runs, the resolver never sees the final,
already-disambiguated identifier (`ErrorHealth`) when checking for
forbidden-name collisions, so it picks the uglier alternative unnecessarily.
The generated code was correct and round-tripped fine — this was purely
cosmetic.

## Fix

`HaskellRenderer` now overrides `makeNameForEnumCase` to fold the enum-name
suffix into the proposed name itself, via a `DependencyName` (the same
pattern already used elsewhere for names that depend on another name, e.g.
`proposeUnionMemberName`). This means forbidden-word collision checking now
sees the final, suffixed identifier (`ErrorHealth`) rather than the bare
case name (`Error`), so it no longer needlessly renames it.

Emit sites (`emitEnumDefinition`, the `ToJSON`/`FromJSON` instances) now use
a small `enumCaseName` helper: for the common case (a `DependencyName` that
already includes the suffix) it's used as-is; for `FixedName`s (explicit
user-assigned accessor names, which bypass `makeNameForEnumCase`) it falls
back to the previous behavior of textually appending the enum name, since
those names are user-specified and shouldn't be silently changed.

## Test coverage

Added a new JSON Schema fixture:
- `test/inputs/schema/haskell-enum-forbidden.schema` — an object with a
  string enum `["ok", "error"]`, i.e. the schema from the issue.
- `test/inputs/schema/haskell-enum-forbidden.1.json` — a matching sample.

Round-tripping alone can't distinguish generated constructor names (both
`HealthErrorHealth` and `ErrorHealth` round-trip correctly), so
`HaskellLanguage.runCommand` in `test/languages.ts` was extended to also
`grep` the generated `QuickType.hs` for the expected `ErrorHealth`
constructor when running this specific sample, failing the fixture if the
old, over-renamed identifier reappears.

## Verification

- `npm run build` passes.
- `npm run test:unit` passes (163 tests).
- Confirmed the exact repro from the issue now generates `ErrorHealth | OkHealth` instead of `HealthErrorHealth | OkHealth`.
- Ran the Haskell code generator against several other enum-bearing schema fixtures (`enum.schema`, `enum-with-values.schema`, `enum-large.schema`, `optional-enum.schema`, `keyword-enum.schema`, `top-level-enum.schema`, `multi-type-enum.schema`, `cut-enum.schema`) to check for regressions — all generate without error and with sensible names.
- `biome check` passes on the changed files.
- The full `FIXTURE=schema-haskell script/test` run (which compiles and executes the generated Haskell via `stack`) was not run locally because the Haskell toolchain (`stack`/`ghc`) is not installed in this environment; CI will validate this.

Fixes #2868


test/languages.ts
```diff
@@ -1730,7 +1691,11 @@ export const HaskellLanguage: Language = {
     setupCommand: "stack install",
     compileCommand: "true",
     runCommand(sample: string) {
-        return `stack run haskell -- "${sample}"`;
+        // Round-tripping cannot distinguish generated constructor names.
+        const sourceCheck = sample.endsWith("haskell-enum-forbidden.1.json")
```
schani: This is a shitty way to test this. Figure out a principled, better way.

schani:
Retriggering CI (no CI run was created for the latest push).

test/languages.ts
schani: Agreed, and thanks for flagging it. I removed the filename-matching branch from the Haskell runCommand. That grep was also dead code by now: b75e57f1 added haskell-enum-forbidden.schema to skipsEnumValueValidation, and since the Haskell driver reports a failed decode as "null" with exit 0 its fail sample can't be detected, so the whole schema is skipped for the Haskell fixture and the grep never ran.

In its place ee29122f adds test/unit/haskell-enum-case-naming.test.ts, which renders the schema to Haskell and asserts the constructors are OkHealth/ErrorHealth rather than the over-renamed HealthErrorHealth. A round-trip fixture genuinely can't cover this, because the constructor name is invisible to JSON serialization, so both the correct and the over-renamed output round-trip identically. The unit test fails against the pre-fix renderer and passes against the fix.

Also resolved the master conflict in test/languages.ts (kept both additions to skipsEnumValueValidation).

PR #2979, opened 2026-07-20
fix(schema): respect unevaluatedProperties: false

### LINKED ISSUE #2372: `"unevaluatedProperties": false` not working
```json
{
  "type": "object",
  "properties": {
    "string": {
      "type": "string"
    }
  },
  "unevaluatedProperties": false
}
```

results in

```typescript
export interface District {
    string?: string;
    [property: string]: any;
}
```

while

```json
{
  "type": "object",
  "properties": {
    "string": {
      "type": "string"
    }
  },
  "additionalProperties": false
}
```

results correctly in

```typescript
export interface District {
    string?: string;
}
```

schani:
## Bug

`unevaluatedProperties: false` in a JSON Schema was silently ignored by quicktype. The equivalent `additionalProperties: false` correctly closes the generated interface, while `unevaluatedProperties: false` produced an open index signature instead.

Repro:

```json
{
  "type": "object",
  "properties": { "foo": { "type": "string" } },
  "unevaluatedProperties": false
}
```

Expected TypeScript:

```typescript
export interface Obj {
    foo?: string;
}
```

## Fix

Master now contains the additional-properties fallback that handles the reported typed-object case. This PR completes that support by including `unevaluatedProperties` in the object/union-detection predicate, matching `additionalProperties`. This matters for schemas where `unevaluatedProperties` is the only object-applicator keyword: non-object values remain allowed by JSON Schema, but the object branch is represented and respects the closed-object constraint.

This is a pragmatic, local implementation; it does not implement full JSON Schema evaluation semantics across `allOf`/`$ref` sibling subschemas.

## Test coverage

The existing cross-language `unevaluated-properties.schema` fixture now references a keyword-only `Closed` subschema with `unevaluatedProperties: false`. This makes the change visible in generated output and exercises it through the normal fixture compilation/round-trip pipeline. The fixture retains its positive samples and expected-failure sample for the schema-valued form of `unevaluatedProperties`.

## Verification

- Current generated-output report: 26 files, all for `unevaluated-properties.schema`; no unrelated fixtures or languages.
- Current GitHub Actions test workflow passes all jobs.
- `npm run build`
- `npm run test:unit`
- `CPUs=4 FIXTURE=schema-typescript,schema-rust,schema-golang,schema-javascript,schema-python npm run test:fixtures -- test/inputs/schema/unevaluated-properties.schema`
- `npm run lint`

Fixes #2372


schani:
Why does this not have a fixture test, instead of the unit test?

PR #2987, opened 2026-07-20
fix(input): infer Double for integral float literals in string-based JSON input

### LINKED ISSUE #2195: Using web interface creates swift Codable with type Int for 0.0 value
### Steps to reproduce:
1. Open https://app.quicktype.io/
2. Enter this JSON as the model:
```
{
 "double_value": 0.0      
}
```
3. Generated codable:
```
struct CustomModel: Codable {
    let doubleValue: Int

        case doubleValue = "double_value"
    }
}
```

I would expect to get:
```
let doubleValue: Double
```
and it works correctly when value is different from `0.0`

schani:
## Bug

String-based JSON input (`jsonInputForTargetLanguage` / `addSource({ samples: [<string>] })`), as used by the web app, parsed samples with `JSON.parse`. That discarded the source distinction between `0` and integral-valued floating-point literals such as `0.0`, `2.0`, and `2e0`, so those literals were incorrectly inferred as integers.

```json
{ "double_value": 0.0 }
```

Swift therefore generated `let doubleValue: Int` instead of `Double`. The issue affects every target that distinguishes integers from floating-point numbers.

## Fix

Both JSON input paths now use stream-json's source-preserving tokenizer and the same token-to-`CompressedJSON` handling:

- `CompressedJSONFromString` drives stream-json's synchronous tokenizer directly.
- `CompressedJSONFromStream` keeps the streaming adapter, but delegates token handling to the shared implementation in `CompressedJSON`.

Number literals containing `.`, `e`, or `E` are therefore classified as floating point in both paths. Plain integer literals continue to use the target language's supported integer range. This also removes the previous double-parse/marker workaround.

Because stream-json 3.x is ESM-only and quicktype-core publishes CommonJS, quicktype-core's Node floor is aligned with the root package at Node 20.19, the first Node 20 release supporting `require(esm)`.

## Test coverage

- Added `test/inputs/json/priority/bug2195.json`, exercising the issue's exact `0.0` input through the fixture harness and all JSON target renderers.
- Kept API-level unit coverage for string input: `0.0`, `2.0`, and `2e0` infer as Swift `Double`; plain `2` remains `Int`; and quoted `"3.0"` remains `String`.
- Existing streaming and string-input integer-range boundary tests cover fixed-size and arbitrary-precision targets.

## Verification

- `npm run build`
- `npm run test:unit` (219 tests)
- `npm run lint`
- `QUICKTEST=true FIXTURE=javascript CPUs=2 npm run test:fixtures -- test/inputs/json/priority/bug2195.json`
- Output-only fixture generation for all 35 JSON fixture variants (34 generated files); Swift output contains `let doubleValue: Double`.

The local environment does not have `swiftc`, so the full Swift fixture could not compile locally; CI provides target-language fixture coverage.

Fixes #2195


schani:
This is a super hacky way of solving this problem! Why can't we use the same JSON parsing for both paths?  Also, this should have a fixture test case!

schani:
## Performance benchmark results

Compared PR `d7fa908a` with merge base `a49b94e8` on Node v24.6.0 / Linux x64 / AMD Ryzen 7 5825U. I ran both the synthetic pipeline benchmark and canonical real-world benchmark twice, reversing commit order for the second pass.

### Synthetic benchmark (reverse-order repeat)

| Case | PR median change | Heap change |
|---|---:|---:|
| Small JSON | ~0% | +0–2% |
| Large JSON / TypeScript | **+44.1%** | +13.9% |
| Large JSON / Rust | **+45.6%** | +17.6% |
| JSON Schema cases | -1.5% to -3.5% | ~unchanged |

Large JSON parsing itself became approximately **2× slower**.

### Canonical benchmark (reverse-order repeat)

| Input | TypeScript | Rust | Heap impact |
|---|---:|---:|---:|
| USGS earthquakes | **+61.4%** | **+63.1%** | +49–75% |
| GitHub REST OpenAPI | +8.4% | +11.7% | +6–23% |
| NVD CVE 2024 | **+65.8%** | **+53.4%** | +89–98% |
| HL7 FHIR R5 schema | +2.5% | +1.8% | mostly unchanged |
| Kestra 0.19 schema | +2.3% | +1.7% | unchanged |

The first pass showed the same pattern:

- USGS: +58.4% / +63.6% (TypeScript / Rust)
- GitHub: +6.4% / +6.1%
- NVD: +69.8% / +81.9%
- FHIR: -0.0% / +0.4%
- Kestra: +1.1% / +1.4%

**Conclusion:** this PR introduces a substantial performance and memory regression for large string-based JSON input. The regression is concentrated in parsing; schema workloads are effectively unchanged.

PR #2989, opened 2026-07-20
fix(schema): keep distinctly-named enums separate when cases match

### LINKED ISSUE #2484: Enum schemas aren't generated if the enumerator list matches
I have two JSON schema files, where the type names are different, but the enumerators are matching.
```
{ "title":"firstEnumeration","enum": ["a","b","c"],"type": "string"}
```
```
{ "title":"secondEnumeration","enum": ["a","b","c"],"type": "string"}
```
When I use quicktype to generate source code, I only get output for the `secondEnumeration`, even though they are two different types - there is also no alias type generated for `firstEnumeration`, it is simply an unknown symbol. This problem is observed across multiple output languages, so the logic to blame must be in the JSON schema parsing logic.

schani:
## Bug

Two top-level JSON Schema enums with different titles but an identical case
set (e.g. `["a","b","c"]`) collapsed into a single generated type. The
dropped enum's name was still referenced by generated code (e.g. Go's
`UnmarshalFirstEnumeration`), producing output that doesn't compile.

Repro:
```
node dist/index.js --lang go --src-lang schema first.schema.json second.schema.json
```
```json
// first.schema.json
{ "title":"firstEnumeration","enum": ["a","b","c"],"type": "string"}
// second.schema.json
{ "title":"secondEnumeration","enum": ["a","b","c"],"type": "string"}
```
Before the fix, only `SecondEnumeration` was declared, while
`UnmarshalFirstEnumeration` still referenced the never-defined
`FirstEnumeration`. The same collapse also happens within a single schema
file (two `$ref`'d definitions with different names but the same enum
cases), which is what the new fixture test covers.

## Root cause

`TypeBuilder.getEnumType` interns enum types by `enumTypeIdentity`
(`packages/quicktype-core/src/Type/Type.ts`), which is keyed only on the
case set plus non-identity attributes. An enum's given (explicitly-named,
non-inferred) name was never part of its identity, so two differently
named-but-structurally-identical enums were deduplicated into one type.

## Fix

Added a narrow `enumTypeNames` identity attribute
(`packages/quicktype-core/src/attributes/TypeNames.ts`) that captures only
an enum's explicitly given (non-inferred) name(s). It's:
- attached in `JSONSchemaInput.ts` when converting a schema enum whose only
  included type is `string`, and
- forwarded through `ExpandStrings.ts`, where the actual `EnumType` is
  materialized from the annotated string type (previously this step dropped
  all attributes via `emptyTypeAttributes`, which is also why the identity
  attribute needed threading through here rather than just at the schema
  conversion site).

Enums that share a case set but have distinct given names now stay
distinct. Unnamed/inferred enums with identical cases still dedupe into a
single shared type, same as before — verified by a dedicated regression
test.

## Test coverage

- `test/inputs/schema/distinct-named-enums.schema` (+ `.1.json`): an object
  with two `$ref`'d definitions of different names but identical enum
  cases; runs through the standard schema fixture suite.
- `test/unit/enum-schema-identity.test.ts`: focused checks that (a)
  distinctly-named enums stay separate, and (b) unnamed enums with
  identical cases still dedupe — codifying the non-regression on structural
  dedup that the rest of the type graph relies on.

## Verification

- `npm run build` passes.
- Full unit suite: 165/165 tests pass.
- Full `schema-golang` fixture suite (`QUICKTEST=true FIXTURE=schema-golang script/test`): 68/68 pass, including the new `distinct-named-enums.schema` fixture and the pre-existing `enum-with-values.schema` fixture (which exercises a separate non-identity attribute, `qt-enum-values`, that briefly regressed during development and was fixed by narrowing exactly what `ExpandStrings.ts` forwards into the enum type).
- Manually re-ran both the original two-file repro and the single-file `$ref` repro from the issue; Go output now declares and correctly types both enums and compiles.
- CI will additionally validate the other language fixtures.

Fixes #2484


schani:
Reopening to retrigger CI (push webhook didn't create a run).

PR #2997, opened 2026-07-20
fix(schema): merge array item types across allOf members

### LINKED ISSUE #1840: allOf in conjunction with arrays didn't works as expected (or the json schema spec intends)
This schema works as expected:
<details>
<summary>Working schema - click to expand!</summary>

```
{
     "definition": {
        "person": {
            "type": "object",
            "required": ["age"],
            "properties": {
                "age": {"type": "integer"}
            }
        },
        "persons": {
            "type": "array",
			"items": {
				"type": "object",
				"properties": {
					"AgedPerson": {
						"$ref": "#/definition/person"
					}
				}}}

    }, 
  "$id": "person.json",
  "title": "person",
  "type": "object",
  "properties": {
      "PersonList": { "allOf": [{"$ref": "#/definition/persons"}] }
  }
}
```
</details>

However, if you add a additional schema to the `allOf`-section of the `PersonList` the first schema is no longer considered.

So a schema containing code like the following, only considers the `name`property and not the properties in the referenced schema `persons`.
```
"PersonList": {
  "allOf": [
    {"$ref": "#/definition/persons"},
    {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "name": {
            "type": "string"
}}}}]}
```

<details>
<summary>Schema generating wrong code - click to expand!</summary>

```
{
     "definition": {
        "person": {
            "type": "object",
            "required": ["age"],
            "properties": {
                "age": {"type": "integer"}
            }
        },
        "persons": {
            "type": "array",
			"items": {
				"type": "object",
				"properties": {
					"AgedPerson": {
						"$ref": "#/definition/person"
					}
				}}}

    }, 
  "$id": "person.json",
  "title": "person",
  "type": "object",
  "properties": {
      "PersonList": { "allOf": [{"$ref": "#/definition/persons"}, {
          "type": "array",
					"items": {
						"type": "object",
						"properties": {
							"name": {
								"type": "string"
							}}}}] }
  }
}
```
</details>

Please let me know if I was unclear of if I could contribute additional information.

schani:
## Bug

When a property's JSON Schema is an `allOf` with **more than one** member that each describe an array, quicktype only keeps the array item type from the *last* member and silently drops the item type(s) contributed by earlier members. A single-member `allOf` (or a bare `$ref`) worked correctly, which pointed at the multi-member merge path specifically.

Repro (from the issue): a `PersonList` property defined as
```json
"PersonList": {
  "allOf": [
    { "$ref": "#/definitions/persons" },
    { "type": "array", "items": { "type": "object", "properties": { "name": { "type": "string" } } } }
  ]
}
```
where `#/definitions/persons` is itself an array of objects with an `AgedPerson` property, generated a `PersonList` item type containing only `name` — the `AgedPerson` property from the first `allOf` member was missing.

## Root cause

`IntersectionAccumulator.updateArrayItemTypes` in `packages/quicktype-core/src/rewrites/ResolveIntersections.ts` accumulates the item types of every array-typed schema encountered while resolving an intersection (`allOf`), later unifying them into a single item type. But when initializing that accumulator for the *first* array member, the code created an **empty** `Set` instead of seeding it with that member's own item type:

```ts
if (this._arrayItemTypes === undefined) {
    this._arrayItemTypes = new Set(); // first member's item type never added
} else if (this._arrayItemTypes !== false) {
    this._arrayItemTypes.add(maybeArray.items);
}
```

So the first array's item type was discarded before any subsequent member's item type got added on top of it — only the last (and any middle) members' item types survived.

## Fix

Seed the set with the first array's item type instead of leaving it empty:

```ts
if (this._arrayItemTypes === undefined) {
    this._arrayItemTypes = new Set([maybeArray.items]);
} else if (this._arrayItemTypes !== false) {
    this._arrayItemTypes.add(maybeArray.items);
}
```

## Test coverage

Added an end-to-end JSON Schema fixture reproducing the issue's exact scenario:
- `test/inputs/schema/all-of-array-items.schema` — a `PersonList` property with a two-member `allOf` over array schemas, one reached via `$ref`.
- `test/inputs/schema/all-of-array-items.1.json` — a sample JSON payload exercising properties from both `allOf` members.

This fixture is picked up automatically by every language's schema fixture suite (e.g. `FIXTURE=schema-typescript script/test`), so it round-trips through codegen + compile + deserialize/reserialize for each generator.

## Verification

- `npm run build` passes.
- `npm run test:unit` — 163/163 tests pass.
- Re-ran the issue's exact repro (`node dist/index.js --src-lang schema --lang ts schema.json`): the generated `PersonListElement` interface now correctly contains both `AgedPerson` and `name`.
- Ran the new fixture standalone end-to-end: `FIXTURE=schema-typescript script/test test/inputs/schema/all-of-array-items.schema` — passes (codegen, `tsc` compile, and JSON roundtrip).
- Ran the full `schema-typescript` fixture suite locally (excluding two schema files — `vega-lite.schema` and `class-map-union.schema` — that fail `tsc` compilation on `master` as well, confirmed by reproducing the same failures on the unmodified base commit; these are pre-existing, unrelated issues). All other schema-typescript fixtures pass.
- CI will additionally validate the fixture across all other target languages.

Fixes #1840


schani (commented):
This should have a negative test case!

schani:
Good call. The negative case was added in a0bca09d as `test/inputs/schema/all-of-array-items.1.fail.no-defaults.json` (a `PersonList` entry whose nested `AgedPerson` omits the required `age` field), alongside the positive `.1.json`.

That fail case surfaced a pre-existing enforcement gap in Elixir: like `required.schema`, `intersection.schema`, and `strict-optional.schema`, the generated Elixir code cannot enforce required struct keys at runtime (a missing key is just set to null), so the program never exits nonzero and the expected failure did not occur. Fixed in 09ac3821 by adding `all-of-array-items.schema` to `ElixirLanguage.skipSchema`, in the same block already documenting those three, mirroring the equivalent cjson skip already on this branch. The negative case still runs and enforces under every language that does check required properties (java, csharp, python, and the rest).

PR #3001, opened 2026-07-20
fix(json-schema): keep "cookie" singular when inferring array item names

### LINKED ISSUE #1867: Class names randomly changed from `ie` to `y` suffix
In my generated classes, some type-names are randomly changed from ending in `ie` to ending in `y`.

Uh, yeah, I know - but look:

image

I swear, this is not a prank - some types (not all of them) in my input JSON schema are somehow getting renamed.

It does not appear to have anything to do with the output format - this happens with at least both TypeScript and C# output.

Again, this is not a prank - but maybe some joker introduced this into your codebase as a prank?

I gotta say, it is kind of funny. 😉

What could possibly cause this?

schani:
## Bug

When quicktype infers a name for a JSON Schema array item type (because the
schema doesn't give it an explicit `title`), it singularizes the array
property's name via the `pluralize` npm package. For a `cookies` array
property, this produced a class named `Cooky` instead of `Cookie`.

Repro:
```json
{
  "type": "object",
  "properties": {
    "cookies": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "name": { "type": "string" },
          "value": { "type": "string" }
        }
      }
    }
  }
}
```
```
node dist/index.js --lang cs --src-lang schema schema.json
```
Before: `public partial class Cooky { ... }`
After: `public partial class Cookie { ... }`

## Root cause

`TypeNames.singularize()` (`packages/quicktype-core/src/attributes/TypeNames.ts`)
calls `pluralize.singular(...)`. The `pluralize` package's built-in
singularization rules include a generic `/ies$/ -> y` rule plus a hardcoded
exception list of prefixes that instead map back to `-ie` (`pl`, `zomb`,
`neck?t`, `cross?t`, `coll`, `faer`, `food`, `gen`, `goon`, `group`, `lass`,
`talk`, `goal`, `cut` — which is why `movies`/`zombies`/`pies` singularize
correctly). `cook` isn't in that list, so `cookies` falls through to the
generic rule and becomes `cooky`.

This was reported in #1867; the reporter worked around it in their own code
(since they use the API, not the CLI) with
`pluralize.addIrregularRule("cookie", "cookies")`.

## Fix

Register an explicit singular rule for `cookie` (covering compounds like
`documentCookies` too) on the shared `pluralize` instance quicktype already
depends on, in `packages/quicktype-core/src/attributes/TypeNames.ts`, so this
works out of the box for CLI users as well as API users.

## Test coverage

- `test/inputs/schema/ie-suffix-singularization.schema` +
  `.1.json`: end-to-end JSON Schema fixture with a `cookies` array property,
  run across all `schema-<language>` fixtures.
- `test/unit/ie-suffix-singularization.test.ts`: asserts the generated
  TypeScript identifier is `Cookie`, not `Cooky`. Added as a unit test in
  addition to the fixture because the fixture harness only checks JSON
  round-tripping and can't catch a misspelled class name by itself.

Both were verified to fail (`Cooky`) before the fix and pass (`Cookie`)
after.

## Verification

- `npm run build` passes.
- `npx vitest run test/unit` — all 164 unit tests pass, including the new one.
- Re-ran the CLI repro for both C# and TypeScript output — item class is now
  named `Cookie`.
- Generated TypeScript output for the new fixture input type-checks cleanly
  with `tsc --strict`.
- Attempted a full `QUICKTEST=true FIXTURE=schema-typescript script/test`
  run; it aborts partway through on `test/inputs/schema/vega-lite.schema`
  with a pre-existing, unrelated `tsc` error (verified this also happens on
  unmodified `master`), so the full fixture batch couldn't complete locally.
  The new fixture input itself was verified directly (see above) and CI will
  exercise the full fixture suite across all languages.

Fixes #1867


packages/quicktype-core/src/attributes/TypeNames.ts
```diff
@@ -13,6 +13,11 @@ import { assert, defined, panic } from "../support/Support.js";
 
 import { TypeAttributeKind, type TypeAttributes } from "./TypeAttributes.js";
 
+// `pluralize` otherwise treats "cookies" as ending in the generic `-ies`
+// suffix and produces "cooky".  Keep `cookie` intact, including in compound
+// names such as "documentCookies".
+pluralize.addSingularRule(/(cookie)s$/i, "$1");
```
schani: I don't want this to happen in the global scope. Find a better place!

PR #3009, opened 2026-07-20
fix(js): export all-objects converters from module.exports

### LINKED ISSUE #1655: Javascript all-objects output should export all converters
# Version

quicktype version 15.0.258

# Command

```bash
quicktype \
  --src-lang schema \
  json.schema \
  --lang js \
  --runtime-typecheck \
  --acronym-style pascal \
  --converters all-objects \
  --raw-type any \
  -o output.js
```

# Actual Output

output.js:
```javascript
...
function toData(json) {
    return cast(json, r("Data"));
}

function dataToJson(value) {
    return uncast(value, r("Data"));
}

function toData123(json) {
    return cast(json, r("Data123"));
}

function data123ToJson(value) {
    return uncast(value, r("Data123"));
}
...
module.exports = {
    "dataToJson": dataToJson,
    "toData": toData,
};
```

# Expected Output

Expected output should also export data123ToJson and toData123:

```javascript
...
module.exports = {
    "dataToJson": dataToJson,
    "toData": toData,
    "data123ToJson": data123ToJson,
    "toData123": toData123,
};
```

schani:
## Bug

With `--converters all-objects`, the JavaScript renderer generates
`toX`/`xToJson` converter functions for every object type in the schema, not
just the top-level types. However, `module.exports` only ever listed the
top-level converters, so the extra per-object converters were generated but
inaccessible/unusable to callers of the module.

Repro:
```
node dist/index.js --src-lang schema json.schema --lang js --runtime-typecheck --acronym-style pascal --converters all-objects -o output.js
```
produced `toData123`/`data123ToJson` functions in the file body, but
`module.exports` only contained the top-level `toOutput`/`outputToJson`.

## Root cause

In `packages/quicktype-core/src/language/JavaScript/JavaScriptRenderer.ts`,
`emitConvertModuleBody()` already branches on
`this._jsOptions.converters === ConvertersOptions.AllObjects` to decide
whether to iterate `forEachObject` or `forEachTopLevel` when emitting the
converter functions themselves. `emitModuleExports()` did not have the same
branch — it unconditionally used `forEachTopLevel`, so the export list never
matched what was actually generated when `all-objects` was selected.

## Fix

`emitModuleExports()` now mirrors `emitConvertModuleBody()`'s switch on
`this._jsOptions.converters`, using `forEachObject` for
`ConvertersOptions.AllObjects` and `forEachTopLevel` otherwise (unchanged
default behavior).

## Test coverage

- Added `test/inputs/json/samples/issue-1655.json`, a sample with a nested
  object type, wired into `test/languages.ts`'s
  `JavaScriptLanguage.quickTestRendererOptions` with
  `{ converters: "all-objects" }` (same tuple pattern already used for other
  targeted regression samples like `pokedex.json`).
- Extended `test/fixtures/javascript/main.js` to additionally call the
  nested object's exported converter (`toData123`/`data123ToJson`) when the
  generated source defines it, so the fixture driver actually exercises the
  export and fails with a runtime `TypeError` if it's missing.
- Verified locally that reverting just the renderer fix makes this new
  fixture fail (`TypeError: TopLevel.toData123 is not a function`), and that
  with the fix in place it passes.

## Verification

- `npm run build` passes.
- `QUICKTEST=true FIXTURE=javascript script/test` — all 59 JavaScript
  fixture tests pass, including the new `issue-1655.json` case run with
  `converters: all-objects`.
- Manually re-ran the exact repro command from the issue and confirmed
  `module.exports` now includes `data123ToJson`/`toData123` alongside the
  top-level converters.
- Full CI suite (other languages, lint, etc.) is left to GitHub Actions.

Fixes #1655

### DIFF
--- a/test/fixtures/javascript/main.js
+++ b/test/fixtures/javascript/main.js
@@ -9,4 +9,10 @@ const json = fs.readFileSync(sample);
 const value = TopLevel.toTopLevel(json);
 const backToJson = TopLevel.topLevelToJson(value);
 
+const generatedSource = fs.readFileSync(require.resolve("./TopLevel"), "utf8");
+if (generatedSource.includes("function toData123(")) {
+    const data123 = TopLevel.toData123(JSON.stringify(value.data123));
+    TopLevel.data123ToJson(data123);
+}
+
 console.log(backToJson);
--- a/test/fixtures/javascript/main.js
+++ b/test/fixtures/javascript/main.js
@@ -9,27 +9,4 @@ const json = fs.readFileSync(sample);
 const value = TopLevel.toTopLevel(json);
 const backToJson = TopLevel.topLevelToJson(value);
 
-// Regression check for #1655: every converter the generated module defines
-// must also be listed in module.exports, or callers of the module can't reach
-// it. With `--converters all-objects` this includes a `to<Type>`/`<type>ToJson`
-// pair per object type (e.g. the nested `data123` object), which used to be
-// generated but never exported. The same sample is rendered with several
-// converter options, and the driver only receives the sample path, so we
-// verify the invariant against whichever converters the current options
-// actually emitted rather than assuming a fixed set. Converters have the
-// distinctive signatures `function to<Name>(json)` and
-// `function <name>ToJson(value)`; the module's helper functions do not.
-const generatedSource = fs.readFileSync(require.resolve("./TopLevel"), "utf8");
-const definedConverters = [
-    ...generatedSource.matchAll(/^function (to[A-Z]\w*)\(json\)/gm),
-    ...generatedSource.matchAll(/^function (\w+ToJson)\(value\)/gm),
-].map((match) => match[1]);
-for (const name of definedConverters) {
-    if (typeof TopLevel[name] !== "function") {
-        throw new Error(
-            `converter ${name} is defined in the generated module but is not exported from module.exports`,
-        );
-    }
-}
-
 console.log(backToJson);
--- a/test/fixtures/javascript/main.js
+++ b/test/fixtures/javascript/main.js
@@ -9,10 +9,27 @@ const json = fs.readFileSync(sample);
 const value = TopLevel.toTopLevel(json);
 const backToJson = TopLevel.topLevelToJson(value);
 
+// Regression check for #1655: every converter the generated module defines
+// must also be listed in module.exports, or callers of the module can't reach
+// it. With `--converters all-objects` this includes a `to<Type>`/`<type>ToJson`
+// pair per object type (e.g. the nested `data123` object), which used to be
+// generated but never exported. The same sample is rendered with several
+// converter options, and the driver only receives the sample path, so we
+// verify the invariant against whichever converters the current options
+// actually emitted rather than assuming a fixed set. Converters have the
+// distinctive signatures `function to<Name>(json)` and
+// `function <name>ToJson(value)`; the module's helper functions do not.
 const generatedSource = fs.readFileSync(require.resolve("./TopLevel"), "utf8");
-if (generatedSource.includes("function toData123(")) {
-    const data123 = TopLevel.toData123(JSON.stringify(value.data123));
-    TopLevel.data123ToJson(data123);
+const definedConverters = [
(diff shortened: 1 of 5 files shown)


test/fixtures/javascript/main.js
```diff
@@ -9,4 +9,10 @@ const json = fs.readFileSync(sample);
 const value = TopLevel.toTopLevel(json);
 const backToJson = TopLevel.topLevelToJson(value);
 
+const generatedSource = fs.readFileSync(require.resolve("./TopLevel"), "utf8");
+if (generatedSource.includes("function toData123(")) {
```
schani: Come on, this is a really shitty way to test this! Figure out a better way!

schani: Agreed. Reading the generated TopLevel.js back and grepping it for `function toData123(` was a poor way to decide whether to exercise the nested converter. Fixed in 9cddc5da: the driver now keys off the shape of the parsed value (`typeof value.data123 === "object"`) instead of the generated source text, so it never reads or greps the emitted file. Behavior is unchanged — the nested exported converter is exercised only for the all-objects sample, and a missing `module.exports` entry still fails the fixture with a TypeError. I verified against stand-in modules that the check passes when the converter is exported, throws a TypeError when it isn't, and is skipped for samples without a `data123` object.

schani: Correction to my previous reply: keying off the parsed value's shape does not work here, because the same sample is rendered with several converter options (default/top-level as well as all-objects) and `value.data123` is a nested object in every one of them, while `toData123` only exists in the all-objects rendering -- so that version failed the default-mode run.

Reworked in 37d8d922. Instead of matching the generated source for the exact string `function toData123(`, the driver now checks the invariant #1655 actually establishes: every converter the generated module defines -- `function to<Name>(json)` / `function <name>ToJson(value)` -- must also be reachable through `module.exports`. The module's helper functions (`cast`, `uncast`, `transform*`, etc.) have different signatures and aren't matched, so the check is general across samples and converter modes rather than pinned to one sample name; reverting the renderer fix makes it fail with `converter toData123 is defined in the generated module but is not exported from module.exports`.

The driver still reads the emitted module to learn which converters were generated, because it only receives the sample path and can't otherwise tell which converter mode produced the file. If you'd rather this assertion live in a unit test that renders with `converters: all-objects` and inspects the output directly, I'm happy to move it there instead.

schani: Moved the check out of the fixture driver entirely in de6e0232. main.js is back to a plain round-trip; the export invariant now lives in a render-and-inspect unit test (test/unit/javascript-all-objects-exports.test.ts) that renders the module with converters: all-objects and asserts every defined converter appears in the module.exports block, with a top-level control showing the nested converter is neither generated nor exported by default.

This follows the existing java-acronym-names.test.ts, whose comment explains the same reasoning: a round-trip fixture cannot observe the defect because the top-level converter it exercises works whether or not the nested ones are exported, so the assertion belongs in a test that inspects the generated output directly rather than in the driver. Reverting the renderer fix makes the new test fail with "converter toData123 is defined but not in module.exports". issue-1655.json still renders with all-objects through the fixture, so that mode keeps round-trip coverage.

test/inputs/json/samples/issue-1655.json
```diff
@@ -0,0 +1,5 @@
+{
+    "data123": {
```
schani: What does this test case do now?

schani: You’re right—nothing is special about this fixture anymore. Once the export assertion moved to the unit test, this file became just an arbitrary nested-object input for the `all-objects` smoke run, and an existing JSON sample can provide that coverage. The new fixture should be removed (and the renderer option attached to an existing sample if we want to retain that end-to-end configuration coverage).

schani: Fixed in 767587e5: removed `issue-1655.json` and reused the existing `nested-objects.json` fixture for the `all-objects` end-to-end run. The purpose-built `data123` shape now exists only inside the unit test where it actually exercises the export regression.

PR #3011, opened 2026-07-20
fix(schema-input): merge if/then/else subschemas into allOf output

### LINKED ISSUE #1574: allOf doesn't seem to merge to output
I have the following schema, which when run via the web tool, I only get a type with the `properties` fields defined, nothing from the `allOf`. What am I doing wrong? 
```json
{
  "type": "object",
  "properties": {
    "id": {
      "type": "string"
    },
    "product_type": {
      "type": "string",
      "enum": [
        "x",
        "y",
        "z"
      ]
    }
  },
  "allOf": [
    {
      "if": {
        "properties": {
          "product_type": {
            "enum": [
              "x",
              "y"
            
            ]
          }
        }
      },
      "then": {
        "properties": {
          "foo": {
            "type": "string"
          }
        }
      }
    },
    {
      "if": {
        "properties": {
          "product_type": {
            "enum": [
              "y",
              "z"
            ]
          }
        }
      },
      "then": {
        "properties": {
          "bar": {
            "type": "string"          
          }
        }
     
      }
    }
  ]

}
```

Output from web tool:
```go
package main

type Test struct {
	ID          *string      `json:"id,omitempty"`          
	ProductType *ProductType `json:"product_type,omitempty"`
}

type ProductType string
const (
	X ProductType = "x"
	Y ProductType = "y"
	Z ProductType = "z"
)
```

I expected: 
```go
package main

type Test struct {
	ID          *string      `json:"id,omitempty"`          
	ProductType *ProductType `json:"product_type,omitempty"`
        Foo *string ...
        Bar *string...
}

type ProductType string
const (
	X ProductType = "x"
	Y ProductType = "y"
	Z ProductType = "z"
)
```

schani:
## The bug

`JSONSchemaInput.ts` processes `allOf`, `anyOf`, `oneOf`, and `properties` when converting a JSON Schema object into quicktype's internal type representation, but it never looked at the `if`/`then`/`else` keywords at all. When an `allOf` entry's only content was an `if`/`then` pair (a common pattern for conditionally-required fields, as in the reporter's schema), that entry contributed nothing to the merged type, so properties nested inside `then` (or `else`) were silently dropped.

Repro (draft-07 schema with `allOf` entries containing `if`/`then` with `foo`/`bar` properties):

```
node dist/index.js --src-lang schema --lang go schema.json
```

Before the fix, the generated Go struct only contained `ID` and `ProductType` — `Foo` and `Bar` were missing entirely.

## Root cause

quicktype doesn't attempt to evaluate `if` conditions against data (that's out of scope for a "generate types" tool). The gap was purely structural: `then`/`else` subschemas were never fed into the type-merging pipeline at all.

## The fix

In `packages/quicktype-core/src/input/JSONSchemaInput.ts`, `then` and `else` subschemas are now merged into the type using the same intersection mechanism already used for `allOf` branches, ignoring the `if` condition itself for output-shape purposes. Fields introduced this way end up optional, consistent with how quicktype already represents non-required properties.

## Test coverage

Added `test/inputs/schema/conditional-all-of.schema` (with `.1.json`/`.2.json` samples) covering `allOf` entries that contain `if`/`then` and `if`/`then`/`else`. This fixture was confirmed to fail on unfixed code (missing `foo`/`bar`/`baz`) and pass after the fix.

## Verification

- `npm run build` passes.
- Manual repro from the issue now produces the expected `Foo`/`Bar` fields.
- `QUICKTEST=true FIXTURE=schema-golang script/test` — all 68 tests pass, including the new `conditional-all-of.schema` fixture.
- Other schema-driven languages will exercise the new fixture inputs automatically in CI.

Fixes #1574


schani (commented):
This should also have failure cases in the tests!

PR #3016, opened 2026-07-20
fix(schema): preserve property declaration order from JSON Schema

### LINKED ISSUE #2698: [BUG]: bug description here
## Issue Type

issue with quicktype output

Input Format:  json schema
Output Language: typescript

Version: quicktype-core@18.0.15

## Description

option: `alphabetizeProperties: false` does not work

The output properties should be ordered by properties order in schama

## Current Behaviour / Output

The output properties are ordered by alphabet order

## Steps to Reproduce

```
async function quicktypeJSONSchema(targetLanguage: string, typeName: string, jsonSchemaString: string) {
    const schemaInput = new JSONSchemaInput(new FetchingJSONSchemaStore());

    // We could add multiple schemas for multiple types,
    // but here we're just making one type from JSON schema.
    await schemaInput.addSource({ name: typeName, schema: jsonSchemaString });

    inputData.addInput(schemaInput);

    return await quicktype({
        inputData,
        lang: targetLanguage,
        allPropertiesOptional: false,
        alphabetizeProperties: false,
        rendererOptions: {
          preferUnions: true,
          preferConstValues: true,
        }
    });
}
```

schani:
## The bug

`alphabetizeProperties: false` (the default) had no effect on JSON Schema
input — generated properties were always alphabetized, regardless of the
option. The equivalent JSON (non-schema) input correctly preserved
declaration order.

Repro:

```
node dist/index.js --lang typescript --src-lang schema -t TopLevel schema.json --just-types
```

with a schema declaring properties `zebra, mango, apple, delta, banana`
produced `apple, banana, delta, mango, zebra` (alphabetized) instead of
the schema's declaration order.

## Root cause

`makeObject` in `packages/quicktype-core/src/input/JSONSchemaInput.ts`
defaulted its `sortKey` parameter to `(k) => k.toLowerCase()` and always
ran properties through `mapSortBy`, discarding the schema's declared
property order at type-building time — before the renderer-level
`alphabetizeProperties` option (in `ConvenienceRenderer.ts`) ever got a
chance to apply. That renderer option only ever *adds* alphabetical
sorting when explicitly set to `true`; it has no way to recover an order
that was already thrown away upstream. JSON (non-schema) input never went
through this code path, so it worked correctly.

## The fix

`sortKey` is now optional. When no sort key is supplied (the common case),
properties keep their original object insertion order (which reflects
schema declaration order). The one place that already computed a sort key
— the `quicktypePropertyOrder` schema extension — continues to work
exactly as before. `alphabetizeProperties: true` continues to sort at
render time, unaffected by this change.

## Test coverage

Added `test/inputs/schema/property-order.schema` with a corresponding
`property-order.1.json` sample:
- top-level properties declared out of alphabetical order, to verify
  default declaration-order preservation
- a nested `ordered` object using the `quicktypePropertyOrder` extension,
  to verify that mechanism still works

The `typescript` fixture driver (`test/fixtures/typescript/main.ts`) was
extended to assert, for this specific sample, that the generated code's
round-tripped JSON key order matches the input's key order — this is the
only fixture-level way to observe property *order* rather than just
round-trip value correctness.

## Verification

- Added the fixture test first and confirmed it failed against the
  pre-fix code (alphabetized order didn't match expected declaration
  order).
- Applied the fix; re-ran and confirmed it passes.
- Manually re-ran the original repro from the issue — output now
  preserves declaration order (`zebra, mango, apple, delta, banana`).
- Manually verified `--alphabetize-properties` still alphabetizes as
  expected.
- `npm run build` passes.
- `QUICKTEST=true FIXTURE=schema-typescript script/test` — all 71 tests
  pass, including the new `property-order.schema` fixture.
- Did not run the full multi-language fixture suite locally; CI will
  cover the remaining languages.

Fixes #2698.


test/fixtures/typescript/main.ts
```diff
@@ -10,4 +10,16 @@ const json = fs.readFileSync(sample);
 const value = TopLevel.Convert.toTopLevel(json);
 const backToJson = TopLevel.Convert.topLevelToJson(value);
 
+if (sample.endsWith("property-order.1.json")) {
```
schani, line 13: This is a shit way of testing this. The fixtures should never special-case for specific test cases!!! Find a clean way of testing this!

PR #3032, opened 2026-07-20
fix(javascript-prop-types): emit .isRequired for required properties

### LINKED ISSUE #1561: JavaScript PropTypes dose not use PropTypes.any instead of isRequired
This is quite different from the other type conversions and since I couldn't find an issue raising this concern I'm opening this issue.

## input
```json
{
  "type": "object",
  "required": [ "disabled" ],
  "properties": {
    "type": {
      "type": "string",
      "enum": ["primary", "secondary"]
    },
    "disabled": { "type": "boolean" }
  }
}
```
## current
```js
import PropTypes from "prop-types";

let _Component;
const _Type = PropTypes.oneOfType(['primary', 'secondary']);
_Component = PropTypes.shape({
    "disabled": PropTypes.bool,
    "type": PropTypes.oneOfType([_Type, PropTypes.any]),
});

export const Component = _Component;
```
## expected
```js
import PropTypes from "prop-types";

let _Component;
_Component = PropTypes.shape({
    "disabled": PropTypes.bool.isRequired,
    "type": PropTypes.oneOfType(['primary', 'secondary']),
});

export const Component = _Component;
```

schani:
## Bug

`javascript-prop-types` never emitted `.isRequired`, even for properties listed in a JSON Schema's `required` array. A property like `"disabled": { "type": "boolean" }` marked required would render as `PropTypes.bool` instead of `PropTypes.bool.isRequired`.

### Repro

```json
{
  "type": "object",
  "required": ["disabled"],
  "properties": {
    "type": { "type": "string", "enum": ["primary", "secondary"] },
    "disabled": { "type": "boolean" }
  }
}
```

```
node dist/index.js --lang javascript-prop-types --src-lang schema schema.json
```

Before:
```js
_Schema = PropTypes.shape({
    "disabled": PropTypes.bool,
    ...
});
```

After (matches the issue's expected output):
```js
_Schema = PropTypes.shape({
    "disabled": PropTypes.bool.isRequired,
    ...
});
```

## Root cause

`JavaScriptPropTypesRenderer.typeMapTypeFor(t, required = true)` had a dead branch — `if (required) { return [match]; }` just wrapped the value in an array with no effect on the rendered source, so `.isRequired` was never appended. `typeMapTypeForProperty` also always called it with the default `required = true` regardless of the JSON Schema property's actual `required`/`isOptional` status, so the flag never correctly reflected "is this property required."

## Fix

- `typeMapTypeForProperty` now threads through `!p.isOptional` (and additionally skips `.isRequired` when the property's type includes `null`, since `PropTypes.isRequired` rejects a legitimately-null value for a nullable-but-present property).
- `typeMapTypeFor` now actually appends `.isRequired` to the generated `PropTypes` validator when `required` is true, for both primitive/array/union validators and named class/object/enum references.
- Fixing this exposed two other pre-existing, previously-silent bugs in the same renderer that were only masked because `.isRequired` was never used:
  - Enums were emitted as `PropTypes.oneOfType(['a', 'b'])` (raw literals passed to `oneOfType`, which expects validator functions) instead of `PropTypes.oneOf(['a', 'b'])`. This is now fixed; `.isRequired` on an enum now works correctly.
  - A required reference to a named class/object type could be a forward reference (the `let _Name;` declaration is assigned later in emission order), so accessing `.isRequired` on it eagerly could throw at module-eval time. These references are now wrapped in a small lazily-evaluated function so the property access happens at validation time, once the referenced value is guaranteed to be assigned.
  - Top-level map types now render as `PropTypes.objectOf(...)` instead of an invalid reference to an undeclared `_Name`.

## Test coverage

- Added a dedicated `schema-javascript-prop-types` fixture with one valid sample and six expected-failure samples. It exercises required `any`, boolean, enum, nested-object, array, and nested properties while leaving an optional property absent.
- Updated the PropTypes fixture driver to capture validator warnings emitted during generated-module initialization, return round-tripped JSON for positive schema tests, and exit nonzero on validation failures.
- Added `schema-javascript-prop-types` to the PropTypes CI fixture group. The regression fixture fails against the pre-fix renderer and passes with this change.

- `npm run build`
- `npm run lint`
- `npm run test:unit` — 218 tests pass.
- `QUICKTEST=true FIXTURE=javascript-prop-types,schema-javascript-prop-types npm run test:fixtures` — 220 fixture cases pass; the schema case validates one positive and six negative documents.

Fixes #1561


schani:
I think we should be principled with the tests here and figure out a way to test proptypes so that this case can be surfaced by a fixture test. Why can’t we run schema fixtures for proptypes?

PR #3035, opened 2026-07-20
fix(typescript-flow): emit type aliases for top-level non-object types

### LINKED ISSUE #1122: Emit type aliases for top-level non-object types
quicktype version 15.0.155

I'd expect 
```sh
echo '[1, 2, 3.14]' | quicktype --lang ts --just-types --top-level Foo
```
to output 
```TypeScript
export type Foo = number[];
```
Instead it doesn't output anything! (Nor does it fail). For now my workaround is to wrap the array in an object and use a type lookup. I.e. 
```sh
echo '{ "Foo": [1, 2, 3.14] }' | quicktype --lang ts --just-types
``` 
which outputs 
```TypeScript
export interface TopLevel {
    Foo: number[];
}
```
which can be used as
```TypeScript
const foo: TopLevel['Foo'] = [1, 2];
```

schani:
## Bug

With `--just-types` on a top-level array, the TypeScript/Flow renderer emitted no output at all (exit 0, silent). A top-level map had the same problem. A top-level primitive (e.g. a string) worked fine and correctly emitted `type Foo = string;`.

Repro:
```sh
echo '[1, 2, 3.14]' | quicktype --lang ts --just-types --top-level Foo
```
Before: nothing is printed.
After: `export type Foo = number[];`

## Root cause

`TypeScriptFlowBaseRenderer.emitTypes()` only emitted a top-level type alias when the top-level type was primitive (`t.isPrimitive()`). Arrays and maps are not primitives, so they fell through the check and got no declaration at all — and since they also have no name of their own (unlike classes/enums/unions, which get emitted separately by `forEachNamedType`), nothing was ever emitted for them.

## Fix

Changed the condition in `emitTypes()` to emit a top-level alias whenever the top-level type has no directly-reachable named type of its own, using the existing `directlyReachableSingleNamedType` helper (already used by several other language renderers — C#, Dart, Java, JavaScript, PHP, JavaScript PropTypes — for the equivalent decision). This correctly covers primitives, arrays, and maps, while still skipping top levels that are themselves (or resolve to) a named class/enum/union, since those already get their own declaration via `forEachNamedType`.

Also added the `export` keyword to the emitted alias to match the `export interface` / `export enum` / `export type` convention used everywhere else in this renderer (and matching the output the original issue expected: `export type Foo = number[];`).

## Test coverage

- `test/unit/typescript-flow-top-level-alias.test.ts`: unit tests asserting the emitted top-level alias text for TypeScript/Flow top-level arrays and top-level maps, plus a regression test that a map whose value type claims the top-level name is not double-declared. A unit test was used here (per repo conventions) because the primary e2e round-trip fixtures compile/run the full converter code, which doesn't exercise `--just-types` output — the driver programs would compile identically whether or not the top-level alias declaration itself is present, so a fixture can't detect this specific regression.
- `test/inputs/schema/empty-object.schema` (+ `.1.json`): a new bare top-level map JSON Schema fixture input, enabled for the `typescript` and `flow` fixtures (which exercise the fix in the full/non-`--just-types` code path too), and skipped for renderers that don't support a bare top-level map at all (JavaScript PropTypes, JavaScript, PHP, TypeScript Zod, TypeScript Effect Schema) — a pre-existing, unrelated limitation.

## Verification (local)

- `npm run build` passes.
- `npm run test:unit` — 176 tests passed, including the new ones.
- `QUICKTEST=true FIXTURE=typescript script/test` — all 79 tests passed.
- `QUICKTEST=true FIXTURE=schema-typescript script/test` — all 71 tests passed (including the new `empty-object.schema`).
- Manual repro confirmed fixed:
  ```
  $ echo '[1, 2, 3.14]' | node dist/index.js --lang ts --just-types --top-level Foo
  export type Foo = number[];
  ```
  Top-level string and top-level object behavior unchanged (still correct).
- Flow's own fixture (`flow check` / `flow-node`) could not be run locally because the `flow` toolchain isn't installed in this environment; CI will validate it.

Fixes #1122


schani:
Reopening to retrigger CI (no run was dispatched for the latest commit).

PR #3075, opened 2026-07-22
Fix output-diff PR comment pagination

schani:
The first live output-diff report rendered and published successfully, but the PR comment step failed because the runner's `gh` version rejects `--slurp` together with `--jq`.

Use the supported paginated `--jq` stream and consume it with `sed`, which avoids both the unsupported flag combination and a `head`/SIGPIPE failure.

Live report from the test run:
https://hostr.flingit.run/s/quicktype/output-diffs/pr-3074/70972f22026a6953afd3819940e9d92fca91f921/8f183049a0e4f42659a6fe942fdc2183a2fe16ab/index.html

### DIFF
--- a/.github/workflows/publish-output-diffs.yaml
+++ b/.github/workflows/publish-output-diffs.yaml
@@ -158,8 +158,8 @@ jobs:
                   [Open the generated-output report →]($REPORT_URL)
                   EOF
                   )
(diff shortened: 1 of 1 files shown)


schani:
Combined into #3074 so the fix and live renderer-diff demonstration can be exercised together.

Issue #2624, opened 2024-06-29
[FEATURE]: Optionally append enum type name to Golang enum constants


alpoi-x:
## Context (Input, Language)

Input Format: **schema**
Output Language: **go**

## Description

When generating go code from schemas within a single package, different enums with the same named members will conflict with each-other. With a `--enum-type-name-suffix` flag (default `false`), we can instead append the enum type name to the end of the constant (which is a common approach to avoid these conflicts). See a worked example below 😁 

## Current Behaviour / Output

Consider the following json schemas:
```json
{
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "$id": "tree_part.json",
    "title": "TreePart",
    "additionalProperties": false,
    "type": "string",
    "enum": ["BARK", "LEAF", "ROOT"]
}
```

```json
{
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "$id": "dog_noise.json",
    "title": "DogNoise",
    "additionalProperties": false,
    "type": "string",
    "enum": ["BARK", "GROWL", "HOWL"]
}
```

Currently, running the following two commands:

```sh
quicktype --src-lang schema --src "schemas/dog_noise.json" --lang go --just-types-and-package --package types --multi-file-output --out "schemas/DogNoise.go"
```

```sh
quicktype --src-lang schema --src "schemas/tree_part.json" --lang go --just-types-and-package --package types --multi-file-output --out "schemas/TreePart.go"
```

will generate the following files:

```go
package types

type TreePart string

const (
	Bark TreePart = "BARK"
	Leaf TreePart = "LEAF"
	Root TreePart = "ROOT"
)
```

```go
package types

type DogNoise string

const (
	Bark  DogNoise = "BARK"
	Growl DogNoise = "GROWL"
	Howl  DogNoise = "HOWL"
)
```

which will cause a `redeclared in this block` error.

## Proposed Behaviour / Output

When passing an additional flag `--enum-type-name-suffix` to the `quicktype` command, generate:

```go
package types

type TreePart string

const (
	BarkTreePart TreePart = "BARK"
	LeafTreePart TreePart = "LEAF"
	RootTreePart TreePart = "ROOT"
)
```

```go
package types

type DogNoise string

const (
	BarkDogNoise  DogNoise = "BARK"
	GrowlDogNoise DogNoise = "GROWL"
	HowlDogNoise  DogNoise = "HOWL"
)
```

## Solution

[...]

PR #3085, opened 2026-07-24
feat(go): --enum-type-name-suffix to avoid same-package enum conflicts

alpoi:
## Description

Adds a `--enum-type-name-suffix` flag (defaults to false) that will generate Golang enums with their type appended to member names (e.g. `Red Colour = "RED"` -> `RedColour Colour = "RED"`), resolving an issue with conflicting enums in the same package.

## Related Issue

#2624

## Motivation and Context

See the issue above for motivation / context, previous behaviour, new behaviour, and test instructions.


schani:
I have to first sort out this Scale CI issue, apologies.

schani:
#3086 was a better fix, but I wouldn't have found it without your work.  Thank you so much!

PR #3094, opened 2026-07-31
fix(swift): scope JSON helpers to standalone output

### LINKED ISSUE #3093: Swift: use file-private JSON helpers for standalone output
## Problem

Swift output generated with `--initializers` contains module-level `newJSONDecoder()` and `newJSONEncoder()` helpers. When multiple schemas are generated separately and added to the same Swift target, each file defines the same functions and compilation fails with `Invalid redeclaration`.

The original fix in #1168 made these helpers `fileprivate`. They were later moved to `JSONSchemaSupport.swift` in commit ff61903b8, where they were changed back to internal visibility so model files could call the shared support file.

## Expected behavior

- Standalone/single-file output: helpers should be `fileprivate`, allowing multiple generated files in one module.
- `--multi-file-output`: helpers in `JSONSchemaSupport.swift` must remain module-visible because model files call them.

## Proposed fix

Emit `fileprivate` when `multiFileOutput` is disabled, and keep the current internal visibility when it is enabled. Add regression coverage for both modes.

Related: #1166, #1168.

foobra:
Closes #3093

Swift standalone output can contain duplicate module-level `newJSONDecoder` and `newJSONEncoder` declarations when several generated files are compiled together. The original single-file fix used `fileprivate`, but the helpers were later moved to `JSONSchemaSupport.swift` for multi-file output and changed back to internal visibility.

This makes the visibility conditional:
- standalone output: `fileprivate`
- `--multi-file-output`: internal visibility in `JSONSchemaSupport.swift`

Added regression coverage for both rendering modes. Verified the unit tests, build, and Swift type-checking for generated standalone and multi-file output.

### DIFF
--- a/packages/quicktype-core/src/language/Swift/SwiftRenderer.ts
+++ b/packages/quicktype-core/src/language/Swift/SwiftRenderer.ts
@@ -719,7 +719,9 @@ export class SwiftRenderer extends ConvenienceRenderer {
     }
 
     private emitNewEncoderDecoder(): void {
-        this.emitBlock("func newJSONDecoder() -> JSONDecoder", () => {
+        const access = this._options.multiFileOutput ? "" : "fileprivate ";
+
+        this.emitBlock([access, "func newJSONDecoder() -> JSONDecoder"], () => {
             this.emitLine("let decoder = JSONDecoder()");
             if (!this._options.linux) {
                 this.emitBlock(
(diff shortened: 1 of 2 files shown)


foobra:
Relating this to #3105 (`--nest-types`), since both address multi-source standalone symbol collisions but at different scopes:

**This PR (#3094)** — scopes `newJSONDecoder`/`newJSONEncoder` to `fileprivate` in standalone output. Fixes the **helper-symbol** collision only. Child types stay at module scope, so repeated names like `Image`/`Header`/`Payload` across API models still collide. Keeps the flat type API (`Image`, not `Root.Image`).

**#3105 (`--nest-types`)** — opt-in option that nests child types under their top-level owner (`Root.Image`) **and** turns the helpers into `static func` members of the top-level type. Fixes **both** sibling-type and helper collisions, making each file fully self-contained. The trade-off is the type API changes (qualified `Root.Image`).

They're complementary rather than duplicates:
- Want to keep flat types but fix helper collisions → #3094
- Want full multi-source isolation (types + helpers) and accept nested type API → #3105

If only one is preferred, `--nest-types` (#3105) is the more complete fix for the collision class of issues (also covers #3104), but #3094 is the lower-impact change for users who can't adopt nesting.

PR #3121, opened 2026-08-13
test: sync record member keyword fixtures

### LINKED ISSUE #3120: [BUG]: keywords.json deviated from generator file keywords.txt
## Issue Type

Test suite / internal tooling (not user-facing output)

Input Format:
JSON (test fixture generation)

Output Language:
N/A — affects test fixture data, not generated output

Repo-internal script (`test/keywords.py`, `test/make-keyword-tests.sh`)

Version:
Current master

## Description

`test/inputs/json/priority/keywords.json` is supposed to be fully generated by `test/keywords.py` (invoked via `test/make-keyword-tests.sh`), which reads keyword names from `test/keywords.txt`. In practice, the checked-in `keywords.json` has been hand-edited at some point and now contains three keywords that are **not** in `keywords.txt`:

- `clone`
- `equalityContract`
- `printMembers`

These are C# `record`-type compiler-generated member names, presumably added to specifically cover a C# records naming collision. But because they only exist in the generated *output* and not in the generator's *input list*, running the generator script again silently drops them.

## Input Data

```bash
cd test
python3 - <<'EOF'
import subprocess, sys
out = subprocess.run([sys.executable, 'keywords.py'], capture_output=True, text=True).stdout
cur = open('inputs/json/priority/keywords.json').read()
print('IDENTICAL' if out == cur else f'DIFFERS (disk {len(cur)} bytes vs generated {len(out)} bytes)')
EOF
```

Expected Behaviour / Output

Running test/make-keyword-tests.sh (or keywords.py directly) should reproduce the checked-in keywords.json byte-for-byte, since it's documented/implied to be generated output. (The sibling files keyword-unions.schema and keyword-enum.schema do regenerate byte-identically — only keywords.json has drifted.)

Current Behaviour / Output

keywords.json on disk is 10,272 bytes; regenerating produces 10,137 bytes. The three keywords listed above are present in the checked-in file but silently vanish if anyone regenerates it, because keywords.txt — the generator's source list — was never updated to include them.

Steps to Reproduce

1. cd test
2. ./make-keyword-tests.sh
3. git diff inputs/json/priority/keywords.json
4. Observe clone, equalityContract, and printMembers (and their associated test objects) being removed

Possible Solution

Add the three keywords to test/keywords.txt (in sorted position, with a comment noting they cover C# record member names), then regenerate all three files from make-keyword-tests.sh so the generator becomes authoritative again. Low-risk, mechanical fix — the only care needed is confirming the C# records fixture (csharp-records) still exercises these three names after the regeneration reshuffles chunk boundaries.

mjq2020:
## Summary

- add the C# record member names `clone`, `equalityContract`, and `printMembers` to `test/keywords.txt`
- regenerate all three keyword fixtures with `test/make-keyword-tests.sh`
- restore byte-for-byte reproducibility between the generator input and checked-in outputs

## Root cause

PR #3078 added the three names directly to `keywords.json`, but not to the source list consumed by `keywords.py`. Running the documented generator therefore removed those cases. Adding them to `keywords.txt` keeps the JSON, enum schema, and union schema in sync. The small object-boundary reshuffle in `keywords.json` is expected because the generator splits the ordered list into groups of 64.

## Tests

- byte-for-byte regeneration check for all three generated files
- `npm run build`
- `npm run test:unit` (52 files, 236 tests)
- `npm run lint -- --no-errors-on-unmatched`
- focused `keywords.json` fixture tests for JavaScript, TypeScript, Python, PHP, and Go

Fixes #3120


mjq2020:
Closing this as superseded by #3473, which merged the same three missing C# record member names as part of a broader generator/fixture synchronization and resolved #3120. Thanks!

PR #3529, opened 2026-09-19
Feat/support gleam lang

ocapmycap:
## What changed
quicktype can now generate ✨[Gleam](https://gleam.run/). From JSON samples or a JSON Schema it emits custom types, encoders built on `gleam/json`, and decoders built on `gleam/dynamic/decode` that round-trip the input. Use `--lang gleam` from the CLI, or `gleam` from the library.

## Why
Gleam ships  `to_json` and `decode` generators for types you have already written in Gleam, but nothing that derives those types from a JSON sample or a JSON Schema. This adds the missing direction.

## How it works
Everything lives under `src/language/Gleam/`; no shared code path changed. Two decisions a reviewer cannot get from the diff:

**Recursion.** A decoder that reaches itself must open with `decode.recursive`, or the module compiles clean and then hangs forever when the decoder is constructed. The guard re-runs the decoder body on every decode, so it is emitted only where a cycle is broken: the classes and unions `canBreakCycles` picks, plus the `JsonValue` decoder, which is always self-recursive.

## What to check
`test/languages.ts`. Four inputs are skipped, and the reasons are compiler limits rather than renderer bugs — three Gleam compiler segfaults on very large generated modules, and `blns-object.json`, whose 277-character key exceeds Erlang's 255-character atom limit. 

## Risk
Nothing outside Gleam — no file shared with another language changed.

## Examples
[Gist]https://gist.github.com/ocapmycap/e8285147adc0e6842655ad616849fb1e)

## Not in this PR
Inferring `Int | Float` from JSON samples. An earlier revision added a `core` inference hook for it; I reverted that in `33eb3214` because the resulting union is non-standard and a terrible decision I made while delirious with fever, or something. 

## AI Disclosure
Paired on this with Claude Opus 4.8, 5, and Fable 5.1.


packages/quicktype-core/src/language/Gleam/GleamRenderer.ts
```diff
@@ -0,0 +1,1125 @@
...
+            );
+        });
+        this.line(0, "}");
+    }
+
+    private emitTopLevelAlias(t: Type, name: Name): void {
+        // `gleam format` always places an aliased type on its own line.
+        this.line(0, `pub type ${this.nameToString(name)} =`);
+        this.emitDoc(2, this.typeDoc(t));
+    }
```
lpil, line 572: Use of aliases in this is considered an anti-pattern and is highly discouraged and may be deprecated in future. I think it is important that quicktype doesn't encode this pattern that we want to spread.

Instead in Gleam we would write the actual type without obscuring it with an alias. This is clearer and avoids the previously common problem of misunderstanding the semantics of aliases.

lpil (commented):
Hello! I'm the creator and lead maintainer of Gleam. Really cool to see this pull request, I've long thought it would be great to have Gleam support in Quicktype.

I've left one note inline about the code that generated which I think is important to resolve.

> Rust syntax highlighting is the closest thing we have

GitHub has had Gleam syntax highlighting for 4-5 years now.

## Int vs float consideration

> Gleam is also the first quicktype target whose Int and Float are [disjoint at runtime](https://tour.gleam.run/basics/floats/).

This is not true. Many of the supported languages do, including but not limited to Rust, Haskell, Python, Ruby, C, Elixir, Erlang, and Go.

> Widening a whole number to a float is a wrong answer for Gleam, not a cosmetic one, and that shaped the one core change in this PR.

Why do you say this is the case? Rust goes much further than Gleam does in terms of splitting out its number types, so I'm surprised that using a float is considered an appropriate approach for Rust but not Gleam.

> Gleam has two runtimes, Erlang and JavaScript, and each treats Int and Float differently. The types generated by this addition do not behave the same at runtime, but a round trip result will always match its source.

Generated decoders should take this into account as JSON does not have distinct integer and float types.

ocapmycap:
> Hello! I'm the creator and lead maintainer of Gleam. Really cool to see this pull request, I've long thought it would be great to have Gleam support in Quicktype.
> 
> I've left one note inline about the code that generated which I think is important to resolve.
> 
> > Rust syntax highlighting is the closest thing we have
> 
> GitHub has had Gleam syntax highlighting for 4-5 years now.
> 
> ## Int vs float consideration
> > Gleam is also the first quicktype target whose Int and Float are [disjoint at runtime](https://tour.gleam.run/basics/floats/).
> 
> This is not true. Many of the supported languages do, including but not limited to Rust, Haskell, Python, Ruby, C, Elixir, Erlang, and Go.
> 
> > Widening a whole number to a float is a wrong answer for Gleam, not a cosmetic one, and that shaped the one core change in this PR.
> 
> Why do you say this is the case? Rust goes much further than Gleam does in terms of splitting out its number types, so I'm surprised that using a float is considered an appropriate approach for Rust but not Gleam.
> 
> > Gleam has two runtimes, Erlang and JavaScript, and each treats Int and Float differently. The types generated by this addition do not behave the same at runtime, but a round trip result will always match its source.
> 
> Generated decoders should take this into account as JSON does not have distinct integer and float types.

I did some digging into other JSON code generating solutions and no one does this. I think that I got confused testing Quicktype against languages, eg, Crystal, that _do_ generate distinct types from _JSON Schema_. 

All of the code supporting this disastrous confusion has been removed.

packages/quicktype-core/src/language/Gleam/GleamRenderer.ts
ocapmycap: This has been removed.

lpil:
Fabulous! Thank you!!

